PR #69, opened 2026-04-12
Provide a motivation case for Protocol

Twon:


jbcoe (commented):
I love this! I’ll get the pre-commits working and merge it.

Thanks

PR #81, opened 2026-04-29
Fix typos in the draft

purpleKarrot:
A few issues that I found while reading the paper.

Some others issues that caught my eye, but I wanted to clarify:

> | Aspect | `protocol` | `proxy` (P3086) |
> | :--- | :--- | :--- |
> | Interaction syntax | Value semantics: `p.draw()` | Pointer semantics: `p->draw()` |

Should this be "Value syntax" and "Pointer syntax"? Or just use the code alone?

> asks the vtable to provide a movecopy of the storage

Is "movecopy" a known term?


jbcoe:
Thanks, the draft has evolved so I've applied these (rather essential!) fixes manually.

PR #96, opened 2026-07-22
[DO NOT MERGE] Add tutorials for type-erasure, reflection and vanishing-this

jbcoe:
I'm not entirely sure that these tutorials are necessary. 

The act of writing them was useful and perhaps they will be useful to readers of the (increasingly novel and technical) code in #95.


Twon:
I think these are great, first 2 down and working through the 3rd now, but these are definitely useful (I'm learning reflection from the 3rd).

Perhaps the files could be renamed with a number indicating the order of the tutorials?

1. 1_type_erasure.cc
2. 2_vanishing_this_pointer
3. 3_reflection.cc

The pedant in me finds it odd to start in the middle, go down and then go up in the file view in VSCode.

tutorials/3_reflection.cc
```diff
@@ -0,0 +1,509 @@
...
+// ---------------------------------------------------------------------------
+namespace section_4 {
+
+struct Point {
+  int x = 0;
+  int y = 0;
+  int z = 0;
+};
+
+consteval std::vector<std::meta::info> data_members_of(std::meta::info type) {
```
Twon, line 258: May as well just use `std::vector`'s range constructor
```suggestion
consteval std::vector<std::meta::info> data_members_of(std::meta::info type) {
  return std::meta::nonstatic_data_members_of(
      type, std::meta::access_context::current());
}
```

tutorials/3_reflection.cc
```diff
@@ -0,0 +1,509 @@
...
+
+struct Widget {
+  int value() const { return 1; }
+
+  void set_value(int) {}
+
+  int payload = 0;
+};
+
+consteval std::vector<std::meta::info> member_functions_of(
```
Twon, line 282:
```c++
consteval std::vector<std::meta::info> member_functions_of(std::meta::info type) {
    // is_function alone would also match Widget's implicit special
    // members (default constructor, destructor, ...);
    // is_special_member_function excludes those, leaving just the
    // two ordinary methods below.
    auto const is_member_function = [](std::meta::info member) {
        return std::meta::is_function(member) && !std::meta::is_special_member_function(member);
    };

    return std::meta::members_of(type, std::meta::access_context::current()) |
           std::ranges::views::filter(is_member_function) |
           std::ranges::to<std::vector<std::meta::info>>();
}
```

tutorials/2_vanishing_this_pointer.cc
```diff
@@ -0,0 +1,227 @@
...
+
+  // owner.greet() calls that member's operator().
+  static_assert(std::is_class_v<decltype(owner.greet)>);
+  static_assert(!std::is_function_v<decltype(owner.greet)>);
+}
+
+}  // namespace section_3
+
+// ---------------------------------------------------------------------------
+// 4. What breaks when offset zero is violated, and the guard against it.
```
RyanJK5, line 168: Similar to my comment on the original PR, but this section should probably touch on the multiple-inheritance trick as well, so it can work regardless of whether or not the compiler applies no_unique_address.

tutorials/3_reflection.cc
```diff
@@ -0,0 +1,492 @@
...
+}
+
+TEST(ReflectionEnumToString, UnmatchedValueFallsThrough) {
+  // A value with no matching enumerator falls through every generated
+  // comparison to the fallback return, reached here via an explicit
+  // cast, since Color itself declares no such value.
+  EXPECT_EQ(enum_to_string(static_cast<Color>(99)), "<unnamed>");
+}
+
+}  // namespace section_7
```
RyanJK5, line 529: You might want to add a small section or sub-section somewhere on `substitute` and `extract`.

tutorials/2_vanishing_this_pointer.cc
```diff
@@ -0,0 +1,299 @@
...
+struct Owner {
+  Wrapper wrapper;  // first data member
+  int other_data = 0;
+};
+
+TEST(ThisPointerOffsetZero, FirstMemberSharesAddress) {
+  Owner owner;
+
+  EXPECT_EQ(static_cast<void*>(&owner), static_cast<void*>(&owner.wrapper));
+  EXPECT_EQ(offsetof(Owner, wrapper), 0u);
```
RyanJK5, line 60: Nitpick: the new reflection utilities give a new version `offsetof`, which looks like `offset_of(^^Owner::wrapper)`. It may be confusing to put that here before the tutorial on reflection though, so I understand if it's not really worth reordering the two sections just for that.

tutorials/2_vanishing_this_pointer.cc
```diff
@@ -0,0 +1,299 @@
...
+}  // namespace section_1
+
+// ---------------------------------------------------------------------------
+// 2. Recovering the owner pointer, "the vanishing this pointer" itself.
+//
+// Because a Wrapper sitting at offset zero of an Owner shares the Owner's
+// address (section 1), a method defined on Wrapper can compute a pointer to
+// its owner just by casting its own `this`, without Wrapper storing a
+// back-pointer anywhere.
+// ---------------------------------------------------------------------------
```
RyanJK5, line 93: It may be worth noting that this is undefined behavior unless `Wrapper` is exactly the first data member of `Owner`, and if `Owner` is a standard layout type. It may not be necessary to get in the weeds of what standard layout actually requires, but perhaps a brief mention.

tutorials/2_vanishing_this_pointer.cc
```diff
@@ -0,0 +1,299 @@
...
+// recovery, that call can reach and mutate the real owner.
+// ---------------------------------------------------------------------------
+namespace section_3 {
+
+struct GreetWrapper {
+  std::string operator()(std::string_view visitor) const;  // defined below
+};
+
+struct Owner {
+  GreetWrapper greet;
```
RyanJK5:
```suggestion
  [[no_unique_address]] GreetWrapper greet;
```

tutorials/2_vanishing_this_pointer.cc
```diff
@@ -0,0 +1,299 @@
...
+
+struct Wrapper {
+  int tag = 0;
+};
+
+// BadOwner puts another member before wrapper, so wrapper is no longer at
+// offset zero.
+struct BadOwner {
+  int preceding_member = 0;
+  Wrapper wrapper;
```
RyanJK5: `[[no_unique_address]]` doesn't change anything here, but may clarify the intent / be more consistent with the prior examples

tutorials/3_reflection.cc
```diff
@@ -0,0 +1,617 @@
...
+// compile time, such as what a type's members are or whether a function is
+// const, and so on. The reflection operator ^^ takes something in the program
+// (a type, a variable, a function, a member...) and lifts it into a
+// std::meta::info value: an opaque, compile-time value describing that
+// thing, distinct from the thing itself (^^int is a value you can pass
+// around, compare, and query; it is not int). This section only
+// establishes that fact; turning an info back into usable code
+// ("splicing") is sections 2 and 3.
+// ---------------------------------------------------------------------------
+namespace section_1 {
```
RyanJK5, line 57: I'm not sure if this is the best section to add this, but a quick piece discussing the following could be useful:

```c++
struct A {};
using B = A;

static_assert(^^A != ^^B);
static_assert(^^A == dealias(^^B));
```

I know it was one of the common gotchas I kept running into when I started working with reflection.

tutorials/3_reflection.cc
```diff
@@ -0,0 +1,617 @@
...
+  int x = 0;
+  int y = 0;
+};
+
+TEST(ReflectionSpliceType, RoundTripsToOriginal) {
+  constexpr std::meta::info reflection_of_int = ^^int;
+
+  // typename [:reflection_of_int:] names int again, so this line declares an
+  // ordinary int variable, just spelled through a splice.
+  typename[:reflection_of_int:] x = 5;
```
RyanJK5, line 140: I may be wrong, but based on the comment, I'm guessing the original code said `typename [:reflection_of_int:]` and then got clang-formatted into `typename[:reflection_of_int:]`. There's an [open issue](https://github.com/llvm/llvm-project/issues/192040) for this in the LLVM project.

Anyway, not saying this to suggest any changes here, except maybe changing the comment to match, but wanted to let you know.

tutorials/3_reflection.cc
```diff
@@ -0,0 +1,617 @@
...
+consteval std::meta::info find_member(std::meta::info type,
+                                      std::string_view name) {
+  for (std::meta::info member :
+       std::meta::members_of(type, std::meta::access_context::current())) {
+    if (std::meta::has_identifier(member) &&
+        std::meta::identifier_of(member) == name) {
+      return member;
+    }
+  }
+  return std::meta::info{};
```
RyanJK5: May want to add a quick note regarding what a default-constructed `std::meta::info` represents

tutorials/3_reflection.cc
```diff
@@ -0,0 +1,617 @@
...
+  constexpr std::meta::info reflected_function = ^^free_function;
+  constexpr std::meta::info reflected_member = ^^Point::x;
+
+  // Each ^^ result carries its own entity kind, queryable with a predicate
+  // like std::meta::is_namespace. What you can do with each (splice it as a
+  // type, call it, read its name) depends on this kind, covered starting in
+  // section 2.
+  static_assert(std::meta::is_namespace(reflected_namespace));
+  static_assert(std::meta::is_function(reflected_function));
+  static_assert(std::meta::is_nonstatic_data_member(reflected_member));
```
RyanJK5: This is relevant throughout the tutorial, and actually also the codebase, as a style question. The reflection proposal was designed so that meta-functions could be used with qualifying them with the `std::meta` namespace via ADL, so the following could be written as:

```c++
  static_assert(is_namespace(reflected_namespace));
  static_assert(is_function(reflected_function));
  static_assert(is_nonstatic_data_member(reflected_member));
```

Without having to `using namespace std::meta;`. It's a completely subjective stylistic choice, but I think it's a common enough pattern in reflection code that it may be worth addressing either here or somewhere in the readme / contribution guide.

tutorials/3_reflection.cc
```diff
@@ -0,0 +1,617 @@
...
+// std::meta::display_string_of is the primitive for turning a type into a
+// readable string.
+consteval std::string simple_signature_string(std::meta::info member) {
+  std::string result(std::meta::identifier_of(member));
+  result += "(";
+  bool first = true;
+  for (std::meta::info parameter : std::meta::parameters_of(member)) {
+    if (!first) result += ",";
+    first = false;
+    result += std::meta::display_string_of(
```
RyanJK5: Maybe add a note: `display_string_of` is implementation-defined and should not be used outside of debug output.

tutorials/3_reflection.cc
```diff
@@ -0,0 +1,617 @@
...
+// useful anywhere generated code needs a struct whose members (their count,
+// types, and names) aren't known until compile time, such as a vtable of
+// function pointers with one entry per member some other type happens to
+// declare.
+//
+// define_aggregate is called from inside a `consteval { ... }` block: a
+// statement, not a function, that always runs during translation regardless
+// of where it appears, even directly inside an ordinary, non-constexpr
+// function body, as the first test below does.
+// ---------------------------------------------------------------------------
```
RyanJK5, line 427: Worth noting that this is the only avenue we have for code generation currently, which is why a lot of the decisions for the implementation were made the way they were.

tutorials/3_reflection.cc
```diff
@@ -0,0 +1,617 @@
...
+
+struct Point {
+  int x = 0;
+  int y = 0;
+  int z = 0;
+};
+
+consteval std::vector<std::meta::info> data_members_of(std::meta::info type) {
+  return std::meta::nonstatic_data_members_of(
+      type, std::meta::access_context::current());
```
RyanJK5: A brief explanation of `access_context` would be helpful, touching on `unprivileged` and `unchecked`, and why you might want to use a particular context.

tutorials/3_reflection.cc
RyanJK5, line 427: Do you think you could place the definition for `data_member_options` here? Adapted from [P2996](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p2996r13.html):

```c++
struct data_member_options {
    optional<name_type> name; // name_type is a string
    optional<int> alignment;
    optional<int> bit_width;
    bool no_unique_address = false;
  };
```

tutorials/3_reflection.cc
```diff
@@ -0,0 +1,617 @@
...
+// A template specialization built by splicing needs one [:e:] slot per
+// argument, each already sitting at a fixed position in source, e.g.
+// Boxed<typename[:int_info:]>. That breaks down when the argument list
+// itself is only known as a range of infos computed at compile time.
+// std::meta::substitute(template_info, argument_infos) instantiates a class,
+// alias, or function template directly from such a range instead, as an
+// ordinary consteval function call, and returns an info reflecting the
+// resulting specialization. Splicing that result, as in section 2, turns it
+// into a usable type.
+// ---------------------------------------------------------------------------
```
RyanJK5, line 545: This could perhaps use a bit more elaboration on *why* you can't just splice `e` directly (because `e` is not a constant expression, even though it's being run at compile-time).

tutorials/3_reflection.cc
```diff
@@ -0,0 +1,617 @@
...
+
+template <typename T>
+struct Boxed {
+  T value = T();
+};
+
+TEST(ReflectionSubstitute, InstantiatesTemplateFromInfos) {
+  constexpr std::meta::info boxed_int_info =
+      std::meta::substitute(^^Boxed, {
+                                         ^^int});
```
RyanJK5: Related to comment on formatting from above: an example of clang-format misunderstanding the `^^` operator as well

tutorials/3_reflection.cc
```diff
@@ -0,0 +1,617 @@
...
+  std::meta::info member_type = std::meta::type_of(find_member(type, name));
+  std::meta::info extent = std::meta::template_arguments_of(member_type)[1];
+  return std::meta::extract<size_t>(extent);
+}
+
+TEST(ReflectionExtract, ReadsANonTypeTemplateArgument) {
+  static_assert(member_array_extent(^^Row, "values") == 4);
+}
+
+}  // namespace section_9
```
RyanJK5, line 661: Two additional things you might consider for this tutorial:

A brief mention (maybe in an already existing section?) of using `reflect_constant` to substitute a constant template parameter, e.g.

```c++
template <int X>
struct S {};

substitute(^^S, { reflect_constant(5) });
```

The other is the new `[[=annotations]]` feature and the related meta-functions. This may make more sense to omit unless the implementation ends up using them.

RyanJK5 (commented):
I just gave the tutorials an in-depth look, and I think they are very good! All of the information is accurate. I left some comments, mostly regarding additional pieces of the reflection tutorial worth acknowledging. Thanks for putting these together!

jbcoe:
@RyanJK5 thanks for the detailed review comments. This is exactly what’s needed for foundational material. Thanks for taking the time.

tutorials/3_reflection.cc
jbcoe: I'll skip `[[=annotations]]` for now but feel free to add in a follow-up PR.

As you say, we're not planning to use them for now.

tutorials/3_reflection.cc
jbcoe: I'm happy to leave this while clang-format does not support reflection.

tutorials/3_reflection.cc
jbcoe: It's ugly but manual formatting feels too onerous and a probable source of such churn so I'm happy to suffer the bad formatting (code is not broken by it) until clang-format is updated.

tutorials/2_vanishing_this_pointer.cc
jbcoe: I'll keep the tutorials decoupled and minimise the learning curve gradient. 

Thanks for pointing this out though.

RyanJK5 (approved):
Looks good to me now, thanks for making those changes!

tutorials/3_reflection.cc
```diff
@@ -0,0 +1,658 @@
...
+// does not, so private members stay invisible to it, just as they would to
+// ordinary code written at this call site). access_context::unprivileged()
+// always restricts the query to public members, however privileged the
+// calling code actually is, useful for reflecting on a type the way an
+// arbitrary external caller would see it. access_context::unchecked()
+// removes access checking entirely, useful for tooling that must reach
+// every member regardless of access, such as a serializer.
+consteval std::meta::info find_member(std::meta::info type,
+                                      std::string_view name) {
+  for (std::meta::info member :
```
Twon: Again, I'd like to encourage the modern "no raw loops" style
```suggestion
    auto const members = members_of(type, std::meta::access_context::current());
  auto const is_member = [&](std::meta::info member) {
    return has_identifier(member) && identifier_of(member) == name;
  };

  if (auto const member = std::ranges::find_if(members, is_member);
      member != members.end()) {
    return *member;
  }
  // A default-constructed std::meta::info is a valid, comparable sentinel value
  // that specifically means "nothing here," rather than describing any real
  // type, function, variable, or namespace.
  return {};
```

jbcoe: I’m not a ranges-native and find this pretty hard to follow compared to the original.

happy to defer to your judgement if you feel strongly.

Twon: Interesting, with native loops I find I have to parse the code and to work out what it does, but with ranges it can just read the keyword, i.e. this whole function is just a "find if it's a member"

But I don't feel strongly enough about it, particularly if other's experience is the opposite.

jbcoe: If the readability is subjective too, I vote we go ranges. Let me amend and update.

tutorials/3_reflection.cc
```diff
@@ -0,0 +1,658 @@
...
+    if (is_function(member) && has_identifier(member) &&
+        identifier_of(member) == name) {
+      result.push_back(member);
+    }
+  }
+  return result;
+}
+
+TEST(ReflectionHelpers, ClassifiesConstCorrectly) {
+  constexpr auto read_candidates =
```
Twon: Personal I'd perfer concept constrained auto here because being unfamilar with this function, and with no explanation of its use, I had to go to [C++ ref](https://en.cppreference.com/cpp/meta/define_static_array) and read the documentation to be confident I know what was happening.  I think the hint is a range is enough knowledge to skip this step
```suggestion
  constexpr std::ranges::range auto read_candidates =
```

tutorials/3_reflection.cc
```diff
@@ -0,0 +1,658 @@
...
+      std::define_static_array(members_named(^^Widget, "read"));
+  constexpr auto write_candidates =
```
Twon: Again concept-constrained auto
```suggestion
  constexpr std::ranges::range auto write_candidates =
```

tutorials/3_reflection.cc
```diff
@@ -0,0 +1,658 @@
...
+// readable string. The output of std::meta::display_string_of is
+// implementation-defined and should not be used outside of debug output.
+consteval std::string simple_signature_string(std::meta::info member) {
+  std::string result(identifier_of(member));
+  result += "(";
+  bool first = true;
+  for (std::meta::info parameter : parameters_of(member)) {
+    if (!first) result += ",";
+    first = false;
+    result += display_string_of(dealias(type_of(parameter)));
```
Twon, line 371: Ashame we don't have dynamic delimiters for range formatting yet.  I feel like this is the line the magic happens on, but I have to wade through additional cognitive noise to get the magic... particularly in light of the example going on to then only use a single parameter per function.  Perhaps add a multi-parameter overload or just simplify for one param

tutorials/3_reflection.cc
```diff
@@ -0,0 +1,658 @@
...
+  }
+  result += ")";
+  return result;
+}
+
+TEST(ReflectionHelpers, SignatureDistinguishesOverloads) {
+  constexpr auto write_candidates =
```
Twon: I'm not so worried about adding concept-constrained auto once the user is exposed to how this method works from earlier examples

tutorials/3_reflection.cc
```diff
@@ -0,0 +1,661 @@
...
+// function body, as the first test below does.
+//
+// In C++26, define_aggregate is the only avenue available for code generation.
+// ---------------------------------------------------------------------------
+namespace section_6 {
+
+// Adapted from
+// https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p2996r13.html
+//
+// struct data_member_options {
```
Twon, line 432: I had to go to C++ ref to understand the significance of this (in part because there is no indication this is a std::meta type). I initially thought this was an example of a struct you were going to generate... Particularly as in count_ratio_specs it's constructed with {}.

tutorials/3_reflection.cc
```diff
@@ -0,0 +1,661 @@
...
+//     optional<name_type> name; // name_type is a string
+//     optional<int> alignment;
+//     optional<int> bit_width;
+//     bool no_unique_address = false;
+// };
+
+consteval std::vector<std::meta::info> count_ratio_specs() {
+  std::vector<std::meta::info> specs;
+  // clang-format off
+  specs.push_back(data_member_spec(^^int, {.name = "count"}));
```
Twon: This, for me, would connect the significance of the `std::meta::data_member_options` definition above and tell me it's a std::meta type.
```suggestion
  specs.push_back(data_member_spec(^^int, data_member_options{.name = "count"}));
```

tutorials/3_reflection.cc
jbcoe: If you need to look at another tutorial to understand this one, that's a BAD sign. 

I'll amend the use of `auto` here and elsewhere.

tutorials/3_reflection.cc
jbcoe: How can I make this clearly useful? 

@RyanJK5, I added this comment in response to your suggestion but I think the reader is still missing some context. What should I add?

RyanJK5: > How can I make this clearly useful?
> 
> @RyanJK5, I added this comment in response to your suggestion but I think the reader is still missing some context. What should I add?

I think the best way would to add a comment before the struct definition explaining what it is. Combining that with the explicit `data_member_options` in the lines below that @Twon mentioned in the other thread, it should convey the significance.

jbcoe:
This is superseded by other merged or in-flight PRs: #146, #141, #133.

PR #99, opened 2026-07-26
Add gh and jj to the dev-environment Docker image

jbcoe:
Both are needed for GitHub PR/issue work and Jujutsu-based history inspection. gh installs from its own apt repository (not reliably available via Ubuntu 24.04's default repos); jj installs from its pinned GitHub release binary, matching the existing bazelisk pattern.


jbcoe:
Works locally. Any approval is good!

philipcraig (approved):

PR #103, opened 2026-07-27
Add a default allocator argument to the protocol primary template (3/20)

jbcoe:
Moves the std::allocator<T> default from the definition to the
forward declaration, so protocol<T> is valid wherever only the
declaration is visible. A default can't be repeated on the same
parameter, so the definition drops its own.

### DIFF
--- a/protocol.h
+++ b/protocol.h
@@ -29,7 +29,7 @@ namespace xyz {
 template <typename T>
 struct is_protocol : std::false_type {};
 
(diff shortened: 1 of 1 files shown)


RyanJK5 (approved):

jbcoe:
Superseded by a fresh PR: rebased directly onto main and renamed off the reflection/NN stack naming.

PR #104, opened 2026-07-27
Add XYZ_PROTOCOL_USE_REFLECTION cmake option to gate the reflection backend (4/20)

jbcoe:
Reuses the existing GCC16+/-freflection toolchain probe rather than
adding a second option, since the new protocol_reflection.h backend
needs the same compiler as the tutorial. Named USE_REFLECTION from
the start rather than introduced as BUILD_REFLECTION_TUTORIAL and
renamed, since it gates the backend's own targets too, not just the
tutorial; no new targets yet.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,12 +30,14 @@ option(ENABLE_TSAN "Enable Thread Sanitizer" OFF)
 option(ENABLE_MSAN "Enable Memory Sanitizer" OFF)
 
 option(
-  XYZ_PROTOCOL_BUILD_REFLECTION_TUTORIAL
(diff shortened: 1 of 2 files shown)


RyanJK5:
It may be desirable to retarget the Docker build on gcc-trunk for better long-term support. There are some bugs with GCC 16 reflection that have since been resolved, e.g.

```c++
#include <meta>

template <typename T>
struct S {};

// Error on GCC 16: non-constant condition for static assertion
// Compiles on GCC trunk
static_assert(substitute(^^S, { ^^int&()}) == ^^S<int&()>); 
```

jbcoe:
@RyanJK5 that sounds like a good idea. Can you raise an issue?

PR #105, opened 2026-07-27
Add naming.hxx: injective vtable/forwarder entry-name escaping (5/20)

jbcoe:
identifier_safe_string escapes every non-alphanumeric byte, including
underscore, into an underscore-plus-hex pair: a bijection on byte
sequences, so it's collision-free by construction rather than only
probabilistically. Updates GEMINI.md's naming description to match.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -201,6 +201,22 @@ if(XYZ_PROTOCOL_IS_NOT_SUBPROJECT)
         FILES
         tutorials/3_reflection.cc)
       target_compile_options(reflection_tutorial PRIVATE -freflection)
+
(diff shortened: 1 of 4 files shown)


RyanJK5:
Is this to be used for creating vtable entries? In order for a function to be added to the vtable, it must be defined in an Interface struct first by the user, which implies it is already a valid identifier. In that case, what does `identifier_safe_string` protect against?

jbcoe:
This is to support operator() operator+ and the like.

We can perhaps simplify the chain by solving problems only as and when they arise. It will simplify the review.

I’ll re-jig the chain when time permits.

jbcoe:

PR #106, opened 2026-07-27
Add members.hxx: enumerating an interface's dispatchable member functions (6/20)

jbcoe:
is_interface_member_function/interface_member_functions enumerate a
type's public, non-static, non-template, non-special member functions,
including operators, in declaration order.


jbcoe:

PR #108, opened 2026-07-27
Add vtable_layout.hxx: view_vtable/const_view_vtable for a bare interface (8/20)

jbcoe:
Synthesizes per-interface vtable structs via define_aggregate,
matching the manual backend's const_view_vtable_X/view_vtable_X
split. define_aggregate can only complete a type declared in the same
scope as the consteval block completing it; worked around with a
consteval function template holding a local incomplete struct.


jbcoe:

PR #110, opened 2026-07-27
Add conformance.hxx: duck-typed candidate merging and the reflection concepts (10/20)

jbcoe:
For each interface member, finds same-named implementation
candidates, merges them into one overload-resolved callable via
using Forwarders::operator()... (mirroring rjk::duck), and requires
the merge callable with the member's exact signature. Defines
reflection_protocol_concept/reflection_protocol_const_concept on top
of that check.


jbcoe:

PR #112, opened 2026-07-27
Wire protocol_reflection.hxx: allocator-free single-method dispatch on Greeter (12/20)

jbcoe:
protocol<T, Allocator> gets a real generic definition under
XYZ_PROTOCOL_ENABLE_REFLECTION: one class template, built from
forwarders.hxx's combinator and a define_vtable_entries-built vtable,
working for any interface satisfying reflection_protocol_concept
with no per-interface codegen. Scope is deliberately narrow:
default-allocator construction and copy/move/destroy/swap only.


jbcoe:

PR #114, opened 2026-07-27
Add overload merging: duck-typed dispatch for overloaded members (C) (14/20)

jbcoe:
vtable_entry_name (identifier-only) names a forwarder's public-facing
data member, so overloads of the same name stay merged into one
callable; vtable_slot_name (signature-qualified, via
display_string_of) is added for the internal vtable struct, which
needs one distinct entry per exact overload instead.


jbcoe:

PR #116, opened 2026-07-27
Add allocator-awareness (16/20)

jbcoe:
protocol<T, Allocator> gains the allocator-extended constructor set
plus allocator-aware assignment and swap, matching the
Python/libclang backend's protocol_A.h: equal allocators steal the
source's pointer, unequal ones move-construct into freshly allocated
storage.


jbcoe:

PR #118, opened 2026-07-27
Add protocol_view<T> and protocol_view<const T> (18/20)

jbcoe:
Non-owning views dispatching through a raw pointer/vtable pair
instead of protocol's allocator-owned storage, reusing the same
forwarder-combinator and thunk machinery; the const view exposes
only const members, through a const void*.


jbcoe:

PR #120, opened 2026-07-27
Swap protocol_test.cc to the reflection backend; enable protocol_test_reflection (20/20)

jbcoe:
Adds protocol_test_reflection, running protocol_test.cc against
protocol_reflection.hxx instead of the generated headers. Every test
passes except interface D's two operator tests, excluded until
operator support lands.


jbcoe:

PR #125, opened 2026-08-06
Add a default allocator argument to the protocol primary template

jbcoe:
Moves the std::allocator<T> default from the definition to the
forward declaration, so protocol<T> is valid wherever only the
declaration is visible. A default can't be repeated on the same
parameter, so the definition drops its own.

### DIFF
--- a/protocol.h
+++ b/protocol.h
@@ -29,7 +29,7 @@ namespace xyz {
 template <typename T>
 struct is_protocol : std::false_type {};
 
-template <typename T, typename Alloc>
+template <typename T, typename Alloc = std::allocator<T>>
 class protocol;
 
 template <typename T, typename Alloc>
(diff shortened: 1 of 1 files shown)


protocol.h
```diff
@@ -123,7 +123,7 @@ get_owning_vtable(const typename protocol_owning_vtable_traits<
                         sizeof(ToVtable), mapping_function));
 }
 
-template <typename T, typename A = std::allocator<T>>
+template <typename T, typename A>
```
RyanJK5, line 126: nit: change `typename A` to `typename Alloc` or `typename Allocator` to match the rest of the codebase

Twon (approved):

protocol.h
jbcoe: We should do this.

PR #126, opened 2026-08-07
Simple tests for supported special member functions

jbcoe:
First step in adding a reflection-based implementation of protocol (and protocol_view). 

No reflection features are currently needed, just C++ 20 concepts and requirement checks to conditionally generate constructors.

Once this is landed we can extend the tests to document and test-drive a full implementation.

`GTEST_SKIP() << "Skipping test";` can be used to skip tests while we are waiting to add implementation support.

Once a complete implementation has been submitted, we can retire the Python implementation.


reflection/protocol.h
```diff
@@ -0,0 +1,62 @@
...
+    requires std::is_move_constructible_v<T>;
+
+  protocol& operator=(const protocol&)
+    requires std::is_copy_assignable_v<T>;
+
+  protocol& operator=(protocol&&)
+    requires std::is_move_assignable_v<T>;
+
+  ~protocol();  // Unconstrained.
+};
```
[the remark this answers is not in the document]
jbcoe: This is intended. Static analysis checks are all that is in scope for this PR.

jbcoe:
@copilot Fix the failing tests by restoring the missing definition of `struct A {};`

reflection/protocol.h
```diff
@@ -0,0 +1,65 @@
...
+#include <memory>
+#include <type_traits>
+
+namespace xyz::reflection {
+
+template <typename T, typename Allocator = std::allocator<std::byte>>
+class protocol {
+ public:
+  // Special member functions.
+  protocol() = delete;
```
RyanJK5: Does an interface with a defaulted copy constructor imply the `protocol` object should be default constructible? e.g.

```c++
struct I {
    I() = default;
};

protocol<I> p{}; // Now has a valid empty state
```

I thought I noticed something like that in the original implementation, but I may very well be mistaken.

reflection/protocol.h
```diff
@@ -0,0 +1,65 @@
...
+    requires std::is_copy_constructible_v<T>;
+
+  protocol(protocol&&)
+    requires std::is_move_constructible_v<T>;
+
+  protocol& operator=(const protocol&)
+    requires std::is_copy_assignable_v<T>;
+
+  protocol& operator=(protocol&&)
+    requires std::is_move_assignable_v<T>;
```
RyanJK5: Is there a design reason for why an Interface might disable moves? Granted, the underlying type may not be movable, but `protocol` can still be moved regardless of that, since it can just store the data on the heap and do a simple pointer swap when being moved. It would disable SBO though.

reflection/protocol.h
```diff
@@ -0,0 +1,65 @@
...
+  protocol() = delete;
+
+  protocol(const protocol&)
+    requires std::is_copy_constructible_v<T>;
+
+  protocol(protocol&&)
+    requires std::is_move_constructible_v<T>;
+
+  protocol& operator=(const protocol&)
+    requires std::is_copy_assignable_v<T>;
```
RyanJK5: What is the reasoning for assignment and construction to be different things in the interface? While it may be possible for a type to support copy construction and not assignment or vice versa, does that necessarily have to restrict `protocol`?

RyanJK5 (commented):
Looks good overall 👍

reflection/protocol.h
jbcoe: You’re quite right. We should follow the design of `polymorphic` for all of the constructors and special member functions.

reflection/protocol.h
jbcoe: Again, we should update this to follow `polymorphic`.

jbcoe:
@copilot look at https://eel.is/c++draft/mem.composite.types and update protocol constructor requirements to model the class template `polymorphic`

reflection/protocol.h
RyanJK5: I think the changes may have removed the ability for the interface to control copyability, however, which is probably a useful thing to have (e.g. `move_only_function` versus `copyable_functoin`)

jbcoe: This is totally unintended. If you have fixes, please push them to the branch.

This feels like a minimal unit of design but does need to (ideally) land in a sane state.

RyanJK5: I pushed changes to match my understanding of how it should work, but please correct me if I have the design wrong. I think it does make sense to restrict copy assignment and construction separately since there are some niche cases where a type might define one and not the other (e.g. reference members).

RyanJK5:
I also pushed a change to add `noexcept` to `protocol_view`'s move constructor and assignment, I can roll it back if that's unwanted. I withheld adding it to `protocol` since according to my understanding, that will depend on how the allocator type specifies POCMA and `is_always_equal`.

jbcoe:
@RyanJK5 Looks good. As a first step this is easy to approve. We can fine-tune in later PRs.

reflection/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+xyz_add_library(
+  NAME reflection_protocol
+  ALIAS xyz_protocol::reflection_protocol)
+target_sources(
+  reflection_protocol INTERFACE
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/protocol.h>)
```
Twon, line 6: Adding ${CMAKE_CURRENT_SOURCE_DIR} to files in the current directory is only necessary for CMake version proior to 3.13, for details see: https://crascit.com/2016/01/31/enhanced-source-file-handling-with-target_sources/ 

```suggestion
  $<BUILD_INTERFACE:protocol.h>)
```

Twon, line 6: I'd expect to see an entry for `INSTALL_INTERFACE` but it looks like we don't have the install step set up for this project yet, we should open a ticket to track that so this can be used by other projects as an installed library.

reflection/protocol_test.cc
```diff
@@ -0,0 +1,97 @@
...
+
+  static_assert(std::is_default_constructible_v<protocol<A>>);
+  static_assert(std::is_copy_constructible_v<protocol<A>>);
+  static_assert(std::is_move_constructible_v<protocol<A>>);
+  static_assert(std::is_copy_assignable_v<protocol<A>>);
+  static_assert(std::is_move_assignable_v<protocol<A>>);
+}
+
+TEST(RProtocolTest, CheckSpecialMembersForStructWithDeletedSpecialMembers) {
+  // protocol's default constructor depends on the interface, but its copy and
```
Twon: Surely this should match the static assets and be:
```suggestion
  // protocol's default constructor, and copy construction and assignment depend on the interface, but its
```

Twon (commented):
A few minor comments but otherwise looks good

reflection/protocol.h
jbcoe: This needs a rethink but for now I think should be suppressed.

jbcoe: My thinking here is that for polymorphic, interfaces are often pure virtual whereas the design of protocol is to use structs without implementations of functions. Requiring the user to explicitly delete default construction seems unhelpful and later constructors (in place type) will allow construction from any conformant type.

RyanJK5: Makes sense, and also possible to change later if need be. It would also complicate matters, since in the default constructor case, `valueless_after_move` would be a misnomer (it could have been empty because it was constructed that way). `duck` similarly avoids the complexity by disallowing default construction.

reflection/CMakeLists.txt
jbcoe: The amended code gives me

```
CMake Error in reflection/CMakeLists.txt:
  Target "reflection_protocol" contains relative path in its
  INTERFACE_SOURCES:

    "protocol.h"
```

We've used the same pattern in CMakeLists.txt at the project root. Perhaps they can all be cleaned up in sync later?

jbcoe: I raised https://github.com/jbcoe/cc-protocol/issues/132. We're not ready to be installed yet.

reflection/protocol_test.cc
jbcoe: This is reworked now to be more in sync with polymorphic. In any case code should reflect the comments.

RyanJK5:
I think it is good to merge now, and we can start seeing how this scales as we add the reflection implementation.

If there's anything I can help with outside of reviewing, i.e. start coding up a piece of the implementation, I would be more than happy to.

jbcoe:
> I think it is good to merge now, and we can start seeing how this scales as we add the reflection implementation.
> 
> If there's anything I can help with outside of reviewing, i.e. start coding up a piece of the implementation, I would be more than happy to.

Thanks @RyanJK5 i think the main constraint right now is time on our side to code things up at a reviewable pace so that everyone follows what’s going on.

Once a minimal reflection implementation is in place we can start looking into extensions and design details.

RyanJK5:
> > I think it is good to merge now, and we can start seeing how this scales as we add the reflection implementation.
> 
> > 
> 
> > If there's anything I can help with outside of reviewing, i.e. start coding up a piece of the implementation, I would be more than happy to.
> 
> 
> 
> Thanks @RyanJK5 i think the main constraint right now is time on our side to code things up at a reviewable pace so that everyone follows what’s going on.
> 
> 
> 
> Once a minimal reflection implementation is in place we can start looking into extensions and design details.

Got it. I don't mean to suggest things are moving too slow, but if we find that part of the bottleneck is deciding what minimal component to take on next, I can certainly help on coding and deciding that to open more time for reviewing.

reflection/protocol_test.cc
```diff
@@ -0,0 +1,77 @@
...
+    D(const D&) = delete;  // Suppresses copy and assigment for `protocol<D>`.
+    D(D&&) = default;      // Ignored.
+    D& operator=(const D&) = default;  // Ignored.
+    D& operator=(D&&) = default;       // Ignored.
+  };
+
+  static_assert(!std::is_default_constructible_v<protocol<D>>);
+  static_assert(!std::is_copy_constructible_v<protocol<D>>);
+  static_assert(std::is_move_constructible_v<protocol<D>>);
+  static_assert(std::is_move_constructible_v<protocol<D>>);
```
RyanJK5:
```suggestion
```

Just a duplicate line here

RyanJK5 (commented):
Just noticed this one duplicate line - looks good other than that

RyanJK5 (approved):

PR #133, opened 2026-08-12
Add a tutorial on polymorphism

### LINKED ISSUE #138: Add a tutorial on polymorphism
A minimal tutorial covering the techniques we use for protocol and general C++ polymorphism will benefit both authors and readers.

The tutorial in https://github.com/jbcoe/cc-protocol/pull/96 needs reworking but could be used for inspiration.

jbcoe:
Implement polymorphism with templates, inheritance and type-erasure.

closes #138

### DIFF
--- a/tutorials/polymorphism.cc
+++ b/tutorials/polymorphism.cc
@@ -50,9 +50,6 @@ TEST(TutorialsPolymorphism, PolymorphismWithTemplates) {
   Cat cat;
   Dog dog;
 
-  // Cat and Dog have no common type so we can't store them directly in a
-  // container. The function template can be used on either instance.
-
   EXPECT_EQ(make_noise(cat), "Meow");
   EXPECT_EQ(make_noise(dog), "Woof");
 }
--- a/tutorials/polymorphism.cc
+++ b/tutorials/polymorphism.cc
@@ -27,7 +27,7 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 // Polymorphism with templates (compile-time polymorphism).
 
-// Cat and Dog have no common type but we can use compile-time polymorphism
+// `Cat` and `Dog` have no common type but we can use compile-time polymorphism
 // (templates) to write a function template to talk to both types.
 namespace xyz::tutorials::compile_time_polymorphism {
 
@@ -58,10 +58,10 @@ TEST(TutorialsPolymorphism, PolymorphismWithTemplates) {
 
 // Polymorphism with inheritance
 
-// Cat and Dog have a common type. We can use a single function for both types.
-// This form of inheritance is intrusive. Cat and Dog must declare Animal as a
-// base class and Animal must declare which member functions can be used
-// polymorphically.
+// `Cat` and `Dog` have a common type. We can use a single function for both
+// types. This form of inheritance is intrusive: `Cat` and `Dog` must declare
+// `Animal` as a base class at class-definition time and `Animal` must declare
+// which member functions can be used polymorphically.
 namespace xyz::tutorials::polymorphism_with_inheritance {
 
 class Animal {
@@ -73,14 +73,14 @@ class Animal {
 
 class Cat : public Animal {
  public:
-  std::string_view noise() const { return "Meow"; }
+  std::string_view noise() const override { return "Meow"; }
 
   std::string_view identity() const { return "Cat"; }  // non-virtual
 };
 
 class Dog : public Animal {
  public:
-  std::string_view noise() const { return "Woof"; }
+  std::string_view noise() const override { return "Woof"; }
 
   std::string_view identity() const { return "Dog"; }  // non-virtual
 };
@@ -91,12 +91,12 @@ TEST(TutorialsPolymorphism, PolymorphismWithInheritance) {
   Cat cat;
   Dog dog;
 
-  // We can store pointers to Cat and Dog as pointers to Animal.
+  // We can store pointers to `Cat` and `Dog` as pointers to `Animal`.
   std::vector<Animal*> animals;
   animals.reserve(2);
   animals.push_back(&cat);
   animals.push_back(&dog);
-  // Note: std::polymorphic could be used to allow the vector to own the
+  // Note: `std::polymorphic` could be used to allow the vector to own the
   // animals.
 
   EXPECT_EQ(make_noise(*animals[0]), "Meow");
@@ -111,9 +111,10 @@ TEST(TutorialsPolymorphism, PolymorphismWithInheritance) {
 
 // Polymorphism with type-erasure.
 
-// Cat and Dog have no common type; we can define a type-erasing wide-pointer to
-// determine which member functions to call at run time.
-
+// `Cat` and `Dog` have no common type; we can define a type-erasing
+// wide-pointer to determine which member functions to call at run time. The
+// decision to dispatch functions polymorphically is made by the `AnimalPtr`
+// class, not `Cat` or `Dog` classes.
 namespace xyz::tutorials::polymorphism_with_type_erasure {
 
 class Cat {
@@ -141,14 +142,14 @@ class AnimalPtr {
   std::string_view noise() const { return noise_func_(data_); }
 };
 
-// `animal` is passed by value as AnimalPtr is small like `std::string_view`.
+// `AnimalPtr` is small like `std::string_view` so `animal` is passed by value.
 std::string_view make_noise(AnimalPtr animal) { return animal.noise(); }
 
 TEST(TutorialsPolymorphism, PolymorphismWithTypeErasure) {
   Cat cat;
   Dog dog;
 
-  // We can store pointers to Cat and Dog using our wide pointer type.
+  // We can store pointers to `Cat` and `Dog` using our wide pointer type.
   std::vector<AnimalPtr> animals;
   animals.reserve(2);
   animals.emplace_back(&cat);
@@ -161,8 +162,8 @@ TEST(TutorialsPolymorphism, PolymorphismWithTypeErasure) {
 }  // namespace xyz::tutorials::polymorphism_with_type_erasure
 
 // When we have multiple functions, we can use a vtable to dispatch them.
-// Note that the decision to dispatch functions polymorphically is made by the
-// AnimalPtr class, not Cat or Dog classes.
+// Storing the vtable as a pointer ensures that `AnimalPtr` is cheap to copy,
+// regardless of how many member functions we want to resolve at run time.
 namespace xyz::tutorials::polymorphism_with_type_erasure_vtable {
 
 class Cat {
@@ -191,6 +192,8 @@ class AnimalPtr {
  public:
   template <typename T>
   AnimalPtr(T* t) : data_(t) {
+    // Making the vtable `constexpr`avoids runtime cost (synchronization between
+    // threads) when accessing static variables.
     constexpr static vtable vtable_for_type = {
         .noise_func_ =
             +[](const void* data) {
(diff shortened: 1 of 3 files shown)


tutorials/polymorphism.cc
```diff
@@ -0,0 +1,296 @@
...
+TEST(TutorialsPolymorphism, PolymorphismWithTemplates) {
+  Cat cat;
+  Dog dog;
+
+  // Cat and Dog have no common type so we can't store them directly in a
+  // container. The function template can be used on either instance.
+
+  EXPECT_EQ(make_noise(cat), "Meow");
+  EXPECT_EQ(make_noise(dog), "Woof");
+}
```
RyanJK5, line 55: Important to note: we cannot do `std::vector<Animal*>` or anything of that sort to store related types together. This approach only really works at an interface entry point (but, of course, is typically the most efficient option for that case).

tutorials/polymorphism.cc
```diff
@@ -0,0 +1,296 @@
...
+class Animal {
+ public:
+  virtual std::string_view noise() const = 0;
+
+  std::string_view identity() const { return "Animal"; }  // non-virtual
+};
+
+class Cat : public Animal {
+ public:
+  std::string_view noise() const { return "Meow"; }
```
RyanJK5:
```suggestion
  std::string_view noise() const override { return "Meow"; }
```

tutorials/polymorphism.cc
```diff
@@ -0,0 +1,296 @@
...
+
+  std::string_view identity() const { return "Cat"; }  // non-virtual
+};
+
+class Dog : public Animal {
+ public:
+  std::string_view noise() const { return "Woof"; }
```
RyanJK5:
```suggestion
  std::string_view noise() const override { return "Woof"; }
```

tutorials/polymorphism.cc
```diff
@@ -0,0 +1,296 @@
...
+  AnimalPtr(T* t) : data_(t) {
+    noise_func_ = +[](const void* data) -> std::string_view {
+      return static_cast<const T*>(data)->noise();
+    };
+  }
+
+  std::string_view noise() const { return noise_func_(data_); }
+};
+
+std::string_view make_noise(const AnimalPtr& animal) { return animal.noise(); }
```
RyanJK5:
```suggestion
std::string_view make_noise(AnimalPtr animal) { return animal.noise(); }
```
Much like `string_view`, I think we should encourage pass-by-copy for our non-owning types (the same will apply to `protocol_view`). If you think it's helpful, maybe a note about why it's cheap to copy could be included as well.

tutorials/polymorphism.cc
```diff
@@ -0,0 +1,296 @@
...
+ public:
+  std::string_view noise() const { return "Woof"; }
+
+  std::string_view identity() const { return "Dog"; }
+};
+
+struct vtable {
+  std::string_view (*noise_func_)(const void* data);
+  std::string_view (*identity_func_)(const void* data);
+};
```
RyanJK5, line 188: Something worth noting, though it may not be necessary to show in the code - this also can store the necessary lifecycle functions, like `copy`, `move`, and `destroy`, for an owning erased type.

tutorials/polymorphism.cc
```diff
@@ -0,0 +1,296 @@
...
+    static vtable vtable_for_type = {
+        .noise_func_ =
+            +[](const void* data) {
+              return static_cast<const T*>(data)->noise();
+            },
+        .identity_func_ =
+            +[](const void* data) {
+              return static_cast<const T*>(data)->identity();
+            },
+    };
```
RyanJK5:
```suggestion
    constexpr static vtable vtable_for_type = {
        .noise_func_ =
            +[](const void* data) {
              return static_cast<const T*>(data)->noise();
            },
        .identity_func_ =
            +[](const void* data) {
              return static_cast<const T*>(data)->identity();
            },
    };
```

Best practice to make this `constexpr`, since `static` variables can have some significant hidden runtime impact (especially non-`const` ones).

tutorials/polymorphism.cc
```diff
@@ -0,0 +1,296 @@
...
+  vtable* vtable_;
+
+ public:
+  std::string_view noise() const { return vtable_->noise_func_(this); }
+};
+
+class Cat : public Animal {
+ public:
+  Cat() {
+    static vtable vtable_for_type = {
```
RyanJK5:
```suggestion
    constexpr static vtable vtable_for_type = {
```

tutorials/polymorphism.cc
```diff
@@ -0,0 +1,296 @@
...
+    vtable_ = &vtable_for_type;
+  }
+
+  std::string_view noise() const { return "Meow"; }
+};
+
+class Dog : public Animal {
+ public:
+  Dog() {
+    static vtable vtable_for_type = {
```
RyanJK5:
```suggestion
    constexpr static vtable vtable_for_type = {
```

tutorials/polymorphism.cc
```diff
@@ -0,0 +1,296 @@
...
+
+    vtable_ = &vtable_for_type;
+  }
+
+  std::string_view noise() const { return vtable_->noise_func_(data_); }
+
+  std::string_view identity() const { return vtable_->identity_func_(data_); }
+};
+
+std::string_view make_noise(const AnimalPtr& animal) { return animal.noise(); }
```
RyanJK5:
```suggestion
std::string_view make_noise(AnimalPtr animal) { return animal.noise(); }
```

Storing a `vtable*` instead of two function pointers makes this copy cheap, regardless of how many functions the interface has. Directing through the vtable adds some runtime overhead, though (I'm not sure if that's necessary to discuss here).

RyanJK5 (changes requested):
The code itself is good, I left a few comments. I think one big thing the tutorial is missing is "why?", since this is fundamentally the motivation for `protocol<T>`. Point out that inheritance is intrusive, and the problems with that, but that manual type erasure requires more boilerplate for each interface you need - compounded even more so if you want an owning type. `protocol` combines the benefits of both approaches. Not sure if that works best at the start or end of this tutorial.

tutorials/polymorphism.cc
jbcoe: I've removed the comment altogether. It's misleading as we never store animals in a container later.

tutorials/polymorphism.cc
jbcoe: Perhaps we can add an extra example? `TypeErasureWithOwnership`?

jbcoe:
> The code itself is good, I left a few comments. I think one big thing the tutorial is missing is "why?", since this is fundamentally the motivation for `protocol<T>`. Point out that inheritance is intrusive, and the problems with that, but that manual type erasure requires more boilerplate for each interface you need - compounded even more so if you want an owning type. `protocol` combines the benefits of both approaches. Not sure if that works best at the start or end of this tutorial.

I want to leave the "why" out of the tutorial (perhaps incorrectly) and just objectively document how different (relevant) polymorphism approaches work. That way we can fairly assume that readers (and authors) have a common understanding.

We'll need to update (rewrite)`DRAFT.md` in due course and can discuss motivation there.

tutorials/polymorphism.cc
RyanJK5: My instinct is almost that it should get its own tutorial, simply because of how complex a complete implementation of that could be. It could start with the non-owning one from this tutorial, then add `destroy`, then `copy`. Then perhaps it could touch on a basic SBO implementation, which would also require adding `move`. You could then take the final result from that tutorial and do a separate one on adding an allocator template if we find that is needed.

jbcoe: Let's leave it for now. What we have here is nice and simple, without being simplistic.

RyanJK5:
> > The code itself is good, I left a few comments. I think one big thing the tutorial is missing is "why?", since this is fundamentally the motivation for `protocol<T>`. Point out that inheritance is intrusive, and the problems with that, but that manual type erasure requires more boilerplate for each interface you need - compounded even more so if you want an owning type. `protocol` combines the benefits of both approaches. Not sure if that works best at the start or end of this tutorial.
> 
> I want to leave the "why" out of the tutorial (perhaps incorrectly) and just objectively document how different (relevant) polymorphism approaches work. That way we can fairly assume that readers (and authors) have a common understanding.
> 
> We'll need to update (rewrite)`DRAFT.md` in due course and can discuss motivation there.

Understood! After resolving the discussion of how to discuss the owning version, then, this LGTM.

tutorials/polymorphism.cc
RyanJK5: I agree, I will open an issue for a separate tutorial on an owning version.

RyanJK5 (approved):

tutorials/polymorphism.cc
```diff
@@ -0,0 +1,298 @@
...
+  EXPECT_EQ(make_noise(cat), "Meow");
+  EXPECT_EQ(make_noise(dog), "Woof");
+}
+
+}  // namespace xyz::tutorials::compile_time_polymorphism
+
+// Polymorphism with inheritance
+
+// Cat and Dog have a common type. We can use a single function for both types.
+// This form of inheritance is intrusive. Cat and Dog must declare Animal as a
```
nbx8: Though it's likely understood what is meant in this context by intrusive, still worth clarifying - maybe connect to the following sentences with a colon?

tutorials/polymorphism.cc
```diff
@@ -0,0 +1,298 @@
...
+// base class and Animal must declare which member functions can be used
+// polymorphically.
```
nbx8: ...using virtual (?)

tutorials/polymorphism.cc
```diff
@@ -0,0 +1,298 @@
...
+ public:
+  std::string_view noise() const { return "Meow"; }
+};
+
+class Dog {
+ public:
+  std::string_view noise() const { return "Woof"; }
+};
+
+class AnimalPtr {
```
nbx8, line 132: add comment (something like): // wide pointer definition

tutorials/polymorphism.cc
```diff
@@ -0,0 +1,298 @@
...
+  // Non-virtual member functions are not found by virtual dispatch.
+  EXPECT_NE(animals[0]->identity(), cat.identity());
+  EXPECT_NE(animals[1]->identity(), dog.identity());
+}
+
+}  // namespace xyz::tutorials::polymorphism_with_inheritance
+
+// Polymorphism with type-erasure.
+
+// Cat and Dog have no common type; we can define a type-erasing wide-pointer to
```
nbx8: wide pointer (instead of wide-pointer)

tutorials/polymorphism.cc
```diff
@@ -0,0 +1,298 @@
...
+};
+
+// `animal` is passed by value as AnimalPtr is small like `std::string_view`.
+std::string_view make_noise(AnimalPtr animal) { return animal.noise(); }
+
+TEST(TutorialsPolymorphism, PolymorphismWithTypeErasure) {
+  Cat cat;
+  Dog dog;
+
+  // We can store pointers to Cat and Dog using our wide pointer type.
```
nbx8: wide-pointer type

tutorials/polymorphism.cc
```diff
@@ -0,0 +1,298 @@
...
+
+    vtable_ = &vtable_for_type;
+  }
+
+  std::string_view noise() const { return vtable_->noise_func_(data_); }
+
+  std::string_view identity() const { return vtable_->identity_func_(data_); }
+};
+
+// Storing the vtable as apointer ensure that `AnimalPtr` is cheap to copy,
```
nbx8: as a pointer ensures that

nbx8 (approved):
Nice tutorial, thanks!

PR #140, opened 2026-08-13
Add a tutorial on allocator support

### LINKED ISSUE #139: Add a tutorial on allocator support
A minimal tutorial covering the techniques we use for protocol will benefit both authors and readers.

We gave a presentation on allocator support for polymorphic https://www.youtube.com/watch?v=hZyJtRY84P4

It would be useful to condense this (with typo fixes) into a GoogleTest tutorial.

RyanJK5:
First draft of a tutorial. Walks through adding basic default-initialized allocator support, then stateful allocators and copy semantics, and finally allocator-aware move and swap semantics.

Closes #139.


RyanJK5:
@jbcoe I put together this draft of the tutorial, but it definitely needs some work. Do you have any suggestions to avoid duplicating `Owner` between each of the steps of the tutorial? I considered out-of-line definitions, but I'm concerned it may be a bit overwhelming to show the entire allocator-aware class definition right after step 1. That being said, it may be less confusing than rewriting the class like I'm doing now.

tutorials/allocators.cc
```diff
@@ -0,0 +1,669 @@
...
+}  // namespace xyz::tutorials::owning_type
+
+// Next, we add very rudimentary support for allocators. We do
+// not expose it anywhere in the public interface, aside from
+// the template argument; instead, we require that the allocator
+// type is default-constructible (stateless).
+namespace xyz::tutorials::simple_allocator {
+
+template <typename T, typename Alloc = std::allocator<T>>
+  requires std::default_initializable<Alloc>
```
jbcoe: I think that requiring that an allocator is default_initializable won't guarantee that it's stateless, just that the state is not controlled on construction. Memory pools could come from a global pool of pools (lake?).

I suspect that this section `namespace xyz::tutorials::simple_allocator` needs dropping. It's all or-nothing with allocator complexity.

jbcoe (commented):
This looks great on a first pass. 

Maybe we could introduce the allocator-extended constructors as a step on their own?

tutorials/allocators.cc
RyanJK5: Good point. I will turn that section into an intro to the constructors.

RyanJK5:
@jbcoe Do we care about the exception guarantee for this demonstration? Right now it's using the basic guarantee because it made showcasing a little simpler, but if `protocol` is going to use a strong guarantee, I could change the tutorial to match.

jbcoe:
> @jbcoe Do we care about the exception guarantee for this demonstration? Right now it's using the basic guarantee because it made showcasing a little simpler, but if `protocol` is going to use a strong guarantee, I could change the tutorial to match.

It might be useful to add as another stage. We would like a strong exception guarantee if possible.

jbcoe:
> @jbcoe I put together this draft of the tutorial, but it definitely needs some work. Do you have any suggestions to avoid duplicating `Owner` between each of the steps of the tutorial? I considered out-of-line definitions, but I'm concerned it may be a bit overwhelming to show the entire allocator-aware class definition right after step 1. That being said, it may be less confusing than rewriting the class like I'm doing now.

I would not worry about duplicating `Owner`. We do similar things in other tutorials, although for smaller classes.

jbcoe:
@RyanJK5 I merged `main`, please let me know if you'd rather that I did not do this.

tutorials/allocators.cc
```diff
@@ -0,0 +1,585 @@
...
+};
+
+TEST(TutorialsAllocators, StatefulAllocator) {
+  using TestOwner = Owner<int, TestAlloc<int>>;
+
+  // Default constructs the allocator.
+  TestOwner o1;
+  EXPECT_EQ(o1.get_allocator().tag, 0);
+
+  // Default constructs the allocator, even though we provide a value.
```
jbcoe: I'd drop "even though we provide a value". Gave me pause for thought as that's expected behaviour.

jbcoe (approved):
Thanks for this.

RyanJK5:
Sorry to re-request, but the exception section turned out to be fairly sizable.

jbcoe:
> Sorry to re-request, but the exception section turned out to be fairly sizable.

Happy to re-review.

tutorials/allocators.cc
```diff
@@ -0,0 +1,827 @@
...
+  }
+
+ public:
+  // This explicitly opts us into allocator support, so the standard library can
+  // interface with Owner correctly.
+  using allocator_type = Alloc;
+
+  Owner()
+    requires std::default_initializable<Alloc>
+  = default;
```
jbcoe, line 176: Formatting here is not ideal. Not blocking but I wonder if our .clang-format settings can be improved?

jbcoe (approved):
Thanks for this!

tutorials/allocators.cc
RyanJK5: Good catch, I would expect the `requires` clause to be on the same line as the constructor.

PR #141, opened 2026-08-14
Add vanishing-this-pointer tutorial

jbcoe:


tutorials/vanishing_this.cc
```diff
@@ -0,0 +1,167 @@
...
+
+TEST(TutorialsVanishingThis, CastMemberToParent) {
+  struct A {
+    int x;
+    std::string name;
+  };
+
+  A a{.x = 5, .name = "Ozymandias"};
+  void* x_ptr = &a.x;
+  auto* a_ptr = static_cast<A*>(x_ptr);
```
RyanJK5:
```suggestion
  auto* a_ptr = reinterpret_cast<A*>(x_ptr)
```

I am not entirely sure if the cast-to-`void*` is legal by the standard. My understanding is that `void*` can only be used for strict aliasing. Regardless, I think it's more precise to use `reinterpret_cast` here.

tutorials/vanishing_this.cc
```diff
@@ -0,0 +1,167 @@
...
+
+TEST(TutorialsVanishingThis, CallMemberData) {
+  class Callable {
+   public:
+    int operator()(int x) { return x * 2; }
+  };
+
+  class A {
+   public:
+    Callable fn;
```
RyanJK5: I think we may as well introduce the `[[no_unique_address]]` here. The section is already pretty light, so we could take a slight bit of complexity from the following section.

tutorials/vanishing_this.cc
```diff
@@ -0,0 +1,167 @@
...
+    struct Callable {
+      int operator()(int x) {
+        static_assert(
+            offsetof(A, fn_) == 0,
+            "The callable member must be A's first data member for the "
+            "vanishing-this-pointer cast to be valid");
+
+        // Use reinterpret_cast as static_cast is not valid here.
+        int offset = reinterpret_cast<A*>(this)->value_;
+        return x + offset;
```
RyanJK5:
```suggestion
        int y = reinterpret_cast<A*>(this)->value_;
        return x + y;
```

The variable name `offset` may be a little confusing, since we're also using the term offset to refer to member variable offsets above.

tutorials/vanishing_this.cc
```diff
@@ -0,0 +1,167 @@
...
+                "add must be AddBase's first data member for the "
+                "vanishing-this-pointer cast to be valid");
+
+  const auto* base = reinterpret_cast<const AddBase*>(this);
+  const auto* owner = static_cast<const A*>(base);
+  return x + owner->value_;
+}
+
+// Multiply::operator() must be defined out of line: the reinterpret_cast and
+// static_cast require `MultiplyBase` and `A` to be complete types.
```
RyanJK5:
```suggestion
```

I think saying this once is fine. The above comment could be updated to say "`Add:operator()` and `Multiply::operator()` must be defined out of line..."

tutorials/vanishing_this.cc
```diff
@@ -0,0 +1,167 @@
...
+
+// Add::operator() must be defined out of line: the reinterpret_cast and
+// static_cast require `AddBase` and `A` to be complete types.
+int Add::operator()(int x) {
+  static_assert(offsetof(AddBase, add) == 0,
+                "add must be AddBase's first data member for the "
+                "vanishing-this-pointer cast to be valid");
+
+  const auto* base = reinterpret_cast<const AddBase*>(this);
+  const auto* owner = static_cast<const A*>(base);
```
RyanJK5, line 143: Depending how `const` correct we want the tutorial to be, I think we should either remove `const` here, since the member function itself is marked non-`const`, or just add `const` to all the member function definitions.

tutorials/vanishing_this.cc
```diff
@@ -0,0 +1,167 @@
...
+
+// Multiply::operator() must be defined out of line: the reinterpret_cast and
+// static_cast require `MultiplyBase` and `A` to be complete types.
+int Multiply::operator()(int x) {
+  static_assert(offsetof(MultiplyBase, multiply) == 0,
+                "multiply must be MultiplyBase's first data member for the "
+                "vanishing-this-pointer cast to be valid");
+
+  const auto* base = reinterpret_cast<const MultiplyBase*>(this);
+  const auto* owner = static_cast<const A*>(base);
```
RyanJK5, line 153: See comment above about `const`-ness.

tutorials/vanishing_this.cc
```diff
@@ -0,0 +1,167 @@
...
+}
+
+TEST(TutorialsVanishingThis, ParentClassAccessFromMultipleMemberDataCalls) {
+  // [[no_unique_address]] and the empty base class optimisation ensure that
+  // inheriting from `AddBase` and `MultiplyBase` has no effect on size.
+  static_assert(sizeof(A) == sizeof(int));
+
+  A a(3);
+  EXPECT_EQ(a.add(5), 8);
+  EXPECT_EQ(a.multiply(5), 15);
```
RyanJK5, line 166: Perhaps a little conclusive comment: "`add()` and `multiply()` now behave just like member functions on `A`."

RyanJK5 (changes requested):
Looks good.

RyanJK5 (commented):
Looks good now overall, the only remaining change seems to be the one comment I left about changing `static_cast` to `reinterpret_cast`, unless you thought otherwise.

jbcoe:
> Looks good now overall, the only remaining change seems to be the one comment I left about changing `static_cast` to `reinterpret_cast`, unless you thought otherwise.

There was a typo. I had:

```
  A a{.x = 5, .name = "Ozymandias"};
  void* x_ptr = &a.x;
  auto* a_ptr = static_cast<A*>(x_ptr);
  EXPECT_EQ(a_ptr->name, "Ozymandias");
```

The code I'd meant to write was:

```
  A a{.x = 5, .name = "Ozymandias"};
  int* x_ptr = &a.x; // <== TYPO FIXED
  auto* a_ptr = static_cast<A*>(x_ptr);
  EXPECT_EQ(a_ptr->name, "Ozymandias");
```

where, as you pointed out, `static_cast` won't work. Typo and cast now fixed. 

Rigour appreciated!

RyanJK5 (approved):

PR #146, opened 2026-08-14
Reflection tutorial

jbcoe:


tutorials/reflection.cc
```diff
@@ -0,0 +1,298 @@
...
+  static_assert(!is_function(^^Point));
+}
+
+TEST(TutorialsReflection, MembersOfWithConstevalBlock) {
+  // `nonstatic_data_members_of` inspects a class's members.
+
+  // We use `throw` inside a `consteval` block rather than `static_assert`
+  // or `EXPECT_*`: `static_assert` needs `members` to independently be a
+  // constant expression, and `EXPECT_*`'s comparison helpers aren't
+  // `constexpr` functions.
```
RyanJK5, line 86: This could probably do with a one-sentence explanation of what a `consteval` block is.

tutorials/reflection.cc
```diff
@@ -0,0 +1,298 @@
...
+  EXPECT_EQ(enum_to_string(Suit::Spades), "Spades");
+}
+
+consteval std::optional<Suit> suit_from_name(std::string_view name) {
+  for (std::meta::info e : enumerators_of(^^Suit)) {
+    if (identifier_of(e) == name) {
+      // `e` is a loop variable, not a constant expression, so we can't
+      // splice it with `return [:e:];`. `extract<Suit>(e)` only needs a
+      // value, not a fixed compile-time constant.
+      return extract<Suit>(e);
```
RyanJK5, line 241: I really like how naturally this was introduced.

tutorials/reflection.cc
```diff
@@ -0,0 +1,298 @@
...
+  // `substitute(^^Template, {args...})` instantiates a template from infos;
+  // `reflect_constant` turns an ordinary value into an info usable as a
+  // non-type template argument.
+  // clang-format off
+  constexpr std::meta::info array_info =
+      substitute(^^std::array, {^^double, std::meta::reflect_constant(3)});
+  // clang-format on
+
+  static_assert(std::is_same_v<typename[:array_info:], std::array<double, 3>>);
+}
```
RyanJK5, line 294: I think this could benefit from a comment explaining the use case of `substitute` - when you want to use a reflection in a `consteval` context even though it itself is not a constant expression.

For me, at least, I found the distinction between constant expressions and `consteval` to be one of the more difficult topics when first learning reflection.

tutorials/reflection.cc
```diff
@@ -0,0 +1,298 @@
...
+  // member functions that have no identifier; `is_function` narrows the
+  // range to functions, and `has_identifier` further narrows it to ones
+  // `identifier_of` accepts.
+
+  constexpr std::ranges::range auto member_functions = std::define_static_array(
+      members_of(^^Point, std::meta::access_context::current()) |
+      std::views::filter([](std::meta::info member) {
+        return is_function(member) && has_identifier(member);
+      }));
+
```
RyanJK5, line 129: nit: I would prefer to write this as two separate filters, but I'm more than happy to approve whatever you or others find more readable. (below code is not clang-formatted)

```c++
  constexpr std::ranges::range auto member_functions = std::define_static_array(
      members_of(^^Point, std::meta::access_context::current())
      | std::views::filter(std::meta::is_function)
      | std::views::filter(std::meta::has_identifier));

```

RyanJK5 (approved):
Really happy with how this came out. It keeps the relevant parts of reflection in clear focus.

tutorials/reflection.cc
jbcoe: @RyanJK5  If you have suggested wording, that would be awesome. I think there are still some gaps in my understanding around this.

RyanJK5: Maybe something like: "A variable that exists during constant evaluation is not necessarily `constexpr`. A `std::meta::info` local variable cannot always be plugged into a template (e.g. `MyType<typename [: myLocal :]>`) since it may not be `constexpr`. To workaround this, we can use `substitute`."

jbcoe: Thanks. I'd misunderstood here and thought that `substitute` was very template-specific.

RyanJK5: No problem. `substitute(^^Something, {arg1, arg2, arg3});` is always equivalent to `^^Something<[:arg1:], [:arg2:], [:arg3:]>`, for any kind of template (variable, function, type, ...)

Twon (approved):
I like the tutorial, its a useful introduction to reflection

PR #147, opened 2026-08-15
Allocator tests

### LINKED ISSUE #143: Add tests to the reflection implementation for allocator support

RyanJK5:
Adds a test suite for the allocator conformance of `protocol`, based on tests from [indirect and polymorphic](https://github.com/jbcoe/value_types). 

Closes #143

### DIFF
--- a/reflection/allocator_tests.cc
+++ b/reflection/allocator_tests.cc
@@ -589,7 +589,7 @@ TEST(ProtocolTest, MoveAssignmentUnequalPocma) {
     EXPECT_EQ(allocs1, 1);
     EXPECT_EQ(allocs2, 1);
 
-    // POCCA is true: p1 takes p2's allocator. Even though alloc1 != alloc2,
+    // POCMA is true: p1 takes p2's allocator. Even though alloc1 != alloc2,
     // p1 can just steal p2's pointer now, because it will have alloc2
     // henceforth.
     p1 = std::move(p2);
--- a/reflection/allocator_tests.cc
+++ b/reflection/allocator_tests.cc
@@ -504,6 +504,7 @@ TEST(ProtocolTest, NarrowingMoveConstructionUnequal) {
   EXPECT_EQ(deallocs2, 1);
 }
 
+// Extends TrackingAllocator with POCCA set to true.
 template <typename T>
 struct PoccaAllocator : xyz::TrackingAllocator<T> {
   using xyz::TrackingAllocator<T>::TrackingAllocator;
@@ -531,13 +532,20 @@ TEST(ProtocolTest, CopyAssignmentUnequalPocca) {
     xyz::protocol<IntLike, PoccaAllocator<std::byte>> p2{std::allocator_arg,
                                                          alloc2, Tester{0}};
 
+    // p1 and p2 are both constructed with different allocators.
     EXPECT_EQ(allocs1, 1);
     EXPECT_EQ(allocs2, 1);
 
+    // POCCA is true: p1 should now take p2's allocator instead of keeping its
+    // own.
     p1 = p2;
+    EXPECT_EQ(p1.get_allocator(), alloc2);
 
-    EXPECT_EQ(allocs1, 1);
+    EXPECT_EQ(allocs1, 1);  // alloc1 only allocated once for p1.
+    // alloc2 allocated once for p2, and again when copying into p1.
     EXPECT_EQ(allocs2, 2);
+
+    // p1 was deallocated when it got assigned to.
     EXPECT_EQ(deallocs1, 1);
     EXPECT_EQ(deallocs2, 0);
   }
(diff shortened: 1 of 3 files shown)


RyanJK5:
@jbcoe I haven't wired the tests into CMake or anything yet - is there a better base branch to target this on instead of main?

reflection/allocator_tests.cc
```diff
@@ -0,0 +1,746 @@
...
+
+  int value() const noexcept { return val_; }
+};
+
+TEST(ProtocolTest, Construction) {
+  static_assert(
+      not std::is_nothrow_constructible_v<TestProtocol, std::allocator_arg_t,
+                                          TestAlloc, Tester>);
+
+  unsigned allocs{};
```
jbcoe: I'd prefer read `unsigned allocs = 0;`

For those of us who write Python most of the time, it saves a mental pause.

reflection/allocator_tests.cc
```diff
@@ -0,0 +1,746 @@
...
+  unsigned allocs2{};
+  unsigned deallocs2{};
+  {
+    TestAlloc alloc1{&allocs1, &deallocs1};
+    TestProtocol p1{std::allocator_arg, alloc1, Tester{100}};
+
+    TestAlloc alloc2{&allocs2, &deallocs2};
+    TestProtocol p2{std::allocator_arg, alloc2, p1};
+
+    EXPECT_EQ(allocs1, 1);
```
jbcoe: Once the logic becomes non-obvious, I'd prefer to read some comments about what's going on any why.

jbcoe:
@RyanJK5 I think you can base this on `main` - it's in a decent enough state for now.

jbcoe (changes requested):
Thanks for this. 

I think some inline comments would make reading this much simpler.

Maybe we can add some copy/move construction tests that compare the allocators in cases where we'd not expect them to compare equal too? This might mean that the allocators need a tag and might work best in a follow-up PR.

RyanJK5:
> Thanks for this.
> 
> I think some inline comments would make reading this much simpler.
> 
> Maybe we can add some copy/move construction tests that compare the allocators in cases where we'd not expect them to compare equal too? This might mean that the allocators need a tag and might work best in a follow-up PR.

On the copy/move construction tests with unequal allocators, I did add some (those are the ones with `allocs1` and `allocs2`). If that wasn't clear, however, it's my fault for not communicating that clearly with comments. I will work on that.

Additionally, if we want more explicit tests on this, I can write some using `TaggedAllocator`.

reflection/allocator_tests.cc
```diff
@@ -0,0 +1,867 @@
...
+
+// A simple test class that we will place in protocol.
+class Tester {
+  int val_;
+
+  // protocol should probably implement small buffer optimization, and therefore
+  // we may find there are fewer allocations than these tests suggest. To
+  // sidestep this, we can use a type that is guaranteed to be larger than
+  // protocol.
+  std::array<std::byte, sizeof(TestProtocol)> padding_;
```
jbcoe, line 54: Nice trick and nice comment!

reflection/allocator_tests.cc
```diff
@@ -0,0 +1,867 @@
...
+  unsigned deallocs1 = 0;
+
+  unsigned allocs2 = 0;
+  unsigned deallocs2 = 0;
+  {
+    TestAlloc alloc1{&allocs1, &deallocs1};
+    TestProtocol p1{std::allocator_arg, alloc1, Tester{100}};
+    EXPECT_EQ(allocs1, 1);
+
+    // alloc2 != alloc1. We cannot perform a simple pointer-swap
```
jbcoe, line 285: Nice. This is easy to follow and a subtle bit of behaviour.

reflection/allocator_tests.cc
```diff
@@ -0,0 +1,867 @@
...
+
+    PocmaAllocator<std::byte> alloc2{&allocs2, &deallocs2};
+    xyz::protocol<IntLike, PocmaAllocator<std::byte>> p2{std::allocator_arg,
+                                                         alloc2, Tester{10}};
+
+    // p1 and p2 are both constructed with different allocators.
+    EXPECT_EQ(allocs1, 1);
+    EXPECT_EQ(allocs2, 1);
+
+    // POCCA is true: p1 takes p2's allocator. Even though alloc1 != alloc2,
```
jbcoe: Is POCCA here a typo? I think it should be POCMA.

jbcoe (approved):
Thanks so much for this. Very easy to follow. 

Now inspired, I shall go and revise the tests for `indirect` and `polymorphic`.

RyanJK5:
> Thanks so much for this. Very easy to follow.
> 
> Now inspired, I shall go and revise the tests for `indirect` and `polymorphic`.

I'm glad they were helpful! The existing tests from value_types were the only reason I was able to put this together to begin with :)

PR #149, opened 2026-08-17
constexpr support in vanishing-this

jbcoe:
Use C++26's support for constexpr casting from void* to support vanishing this at compile-time.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/tutorials/vanishing_this.cc
+++ b/tutorials/vanishing_this.cc
@@ -24,6 +24,7 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 #include <gtest/gtest.h>
 
 #include <cstddef>
+#include <exception>
 #include <string>
 #include <string_view>
 #include <type_traits>
@@ -41,6 +42,12 @@ TEST(TutorialsVanishingThis, CastMemberToParent) {
 
   A a{.x = 5, .name = "Ozymandias"};
   int* x_ptr = &a.x;
+
+  // `A` must be a standard layout type for a cast from a member at offset 0
+  // to the parent object to be valid.
+  static_assert(std::is_standard_layout_v<A>);
+  static_assert(offsetof(A, x) == 0);
+
   auto* a_ptr = reinterpret_cast<A*>(x_ptr);
   EXPECT_EQ(a_ptr->name, "Ozymandias");
 }
@@ -76,10 +83,8 @@ TEST(TutorialsVanishingThis, ParentClassAccessFromMemberDataCall) {
   class A {
     struct Callable {
       int operator()(int x) const {
-        static_assert(
-            offsetof(A, fn_) == 0,
-            "The callable member must be A's first data member for the "
-            "vanishing-this-pointer cast to be valid");
+        static_assert(std::is_standard_layout_v<A>);
+        static_assert(offsetof(A, fn_) == 0);
 
         // Use reinterpret_cast as static_cast is not valid here.
         int value = reinterpret_cast<const A*>(this)->value_;
@@ -135,6 +140,7 @@ struct A : AddBase, MultiplyBase {
 // to be complete types.
 
 int Add::operator()(int x) const {
+  static_assert(std::is_standard_layout_v<AddBase>);
   static_assert(offsetof(AddBase, add) == 0,
                 "add must be AddBase's first data member for the "
                 "vanishing-this-pointer cast to be valid");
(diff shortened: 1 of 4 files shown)


tutorials/vanishing_this.cc
```diff
@@ -41,6 +42,12 @@ TEST(TutorialsVanishingThis, CastMemberToParent) {
 
   A a{.x = 5, .name = "Ozymandias"};
   int* x_ptr = &a.x;
+
+  // `A` must be a standard layout type for a cast from a member at offset 0
+  // to the parent object to be valid.
+  static_assert(std::is_standard_layout_v<A>);
+  static_assert(offsetof(A, x) == 0);
```
RyanJK5, line 49: Good catch. This makes it clearer to the reader about the limitations of this technique.

tutorials/vanishing_this.cc
```diff
@@ -167,3 +174,59 @@ TEST(TutorialsVanishingThis, ParentClassAccessFromMultipleMemberDataCalls) {
 }
 
 }  // namespace xyz::tutorials::access_parent_from_multiple_callable_members
+
+// `reinterepret_cast` cannot be used at constant-evaluation time but C++ 26
+// permits `static_cast` from `void*` to `T*` where `T` is a standard-layout
+// type. We can exploit this to implement the `reinterpret_cast` logic above as
+// two, chained, `constexpr` `static_cast` expressions.
```
RyanJK5, line 181: Unfortunately, I believe this is ill-formed in C++26. Per the [proposal](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p2738r1.pdf):

> We propose to allow casting from a pointer of `void*` to a pointer of type `T` in `constexpr` if the
type of the object at that address is exactly of type `T`. In particular, we are not proposing
allowing conversion to a pointer that would be interconvertible, let alone unrelated.

GCC currently accepts it, however I believe this is a bug. Clang currently rejects it. See more here: [Compiler Explorer](https://godbolt.org/z/MYqbK7xeW)

jbcoe: Sad. I guess we wait for C++29 constexpr.

tutorials/vanishing_this.cc
hanickadot: AFAIK improvement would be to use `std::is_pointer_interconvertible_base_of`

tutorials/vanishing_this.cc
hanickadot: yes, you can do cast to `void *` and you must get same type out of it, anything else is result of GCC's implementation, for that you need to have theP3377 (https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2026/p3377r0.html) of mine which implements the `std::enclosing_cast` which someone from internet implemented here: https://godbolt.org/z/EMK9sK6Yn

tutorials/vanishing_this.cc
```diff
@@ -167,3 +174,59 @@ TEST(TutorialsVanishingThis, ParentClassAccessFromMultipleMemberDataCalls) {
...
+  static_assert(std::is_pointer_v<To>, "Target type must be a pointer");
+  using Target = std::remove_pointer_t<To>;
+  static_assert(
+      std::is_standard_layout_v<Target>,
+      "Target type must be standard-layout for pointer interconvertibility");
+  return static_cast<To>(static_cast<void*>(ptr));
+}
+
+template <typename To, typename From>
+constexpr To type_erased_cast(const From* ptr) noexcept {
```
hanickadot, line 196: you don't need to duplicate these function twice, you can use `std::common_type_t` to copy CV qualifier to `void *`

https://godbolt.org/z/TzT9q4fcs

tutorials/vanishing_this.cc
```diff
@@ -167,3 +174,59 @@ TEST(TutorialsVanishingThis, ParentClassAccessFromMultipleMemberDataCalls) {
...
+};
+
+struct A : AddBase {
+  int value_;
+
+  constexpr A(int value) : value_(value) {}
+};
+
+constexpr int Add::operator()(int x) const {
+  const auto* base = type_erased_cast<const AddBase*>(this);
```
hanickadot, line 222: even when address is same, the provenance is not, stop thinking about location of objects as one integer-like address ... think about it more like as a path from outside the object to inner object:

here the `this` is actually inside `A` path like this:

[object (A) → first base (AddBase) → first member (Add) →] self (Add)

and you are casting it to:

[object (AddBase) →] self (AddBase)

the part in square brackets are implicit and invisible at that call site. Memory location is same, but the provenance is not :(

PR #151, opened 2026-08-18
Owning type erasure tutorial

### LINKED ISSUE #137: Add a tutorial for owning type-erased container
If we find that it would be helpful, a tutorial walking through an owning type-erased container could be useful:

1. Start with the non-owning type-erased example from #133. Extend it to allocate `T` itself and store that in its `void*`. Add `destroy` to the vtable and wire that into the destructor.
2. Fill out the move semantics with simple pointer swaps, then add `copy` to the vtable to allow copy semantics.
3. (Optional) Add a simple SBO to this type. Could use either a `union` with a `bool` or just store the buffer and pointer next to each other (which can be more efficient). Add `move` to the vtable so we can preserve move semantics when using SBO.
4. (Optional) An additional tutorial on adding an `Allocator` template argument, and how that changes the internals.

RyanJK5:
I am not sure how necessary this tutorial ended up being, but I think it couldn't hurt to add.

Closes #137


tutorials/type_erasure.cc
```diff
@@ -0,0 +1,192 @@
...
+    .noise_func_ = +[](const void* data) -> std::string_view {
+      throw std::bad_function_call{};
+    },
+    // Destroying an object with a null vtable does nothing.
+    .destroy_ = +[](void* data) -> void {},
+    // Copying an object with a null vtable does nothing.
+    .copy_ = +[](const void* src) -> void* { return nullptr; }};
+
+class Animal {
+  // By default, Animal is empty.
```
jbcoe: Q: This is good here but to clarify, it's not what we plan to do with `protocol`?

jbcoe (commented):
Thanks for this - super clear and easy to follow.

tutorials/type_erasure.cc
RyanJK5: I would argue it should be used for `protocol` as well. I originally implemented `duck` using null checking instead of a null vtable, but found that the code was simpler when written like this instead. It also is a very tiny performance win, since we dodge null checking inside the lifecycle functions.

RyanJK5: Sorry, previous comment was incorrect. I thought you commented on the null_vtable. Please disregard. Yes, protocol will not have default construction.

RyanJK5: For consistency, I went ahead and removed the default constructor from this tutorial.

jbcoe:
@RyanJK5 would you like me to clean up the CMake issues and push to the branch?

RyanJK5:
> @RyanJK5 would you like me to clean up the CMake issues and push to the branch?

Please do. I'm not sure what happened, did it time out while installing GCC?

jbcoe (approved):

PR #152, opened 2026-08-18
Add infrastructure for C++26 compiler support (GCC16)

jbcoe:


CMakeLists.txt
```diff
@@ -29,6 +29,34 @@ option(ENABLE_UBSAN "Enable Undefined Behaviour Sanitizer" OFF)
...
 
+option(
+  XYZ_PROTOCOL_BUILD_REFLECTION
+  "Also build and test tutorials/reflection.cc, which requires a compiler \
+with C++26 P2996 reflection support (GCC 16+ with -freflection)."
+  OFF)
+
+if(XYZ_PROTOCOL_BUILD_REFLECTION)
+  include(CheckCXXSourceCompiles)
+  set(CMAKE_REQUIRED_FLAGS "-std=c++26 -freflection")
```
Twon: This could be executed on MSCV (or at least I don't see anything to forbit it so it should guard the compiler flags with an if gcc like generator expression for the std component and a gcc generator expression for the reflection flag.

Note that '-std=c++26' will fail on older compiler so should probably be constrained to gcc 16 up and clang 23 onwards

CMakeLists.txt
```diff
@@ -29,6 +29,34 @@ option(ENABLE_UBSAN "Enable Undefined Behaviour Sanitizer" OFF)
...
+  check_cxx_source_compiles(
+    "
+    #include <meta>
+    constexpr std::meta::info reflection_of_int = ^^int;
+    int main() {}
+    "
+    XYZ_PROTOCOL_REFLECTION_SUPPORTED)
+  unset(CMAKE_REQUIRED_FLAGS)
+  if(NOT XYZ_PROTOCOL_REFLECTION_SUPPORTED)
```
Twon, line 52: This may suddenly start working for newer compilers as they add support (assuming they just ignore the reflection flag if they do not have it), so it is not necessarily GCC 16 specific as written, just that's the only compiler that will pass at this snapshot in time

cmake/xyz_add_test.cmake
```diff
@@ -50,7 +50,7 @@ function(xyz_add_test)
   if(NOT XYZ_VERSION)
     set(XYZ_VERSION 20)
   else()
-    set(VALID_TARGET_VERSIONS 11 14 17 20 23)
+    set(VALID_TARGET_VERSIONS 11 14 17 20 23 26)
```
Twon, line 53: This was only added in cmake 3.25, so lets make sure our global cmake requires aligns to this

Twon (approved):
Looks good, a few minor comments but nothing major

jbcoe:
> Looks good, a few minor comments but nothing major

I guess in general we want noisy failure when reflection is not supported rather than silent fallback. Can I change the CMake setup to ensure this?

CMakeLists.txt
Twon: Seems for a compiler agnostic approach like you can use:
```
set(CMAKE_CXX_STANDARD 26)
```

Compilers that do not support C++26 will fail, but that's what we want

CMakeLists.txt
```diff
@@ -29,6 +29,34 @@ option(ENABLE_UBSAN "Enable Undefined Behaviour Sanitizer" OFF)
...
+  "Also build and test tutorials/reflection.cc, which requires a compiler \
+with C++26 P2996 reflection support (GCC 16+ with -freflection)."
+  OFF)
+
+if(XYZ_PROTOCOL_BUILD_REFLECTION)
+  include(CheckCXXSourceCompiles)
+  set(CMAKE_REQUIRED_FLAGS "-std=c++26 -freflection")
+  check_cxx_source_compiles(
+    "
+    #include <meta>
```
Twon, line 46: I might be tempted to use the feature test flag, as this is the official compiler mechanism to flag support for a feature:
```
    #include <version>

    #ifndef __cpp_impl_reflection
    #error No reflection support
    #endif

    int main() {}
```

CMakeLists.txt
Twon:
```suggestion
  set(CMAKE_REQUIRED_FLAGS "$<$<CXX_COMPILER_ID:GNU>:-freflection>")
```

Twon:
> > Looks good, a few minor comments but nothing major
> 
> I guess in general we want noisy failure when reflection is not supported rather than silent fallback. Can I change the CMake setup to ensure this?

I think the fatal error is sufficient once we avoid specific compiler flags or restrict them to the appropriate compiler, as then we avoid failures due to using an unsupported flag and as then it should just work as future compilers add support.

CMakeLists.txt
jbcoe: Agreed. Much improved and forward-maintainable.

CMakeLists.txt
jbcoe: I could not get GCC16 to work with the feature flag - perhaps they've not flipped the switch as there's still some work to do?

CMakeLists.txt
jbcoe:
```
  if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
    set(CMAKE_REQUIRED_FLAGS "-freflection")
  endif()
```

Seemed to behave better. 

`check_cxx_source_compiles` ran before generator expressions were evaluated and took in corrupted flags.

nbx8:
@jbcoe, one suggestion (not essential but could be helpful): split the C++26 compiler support and the sandbox option update into two separate PRs.

jbcoe:
> @jbcoe, one suggestion (not essential but could be helpful): split the C++26 compiler support and the sandbox option update into two separate PRs.

Fine idea. Done in https://github.com/jbcoe/cc-protocol/pull/159

PR #168, opened 2026-08-24
Add member function implementation stubs using reflection

jbcoe:
Adds stubs for member function to both `protocol` and `protocol_view`. Stub member functions cannot be called (there is currently no vtable to forward calls), but their signatures can be checked at compile-time.

### DIFF
--- a/reflection/protocol.hh
+++ b/reflection/protocol.hh
@@ -175,7 +175,8 @@ consteval std::vector<std::meta::info> generate_method_thunk_specs(
     // Build the function-pointer type R(*)(Args...) from the method's
     // return type and parameter types.
     std::vector<std::meta::info> fn_args{dealias(return_type_of(member))};
-    for (std::meta::info parameter : parameters_of(member)) {
+    std::vector<std::meta::info> member_parameters = parameters_of(member);
+    for (std::meta::info parameter : member_parameters) {
       fn_args.push_back(dealias(type_of(parameter)));
     }
     std::meta::info fn_ptr_type = substitute(^^fn_ptr_t, fn_args);
--- a/reflection/protocol.hh
+++ b/reflection/protocol.hh
@@ -129,16 +129,17 @@ struct method_thunk<R (*)(Args...), EnclosingType, IsConst, IsNoexcept> {
   R operator()(Args... /*args*/) noexcept(IsNoexcept)
     requires(!IsConst)
   {
-    [[maybe_unused]] auto* base =
-        static_cast<EnclosingType*>(static_cast<void*>(this));
+    // Use reinterpret_cast as static_cast is not valid here.
+    [[maybe_unused]] auto* base = reinterpret_cast<EnclosingType*>(this);
     std::unreachable();  // vtable dispatch not yet implemented
   }
 
   R operator()(Args... /*args*/) const noexcept(IsNoexcept)
     requires(IsConst)
   {
+    // Use reinterpret_cast as static_cast is not valid here.
     [[maybe_unused]] const auto* base =
-        static_cast<const EnclosingType*>(static_cast<const void*>(this));
+        reinterpret_cast<const EnclosingType*>(this);
     std::unreachable();  // vtable dispatch not yet implemented
   }
 };
--- a/reflection/protocol.hh
+++ b/reflection/protocol.hh
@@ -201,24 +201,24 @@ struct protocol_member_stubs_generator {
 
 }  // namespace detail
 
-// Returns `true` if `U` is a structural subtype of the interface `T`;
+// Returns `true` if `Candidate` is a structural subtype of `Interface`;
 // otherwise returns `false`.
-template <typename T, typename U>
+template <typename Interface, typename Candidate>
 consteval bool is_protocol_conformant() {
-  static_assert(std::is_same_v<T, std::remove_cvref_t<T>>,
-                "the interface type must not be cv/ref-qualified: strip "
-                "qualifiers at the call site with std::remove_cvref_t.");
-  static_assert(std::is_same_v<U, std::remove_cvref_t<U>>,
-                "the candidate type must not be cv/ref-qualified: strip "
-                "qualifiers at the call site with std::remove_cvref_t.");
+  static_assert(std::is_same_v<Interface, std::remove_cvref_t<Interface>>,
+                "Interface must not be cv/ref-qualified: strip qualifiers at "
+                "the call site with std::remove_cvref_t.");
+  static_assert(std::is_same_v<Candidate, std::remove_cvref_t<Candidate>>,
+                "Candidate must not be cv/ref-qualified: strip qualifiers at "
+                "the call site with std::remove_cvref_t.");
 
   // Checking for protocol interface conformance is O(N*M) over member counts,
   // assumed to be negligible at compile time.
   // TODO(jbcoe): Use set/map once there is library support for `constexpr`.
   auto interface_member_functions =
-      detail::protocol_interface_functions_of<^^T>();
+      detail::protocol_interface_functions_of<^^Interface>();
   auto candidate_member_functions =
-      detail::protocol_interface_functions_of<^^U>();
+      detail::protocol_interface_functions_of<^^Candidate>();
 
   return std::ranges::all_of(
       interface_member_functions, [&](std::meta::info interface_member) {
@@ -231,8 +231,9 @@ consteval bool is_protocol_conformant() {
 }
 
 // Variable template for use in requires clauses.
-template <typename T, typename U>
-inline constexpr bool is_protocol_conformant_v = is_protocol_conformant<T, U>();
+template <typename Interface, typename Candidate>
+inline constexpr bool is_protocol_conformant_v =
+    is_protocol_conformant<Interface, Candidate>();
 
 // ---------------------------------------------------------------------------
 // protocol<I, Allocator>
--- a/reflection/protocol.hh
+++ b/reflection/protocol.hh
@@ -91,7 +91,7 @@ consteval bool member_function_conforms_to(std::meta::info candidate,
     return false;
   std::vector<std::meta::info> interface_params = parameters_of(interface);
   std::vector<std::meta::info> candidate_params = parameters_of(candidate);
-  // parameter counts must match.
+  // Parameter counts must match.
   if (interface_params.size() != candidate_params.size()) return false;
   for (std::size_t i = 0; i < interface_params.size(); ++i) {
     // De-aliased parameter types must match.
@@ -113,12 +113,9 @@ consteval auto protocol_interface_functions_of() {
       std::views::filter(std::meta::has_identifier));
 }
 
-// ---------------------------------------------------------------------------
-// Vanishing-this-pointer thunk for a synthesised member stub.
-//
-// The thunk carries a single operator() whose signature mirrors one method
-// of the Interface type.
-// ---------------------------------------------------------------------------
+// Vanishing-this-pointer thunk for a synthesised member stub. The thunk
+// carries a single operator() whose signature mirrors one method of the
+// Interface type.
 template <typename FnPtrType, typename EnclosingType, bool IsConst,
           bool IsNoexcept>
 struct method_thunk;
@@ -129,7 +126,6 @@ struct method_thunk<R (*)(Args...), EnclosingType, IsConst, IsNoexcept> {
   R operator()(Args... /*args*/) noexcept(IsNoexcept)
     requires(!IsConst)
   {
-    // Use reinterpret_cast as static_cast is not valid here.
     [[maybe_unused]] auto* base = reinterpret_cast<EnclosingType*>(this);
     std::unreachable();  // vtable dispatch not yet implemented
   }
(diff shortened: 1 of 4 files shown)


reflection/protocol.hh
```diff
@@ -93,32 +100,120 @@ consteval bool member_function_conforms_to(std::meta::info candidate,
...
-      std::define_static_array(
-          members_of(^^Candidate, std::meta::access_context::unprivileged())) |
+// Vanishing-this-pointer thunk for a synthesised member stub. The thunk
+// carries a single operator() whose signature mirrors one method of the
+// Interface type.
+template <typename FnPtrType, typename EnclosingType, bool IsConst,
+          bool IsNoexcept>
+struct method_thunk;
+
+template <typename R, typename... Args, typename EnclosingType, bool IsConst,
```
RyanJK5, line 124: I would add a note / todo of some sort to this related to handling lvalue- and rvalue-ness correctly here.

reflection/protocol.hh
```diff
@@ -93,32 +100,120 @@ consteval bool member_function_conforms_to(std::meta::info candidate,
...
+template <typename R, typename... Args>
+using fn_ptr_t = R (*)(Args...);
+
+// Returns a list of data_member_spec values, one for each member function
+// implemented by `protocol`. Each data_member_spec names the data member
+// after the interface method (giving the `p.method_name(args)` call syntax)
+// and sets its type to the matching thunk template specialisation.
+//
+// C++ disallows two data members with the same name inside the same class,
+// so overloaded methods will cause compile-time errors; unsupported for now.
```
RyanJK5: The right way to handle this is: instead of making `method_thunk` data members directly, use the following class template:

```c++
template <typename... Thunks>
struct overload_set : Thunks... {
    using Thunks::operator()...;
};
```

Group member functions by name, `substitute` each of them into `overload_set`, and create `data_member_spec`s for the overload sets instead of the individual functions.

That's probably best left for a follow-up PR, but just wanted to note here.

reflection/protocol.hh
```diff
@@ -93,32 +100,120 @@ consteval bool member_function_conforms_to(std::meta::info candidate,
...
+  for (std::meta::info member : members) {
+    std::string_view name = identifier_of(member);
+
+    // Build the function-pointer type R(*)(Args...) from the method's
+    // return type and parameter types.
+    std::vector<std::meta::info> fn_args{dealias(return_type_of(member))};
+    std::vector<std::meta::info> member_parameters = parameters_of(member);
+    for (std::meta::info parameter : member_parameters) {
+      fn_args.push_back(dealias(type_of(parameter)));
+    }
```
RyanJK5:
```suggestion
    std::vector<std::meta::info> member_parameters = parameters_of(member);
    fn_args.append_range(member_parameters
        | std::views::transform(std::meta::type_of));
```

I believe `dealias` is unnecessary here, per this statement from the parameters proposal:

> We propose to extend the type_of metafunction... to return the type adjusted parameter type (de-aliased, cv-unqualified, pointer-decayed). This is to achieve increased consistency of the results of type_of and consistency with function type.

reflection/protocol.hh
```diff
@@ -93,32 +100,120 @@ consteval bool member_function_conforms_to(std::meta::info candidate,
...
+    std::meta::info fn_ptr_type = substitute(^^fn_ptr_t, fn_args);
+
+    std::meta::info thunk_type = substitute(
+        ^^method_thunk, {
+                            fn_ptr_type, enclosing_type,
+                            std::meta::reflect_constant(is_const(member)),
+                            std::meta::reflect_constant(is_noexcept(member))});
```
RyanJK5: Maybe worth disabling clang-format here.

reflection/protocol.hh
```diff
@@ -93,32 +100,120 @@ consteval bool member_function_conforms_to(std::meta::info candidate,
...
+
+// Generates a struct that has named members with `operator()` for each public,
+// non-special, member function from `T`.
+template <typename T>
+struct protocol_member_stubs_generator {
+  struct stubs;
+  consteval {
+    define_aggregate(^^stubs, generate_method_thunk_specs(^^T, ^^stubs));
+  }
+};
```
RyanJK5: Q: This is not currently using the multiple-bases part of the vanishing-this trick. Is that for a follow-up PR?

reflection/protocol.hh
```diff
@@ -93,32 +100,120 @@ consteval bool member_function_conforms_to(std::meta::info candidate,
...
+consteval bool is_protocol_conformant() {
+  static_assert(std::is_same_v<T, std::remove_cvref_t<T>>,
+                "the interface type must not be cv/ref-qualified: strip "
+                "qualifiers at the call site with std::remove_cvref_t.");
+  static_assert(std::is_same_v<U, std::remove_cvref_t<U>>,
+                "the candidate type must not be cv/ref-qualified: strip "
+                "qualifiers at the call site with std::remove_cvref_t.");
+
+  // Checking for protocol interface conformance is O(N*M) over member counts,
+  // assumed to be negligible at compile time.
```
RyanJK5, line 227: Once there's library support for `constexpr` `set` and `map`, this could be changed to use those containers.

reflection/protocol.hh
```diff
@@ -93,32 +100,120 @@ consteval bool member_function_conforms_to(std::meta::info candidate,
...
-                "the call site with std::remove_cvref_t.");
-  static_assert(std::is_same_v<Candidate, std::remove_cvref_t<Candidate>>,
-                "Candidate must not be cv/ref-qualified: strip qualifiers at "
-                "the call site with std::remove_cvref_t.");
-
-  // O(N*M) over member counts, assumed to be negligible at compile time.
-  std::ranges::range auto interface_member_functions =
+// The named, non-static, non-special member functions of `Type`.
+template <std::meta::info Type>
+consteval auto protocol_interface_functions_of() {
```
RyanJK5: Worth a todo or issue, since this is not a complete solution:

```c++
struct S {
    static void foo();
};

struct I {
    void foo();
};

S s;
s.foo(); // OK

protocol_view<I> p{s}; // ERROR
p.foo();
```

reflection/protocol_test.cc
```diff
@@ -394,4 +396,143 @@ TEST(ReflectionProtocolTest, IsConstructibleInPlaceWithArguments) {
...
+TEST(ReflectionProtocolTest, ConstMemberFunction) {
+  struct Interface {
+    int get_value() const;
+  };
+
+  // TODO(jbcoe): replace static assertion with runtime test.
+  static_assert(requires(const protocol<Interface>& p) {
+    { p.get_value() } -> std::same_as<int>;
+  });
+}
```
RyanJK5, line 493: I think a negative test for something like this would be good:

```c++
struct Interface {
    int get_value();
};

const protocol<Interface> p{...};
p.get_value(); // ERROR
```

reflection/protocol_test.cc
```diff
@@ -394,4 +396,143 @@ TEST(ReflectionProtocolTest, IsConstructibleInPlaceWithArguments) {
...
+
+TEST(ReflectionProtocolTest, NoexceptMemberFunction) {
+  struct Interface {
+    double compute(double input) noexcept;
+  };
+
+  // TODO(jbcoe): replace static assertion with runtime test.
+  static_assert(requires(protocol<Interface>& p) {
+    { p.compute(0.0) } noexcept -> std::same_as<double>;
```
RyanJK5, line 523: Neat, I didn't realize you could check `noexcept`-ness like that. I will have to update some `duck` code :)

reflection/protocol_test.cc
```diff
@@ -394,4 +396,143 @@ TEST(ReflectionProtocolTest, IsConstructibleInPlaceWithArguments) {
...
+    double add(double x, double y) const noexcept;
+    double multiply(double x, double y) const noexcept;
+  };
+
+  // TODO(jbcoe): replace static assertion with runtime test.
+  static_assert(requires(const protocol<Interface>& p) {
+    { p.add(1.0, 2.0) } noexcept -> std::same_as<double>;
+    { p.multiply(3.0, 4.0) } noexcept -> std::same_as<double>;
+  });
+}
```
RyanJK5, line 560: Maybe a test for something malicious like:

```c++
struct Interface {
    bool valueless_after_move() const noexcept;
};

protocol<Interface> p{...};
```

I'm not sure what the best way to handle it is, but probably worth acknowledging.

reflection/protocol.hh
```diff
@@ -130,19 +225,18 @@ consteval bool is_protocol_conformant() {
 }
 
 // Variable template for use in requires clauses.
-template <typename Interface, typename Candidate>
-inline constexpr bool is_protocol_conformant_v =
-    is_protocol_conformant<Interface, Candidate>();
+template <typename T, typename U>
+inline constexpr bool is_protocol_conformant_v = is_protocol_conformant<T, U>();
```
RyanJK5: Personally, I think the prior naming scheme was more understandable at a glance. If shortening the type names is desirable, maybe `I` and `T` could be used? It's hard to tell which is which now in my opinion.

RyanJK5: This applies to several locations throughout the file.

RyanJK5 (changes requested):
Good work! The core of it seems to be functioning nicely :)

reflection/protocol.hh
jbcoe: Thanks!

reflection/protocol.hh
```diff
@@ -93,6 +100,119 @@ consteval bool member_function_conforms_to(std::meta::info candidate,
...
+                                   std::meta::access_context::unprivileged())) |
+                    std::views::filter(std::meta::is_function) |
+                    std::views::filter(
+                        std::not_fn(std::meta::is_special_member_function)) |
+                    std::views::filter(
+                        std::not_fn(std::meta::is_static_member)) |
+                    std::views::filter(std::meta::has_identifier))) {
+    member_base_types.push_back(
+        ^^typename member_base_generator<member>::member_base);
+  }
```
RyanJK5: This approach is a bit less noisy and reduces some of the heavyweight template instantiations that come from `template for` (may not be perfect):

```suggestion
  for (std::meta::info member : members_of(InterfaceType, std::meta::access_context::unprivileged()) |
                    std::views::filter(std::meta::is_function) |
                    std::views::filter(std::not_fn(std::meta::is_special_member_function)) |
                    std::views::filter(std::not_fn(std::meta::is_static_member)) |
                    std::views::filter(std::meta::has_identifier)) {
    member_base_types.push_back(dealias(substitute(^^member_base_generator_t, {reflect_constant(member)})));
  }
```

It would require adding:

```c++
template <std::meta::info Member>
using member_base_generator_t = member_base_generator<Member>::member_base
```

reflection/protocol_test.cc
```diff
@@ -394,4 +396,164 @@ TEST(ReflectionProtocolTest, IsConstructibleInPlaceWithArguments) {
...
+TEST(ReflectionProtocolViewTest, NonConstMemberFunctionNotInvocableFromConst) {
+  struct Interface {
+    void update(int value);
+  };
+
+  static_assert(
+      !std::is_invocable_v<
+          decltype((std::declval<const protocol_view<Interface>&>().update)),
+          int>);
+}
```
RyanJK5, line 423: Isn't this behavior incorrect? I thought `protocol_view` used shallow `const`, so a `const protocol_view` *should* be able to call `update`, whereas a `protocol_view<const Interface>` should not.

reflection/protocol.hh
```diff
@@ -93,6 +100,119 @@ consteval bool member_function_conforms_to(std::meta::info candidate,
...
+// conformance.
+template <std::meta::info Type>
+consteval auto protocol_interface_functions_of() {
+  return std::define_static_array(
+      std::define_static_array(
+          members_of(Type, std::meta::access_context::unprivileged())) |
+      std::views::filter(std::meta::is_function) |
+      std::views::filter(std::not_fn(std::meta::is_static_member)) |
+      std::views::filter(std::meta::has_identifier));
+}
```
RyanJK5: Since this is producing a static array anyway, I think it makes sense to make this a variable template. That way the results are cached if this is ever called again.

```suggestion
constexpr inline auto protocol_interface_functions_of = std::define_static_array(
      members_of(Type, std::meta::access_context::unprivileged()) |
      std::views::filter(std::meta::is_function) |
      std::views::filter(std::not_fn(std::meta::is_static_member)) |
      std::views::filter(std::meta::has_identifier));
```

NOTE: I also dropped the inner `define_static_array`, which I believe was unnecessary. Please add it back if I am mistaken.

reflection/protocol.hh
```diff
@@ -151,24 +266,25 @@ class protocol {
...
+  // `T`.
   template <typename U, typename... Ts>
     requires is_protocol_conformant_v<T, std::remove_cvref_t<U>> &&
              (!is_protocol_v<std::remove_cvref_t<U>>)
   explicit protocol(std::in_place_type_t<U>, Ts&&... ts);
 };
 
 template <typename T>
-class protocol_view {
+class protocol_view : public detail::protocol_stubs_t<T> {
```
RyanJK5, line 518: Related to my comment on the `NonConstMemberFunctionNotInvocableFromConst` test: I believe this is incorrect. `protocol_view<I>` should inherit from a slightly different version of `protocol_stubs_t` where every member function is `const` regardless of whether or not it was `const` in the interface. Later on, `protocol_view<const I>` would inherit from a `protocol_stubs_t` that filters out all of the non-`const` functions as well.

RyanJK5 (changes requested):
Good progress, I tagged a few things that I missed in my initial review.

reflection/protocol.hh
jbcoe: Again, I think that longer-term you are probably right. We can (and should) redesign later.

reflection/protocol_test.cc
jbcoe: I think that longer-term you are probably right. As the (somewhat underspecified) `DRAFT.md` stands, I believe that we're implementing the proposal as written. We can (and should) redesign later.

RyanJK5: Opened #186 for this.

RyanJK5 (approved):

PR #169, opened 2026-08-24
Rename protocol.h to protocol.hh to avoid confusing clang-format

jbcoe:


jbcoe:
@RyanJK5 this is mechanical but I think that you need to know about it as it will impact PRs in flight.

RyanJK5 (approved):

Issue #162, opened 2026-08-23
Consider handling of volatile for interface conformance


jbcoe:

RyanJK5:
I think we should reject any interface with volatile members.

jbcoe:
100% agreed. I think that this can be implemented relatively easily and reverted (design-wise) at no cost.

RyanJK5:
> 100% agreed. I think that this can be implemented relatively easily and reverted (design-wise) at no cost.

Do we care about failing early for this type of thing? E.g. should there be a constraint on the `Interface` type passed into `protocol` and `protocol_view` that it is a class type and has no volatile members?

This also would be relevant to making sure something like `protocol<int>` could be reasoned about at compile-time without triggering static assertions, e.g.:

```c++
// currently will say this is OK, but if you try to use it there will be compilation errors.
static_assert(can_substitute(^^protocol, {^^int}));
```

jbcoe:
Failing early sounds good to me.

We can probably tolerate volatile on member functions of conformant types but I've not thought too hard about it

Issue #97, opened 2026-07-24
Handle value category propagation


RyanJK5:
`protocol` already propagates const-ness: that is, a `const protocol<I>` implies that only the `const` members of `I` can be used.

An extension of this is propagating the object's value category, i.e. lvalue or rvalue. How should the following behave?

```c++
struct Interface {
    int foo() &&;
    int bar() &;
};

struct A {
    int foo() && { return 1; }
    int bar() & { return 2; }
};

protocol<Interface> p{A{}};

p.bar(); // OK
protocol<Interface>{A{}}.foo(); // OK?

p.foo(); // ERROR?
std::move(p).bar(); // ERROR?
```

From an implementation perspective, this would require adding overloads to `named_forwarder`, and require carefully persisting `T&` / `T&&` throughout.

This should probably be acknowledged in the draft, but the feature itself could potentially be deferred to a later extension of `protocol`.

jbcoe:
Thanks @RyanJK5 . We definitely want to do this. Worthwhile work is probably blocked on #94.

PR #174, opened 2026-08-25
Use vtable for protocol_view member function forwarding

jbcoe:
Adds a data-pointer and vtable-pointer member to `protocol_view` enabling member function forwarding.

`protocol` remains signature-only for member functions as ownership makes implementation slightly trickier.

Overloads are still unsupported. 

`protocol_view<const T>` needs to be added as constructing `protocol<T>` from a `const` object is (correctly) not allowed.

### DIFF
--- a/reflection/protocol.hh
+++ b/reflection/protocol.hh
@@ -173,7 +173,8 @@ consteval std::vector<std::meta::info> generate_method_thunk_specs(
     // Build the function-pointer type R(*)(Args...) from the method's
     // return type and parameter types.
     std::vector<std::meta::info> fn_args{dealias(return_type_of(member))};
-    for (std::meta::info parameter : parameters_of(member)) {
+    std::vector<std::meta::info> member_parameters = parameters_of(member);
+    for (std::meta::info parameter : member_parameters) {
       fn_args.push_back(dealias(type_of(parameter)));
     }
     std::meta::info fn_ptr_type = substitute(^^fn_ptr_t, fn_args);
--- a/reflection/protocol.hh
+++ b/reflection/protocol.hh
@@ -199,24 +199,24 @@ struct protocol_member_stubs_generator {
 
 }  // namespace detail
 
-// Returns `true` if `U` is a structural subtype of the interface `T`;
+// Returns `true` if `Candidate` is a structural subtype of `Interface`;
 // otherwise returns `false`.
-template <typename T, typename U>
+template <typename Interface, typename Candidate>
 consteval bool is_protocol_conformant() {
-  static_assert(std::is_same_v<T, std::remove_cvref_t<T>>,
-                "the interface type must not be cv/ref-qualified: strip "
-                "qualifiers at the call site with std::remove_cvref_t.");
-  static_assert(std::is_same_v<U, std::remove_cvref_t<U>>,
-                "the candidate type must not be cv/ref-qualified: strip "
-                "qualifiers at the call site with std::remove_cvref_t.");
+  static_assert(std::is_same_v<Interface, std::remove_cvref_t<Interface>>,
+                "Interface must not be cv/ref-qualified: strip qualifiers at "
+                "the call site with std::remove_cvref_t.");
+  static_assert(std::is_same_v<Candidate, std::remove_cvref_t<Candidate>>,
+                "Candidate must not be cv/ref-qualified: strip qualifiers at "
+                "the call site with std::remove_cvref_t.");
 
   // Checking for protocol interface conformance is O(N*M) over member counts,
   // assumed to be negligible at compile time.
   // TODO(jbcoe): Use set/map once there is library support for `constexpr`.
   auto interface_member_functions =
-      detail::protocol_interface_functions_of<^^T>();
+      detail::protocol_interface_functions_of<^^Interface>();
   auto candidate_member_functions =
-      detail::protocol_interface_functions_of<^^U>();
+      detail::protocol_interface_functions_of<^^Candidate>();
 
   return std::ranges::all_of(
       interface_member_functions, [&](std::meta::info interface_member) {
@@ -229,8 +229,9 @@ consteval bool is_protocol_conformant() {
 }
 
 // Variable template for use in requires clauses.
-template <typename T, typename U>
-inline constexpr bool is_protocol_conformant_v = is_protocol_conformant<T, U>();
+template <typename Interface, typename Candidate>
+inline constexpr bool is_protocol_conformant_v =
+    is_protocol_conformant<Interface, Candidate>();
 
 template <typename T, typename Allocator = std::allocator<std::byte>>
 class protocol : public detail::protocol_member_stubs_generator<T>::stubs {
--- a/reflection/protocol.hh
+++ b/reflection/protocol.hh
@@ -133,12 +133,10 @@ consteval std::meta::info find_vtable_member() {
   std::unreachable();
 }
 
-// ---------------------------------------------------------------------------
 // Vanishing-this-pointer thunk for a synthesised member function.
 //
 // The thunk carries a single operator() whose signature mirrors one method
 // of the Interface type.
-// ---------------------------------------------------------------------------
 template <typename FnPtrType, typename EnclosingType, typename ProtocolType,
           typename Vtable, std::meta::info Member, bool IsConst,
           bool IsNoexcept>
@@ -157,8 +155,7 @@ struct method_thunk<R (*)(Args...), EnclosingType, ProtocolType, Vtable, Member,
   // Provides member-function call syntax. Recovers the EnclosingType pointer
   // through the vanishing-this-pointer cast, widens it to the enclosing
   // protocol/protocol_view object, then calls through its stored vtable
-  // pointer's matching function pointer, passing the viewed/owned object
-  // (not the protocol/protocol_view wrapper itself).
+  // pointer's matching function pointer, passing the viewed/owned object.
   R operator()(Args... args) noexcept(IsNoexcept)
     requires(!IsConst)
   {
@@ -176,9 +173,6 @@ struct method_thunk<R (*)(Args...), EnclosingType, ProtocolType, Vtable, Member,
     const auto* protocol_object = static_cast<const ProtocolType*>(enclosing);
     const Vtable* vtable = protocol_object->vtable_;
     constexpr std::meta::info entry = vtable_entry();
-    // IsConst is true on this overload, so vtable_generator generated this
-    // entry with a leading `const void*` parameter; object_ (a plain
-    // void*) converts to that implicitly.
     return (*vtable).[:entry:](protocol_object->object_, args...);
   }
 };
@@ -188,8 +182,6 @@ using fn_ptr_t = R (*)(Args...) noexcept(Noexcept);
 
 // A single-member base wrapping the thunk for one interface member function,
 // named after that method (giving the `p.method_name(args)` call syntax).
-// `ProtocolType` and `Vtable` are threaded through to `method_thunk`; see its
-// comment for why they can't be recovered from `member_base` itself.
 template <std::meta::info Member, typename ProtocolType, typename Vtable>
 struct member_base_generator {
   struct member_base;
@@ -233,8 +225,7 @@ struct wrapper_bases : MemberBases... {};
 
 // Returns a `wrapper_bases` specialisation with one base per public,
 // non-special, member function of `interface_type`, giving named members
-// with `operator()` for each. `ProtocolType`/`Vtable` are forwarded to
-// `member_base_generator`.
+// with `operator()` for each.
 //
 // Two bases defining a member of the same name make that name ambiguous to
 // look up through the derived class, so overloaded methods are unsupported
@@ -256,15 +247,11 @@ consteval std::meta::info generate_wrapper_bases() {
 
 // The generated wrapper type for `T`: a `wrapper_bases` specialisation with
 // named members with `operator()` for each public, non-special, member
-// function from `T`. `ProtocolType` is the enclosing protocol/protocol_view
-// specialisation (protocol<T, Allocator> or protocol_view<T>) and `Vtable`
-// is its vtable_generator<T>::vtable: each thunk needs both to reach
-// ProtocolType's `vtable_` pointer and call through it.
+// function from `T`.
 template <typename T, typename ProtocolType, typename Vtable>
 using protocol_wrappers_t =
     typename[:generate_wrapper_bases<^^T, ProtocolType, Vtable>():];
 
-// ---------------------------------------------------------------------------
 // Returns a list of data_member_spec values, one for each member function
 // implemented by `protocol`, each describing a vtable function pointer with
 // signature R(*)(void*, Args...) for a mutable interface method, or
@@ -273,7 +260,6 @@ using protocol_wrappers_t =
 // Because C++ disallows two data members with the same name inside the same
 // class, overloaded methods will cause compile-time errors.
 // We will address this limitation in a follow-up PR.
-// ---------------------------------------------------------------------------
 consteval std::vector<std::meta::info> generate_vtable_specs(
     std::meta::info interface_type) {
   std::vector<std::meta::info> function_pointer_specs;
@@ -424,6 +410,11 @@ inline constexpr bool is_protocol_conformant_v =
 
 // ---------------------------------------------------------------------------
 // protocol<I, Allocator>
+//
+// WARNING: Vtable layouts in `protocol` and `detail::vtable_generator<I>::vtable`
+// are currently inconsistent. Calling member functions of interface `I` passes
+// signature checks at compile time, but triggers invalid `reinterpret_cast`
+// operations, undefined behavior, and potential segmentation faults at runtime.
 // ---------------------------------------------------------------------------
 template <typename I, typename Alloc = std::allocator<std::byte>>
 class protocol
@@ -692,6 +683,9 @@ class protocol
   constexpr bool valueless_after_move() const { return obj_ == nullptr; }
 };
 
+// ---------------------------------------------------------------------------
+// protocol_view<T>
+// ---------------------------------------------------------------------------
 template <typename T>
 class protocol_view
     : public detail::protocol_wrappers_t<
@@ -709,10 +703,6 @@ class protocol_view
   ~protocol_view() = default;
 
   // Construct from any non-const type U that conforms to the Interface T.
-  // U being const is rejected unconditionally, regardless of whether T
-  // actually declares any non-const methods: a simple, T-independent rule
-  // is easier to reason about than one that only rejects const U when it
-  // would actually be unsound.
   template <typename U>
     requires is_protocol_conformant_v<T, std::remove_cvref_t<U>> &&
                  (!is_protocol_view_v<std::remove_cvref_t<U>>) &&
--- a/reflection/protocol.hh
+++ b/reflection/protocol.hh
@@ -21,9 +21,23 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 #define XYZ_REFLECTION_PROTOCOL_HH_
 
 // A C++26-reflection-based implementation of protocol and protocol_view.
+//
+// Member function stubs are synthesised at compile time for every public
+// non-special member function declared in the Interface type.  The stubs
+// are attached to protocol and protocol_view through data members that
+// provide ordinary member-function call syntax via the "vanishing this
+// pointer" technique described in tutorials/2_vanishing_this_pointer.cc
+// (section 5): each per-method wrapper sits as the sole member of a
+// dedicated base struct; the wrapper's operator() recovers the enclosing
+// base address through a static_cast and hands it to the derived class.
+//
+// The stubs are intentionally unimplemented beyond the signature: they
+// call std::unreachable() so that the type-system plumbing can be
+// developed before the vtable dispatch layer exists.
 
 #include <algorithm>
 #include <cstddef>
+#include <functional>
 #include <memory>
 #include <meta>
 #include <ranges>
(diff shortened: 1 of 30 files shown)


RyanJK5:
Q: should constructing `protocol<Interface>` from a `const T` be allowed if `Interface` only has `const` member functions?

philipcraig:
> Q: should constructing `protocol<Interface>` from a `const T` be allowed if `Interface` only has `const` member functions?

Yes, and I would go further: constructing `protocol<Interface>` from a `const T` should be allowed regardless of whether `Interface` has non-const members. `protocol` copies the object into its own non-const storage, so the cv-qualification of the *source* is irrelevant; what matters is `is_copy_constructible_v<T>` (P3019's `polymorphic(U&&)` accepts const lvalues for the same reason). The reflection implementation on `main` already does this: the constraint is `is_protocol_conformant_v<T, std::remove_cvref_t<U>>`.

The Python implementation currently rejects `protocol<A>` from a `const ALike&` (because `protocol_concept<U>` is evaluated on the reference type) while accepting it for a const-only interface like `A_Subset` — so it answers your question "yes" by accident, and the rejection for interfaces with non-const members looks like a bug worth its own issue.

For views the situation is different, since there is no copy: the "reject const `U` unconditionally, use `protocol_view<const T>`" rule you have in this PR is the right one, and #186 (now implemented in #197) adds `protocol_view<const T>`.

RyanJK5:
> > Q: should constructing `protocol<Interface>` from a `const T` be allowed if `Interface` only has `const` member functions?
> 
> Yes, and I would go further: constructing `protocol<Interface>` from a `const T` should be allowed regardless of whether `Interface` has non-const members. `protocol` copies the object into its own non-const storage, so the cv-qualification of the _source_ is irrelevant; what matters is `is_copy_constructible_v<T>` (P3019's `polymorphic(U&&)` accepts const lvalues for the same reason). The reflection implementation on `main` already does this: the constraint is `is_protocol_conformant_v<T, std::remove_cvref_t<U>>`.
> 
> The Python implementation currently rejects `protocol<A>` from a `const ALike&` (because `protocol_concept<U>` is evaluated on the reference type) while accepting it for a const-only interface like `A_Subset` — so it answers your question "yes" by accident, and the rejection for interfaces with non-const members looks like a bug worth its own issue.
> 
> For views the situation is different, since there is no copy: the "reject const `U` unconditionally, use `protocol_view<const T>`" rule you have in this PR is the right one, and #186 (now implemented in #197) adds `protocol_view<const T>`.

Sorry, I misstated my question, it was supposed to be about `protocol_view`. Thanks for this.

philipcraig:
Filed the Python-generator const-source bug mentioned above as #198.

philipcraig:
Ah — for `protocol_view` the answer is less clear-cut, and it's worth deciding explicitly because the artefacts currently disagree on the "`Interface` has only const member functions" case:

- `DRAFT.md`'s synopsis accepts it structurally: `template <typename T> protocol_view(T& obj)` deduces `T = const U`; the comment says "mutable conforming object" but nothing enforces that.
- The Python generator accepts it (`protocol_view<A_Subset>` from a `const` object compiles today).
- This PR rejects it (`!std::is_const_v<U>`).
- `std::function_ref<R(Args...)>` is the closest precedent and accepts it: a const callable with a const `operator()` satisfies the non-const signature.

Both rules are sound — with an all-const interface there is nothing a `protocol_view<I>` could mutate. I'd still go with the rule in this PR: it's `T`-independent, it keeps `protocol_view<const I>` as the one spelling for "I view a const object", and a function taking `protocol_view<I>` doesn't change what callers may pass when a non-const member is later added to `I`. The price is that read-only parameters must be spelled `protocol_view<const I>`, which is the `span<const T>`/`string_view` habit anyway.

If we agree, the follow-ups are: constrain the `protocol_view(T&)` constructor in the paper so `T` is not const-qualified (can fold into #196), and align the generator (I'll extend #198 or file separately). If we'd rather follow `function_ref`, then this PR's constraint should become "`const U` conforms to `T`" instead of `!is_const_v<U>`, and #197's `protocol_view<const T>` becomes a convenience for interfaces that also have non-const members.

RyanJK5:
> Ah — for `protocol_view` the answer is less clear-cut, and it's worth deciding explicitly because the artefacts currently disagree on the "`Interface` has only const member functions" case:
> 
>     * `DRAFT.md`'s synopsis accepts it structurally: `template <typename T> protocol_view(T& obj)` deduces `T = const U`; the comment says "mutable conforming object" but nothing enforces that.
> 
>     * The Python generator accepts it (`protocol_view<A_Subset>` from a `const` object compiles today).
> 
>     * This PR rejects it (`!std::is_const_v<U>`).
> 
>     * `std::function_ref<R(Args...)>` is the closest precedent and accepts it: a const callable with a const `operator()` satisfies the non-const signature.
> 
> 
> Both rules are sound — with an all-const interface there is nothing a `protocol_view<I>` could mutate. I'd still go with the rule in this PR: it's `T`-independent, it keeps `protocol_view<const I>` as the one spelling for "I view a const object", and a function taking `protocol_view<I>` doesn't change what callers may pass when a non-const member is later added to `I`. The price is that read-only parameters must be spelled `protocol_view<const I>`, which is the `span<const T>`/`string_view` habit anyway.
> 
> If we agree, the follow-ups are: constrain the `protocol_view(T&)` constructor in the paper so `T` is not const-qualified (can fold into #196), and align the generator (I'll extend #198 or file separately). If we'd rather follow `function_ref`, then this PR's constraint should become "`const U` conforms to `T`" instead of `!is_const_v<U>`, and #197's `protocol_view<const T>` becomes a convenience for interfaces that also have non-const members.

Sounds good. I think we're refraining from making changes to the draft until we've implemented the current spec w/ reflection. `protocol_view<I>` injecting functions with `const` though seems like something worth adding to the implementation now and adding to the draft later, @jbcoe may have thoughts on this.

jbcoe:
I'd like to focus all _my_ attention on getting https://github.com/jbcoe/cc-protocol/pull/174 landed and then remove the Python implementation altogether. 

I'm very happy to have design issues discussed/solved/enumerated in the meantime; I just don't want design changes to slow down the implementation. It will be easy enough to update the working implementation.

philipcraig:
> I'd like to focus all _my_ attention on getting #174 landed and then remove the Python implementation altogether.
> 
> I'm very happy to have design issues discussed/solved/enumerated in the meantime; I just don't want design changes to slow down the implementation. It will be easy enough to update the working implementation.

Sounds good! First we implement, then we design :) The time-honoured order of things.

reflection/protocol.hh
```diff
@@ -113,55 +118,89 @@ constexpr inline auto protocol_interface_functions_of =
...
-struct method_thunk<R (*)(Args...), EnclosingType, IsConst, IsNoexcept> {
-  R operator()(Args... /*args*/) noexcept(IsNoexcept)
+template <typename R, typename... Args, typename EnclosingType,
+          typename ProtocolType, typename Vtable, std::meta::info Member,
+          bool IsConst, bool IsNoexcept>
+struct method_thunk<R (*)(Args...), EnclosingType, ProtocolType, Vtable, Member,
+                    IsConst, IsNoexcept> {
+  static consteval std::meta::info vtable_entry() {
+    return find_vtable_member<^^Vtable, Member>();
+  }
```
RyanJK5:
```suggestion
  static constexpr std::meta::info vtable_entry = find_vtable_member<^^Vtable, Member>();
```
Just so we instantiate it once instead of twice, simplifies the code a bit as well.

reflection/protocol.hh
```diff
@@ -113,55 +118,89 @@ constexpr inline auto protocol_interface_functions_of =
...
+template <std::meta::info Member, typename ProtocolType, typename Vtable>
 struct member_base_generator {
   struct member_base;
   consteval {
     std::string_view name = identifier_of(Member);
 
     // Build the function-pointer type R(*)(Args...) from the method's
     // return type and parameter types.
-    std::vector<std::meta::info> fn_args{dealias(return_type_of(Member))};
+    std::vector<std::meta::info> fn_args{std::meta::reflect_constant(false),
```
RyanJK5, line 197: Any reason not to pass in `std::meta::reflect_constant(is_noexcept(Member))`?

jbcoe:
Thanks @philipcraig !

reflection/protocol.hh
```diff
@@ -175,40 +214,163 @@ struct member_base_generator {
...
+// called with, then calls the matching member of `U`.
+template <typename FnPtrType, typename U, std::meta::info CandidateMember>
+struct mutable_view_trampoline;
+
+template <typename R, typename... Args, bool Noexcept, typename U,
+          std::meta::info CandidateMember>
+struct mutable_view_trampoline<R (*)(void*, Args...) noexcept(Noexcept), U,
+                               CandidateMember> {
+  static R call(void* ptr, Args... args) noexcept(Noexcept) {
+    return static_cast<U*>(ptr)->[:CandidateMember:](args...);
```
RyanJK5:
```suggestion
    return static_cast<U*>(ptr)->[:CandidateMember:](std::forward<Args>(args)...);
```
This should still forward, otherwise it will discard rvalue qualification.

reflection/protocol.hh
```diff
@@ -175,40 +214,163 @@ struct member_base_generator {
...
+
+template <typename FnPtrType, typename U, std::meta::info CandidateMember>
+struct const_view_trampoline;
+
+template <typename R, typename... Args, bool Noexcept, typename U,
+          std::meta::info CandidateMember>
+struct const_view_trampoline<R (*)(const void*, Args...) noexcept(Noexcept), U,
+                             CandidateMember> {
+  static R call(const void* ptr, Args... args) noexcept(Noexcept) {
+    return static_cast<const U*>(ptr)->[:CandidateMember:](args...);
```
RyanJK5:
```suggestion
    return static_cast<const U*>(ptr)->[:CandidateMember:](std::forward<Args>(args)...);
```

reflection/protocol.hh
```diff
@@ -113,55 +118,89 @@ constexpr inline auto protocol_interface_functions_of =
...
+  R operator()(Args... args) noexcept(IsNoexcept)
     requires(!IsConst)
   {
-    [[maybe_unused]] auto* base = reinterpret_cast<EnclosingType*>(this);
-    std::unreachable();  // vtable dispatch not yet implemented
+    auto* enclosing = reinterpret_cast<EnclosingType*>(this);
+    auto* protocol_object = static_cast<ProtocolType*>(enclosing);
+    const Vtable* vtable = protocol_object->vtable_;
+    constexpr std::meta::info entry = vtable_entry();
+    return (*vtable).[:entry:](protocol_object->object_, args...);
```
RyanJK5:
```suggestion
    return vtable->[:entry:](protocol_object->object_, args...);
```

reflection/protocol.hh
```diff
@@ -113,55 +118,89 @@ constexpr inline auto protocol_interface_functions_of =
...
     requires(IsConst)
   {
-    [[maybe_unused]] const auto* base =
-        reinterpret_cast<const EnclosingType*>(this);
-    std::unreachable();  // vtable dispatch not yet implemented
+    const auto* enclosing = reinterpret_cast<const EnclosingType*>(this);
+    const auto* protocol_object = static_cast<const ProtocolType*>(enclosing);
+    const Vtable* vtable = protocol_object->vtable_;
+    constexpr std::meta::info entry = vtable_entry();
+    return (*vtable).[:entry:](protocol_object->object_, args...);
```
RyanJK5:
```suggestion
    return vtable->[:entry:](protocol_object->object_, args...);
```

RyanJK5 (approved):
Great milestone!

philipcraig (commented):
Thanks — the vtable design is a clear improvement over the stubbed thunks.

I compiled probes against GCC trunk for each point below. Three are worth fixing before merge:

1. **Argument forwarding** — the thunk and both trampolines pass `args...` as lvalues, so `T&&` and move-only parameters don't compile and by-value parameters are copied three times.
2. **Detached `method_thunk`** — `auto f = view.get_value; f();` compiles and then reads through an unrelated object (ASan: stack-buffer-overflow at protocol.hh:174).
3. **`protocol<I>` regression** — deriving `protocol` from `protocol_wrappers_t` forces `vtable_generator<I>::vtable` to be defined, so interfaces with overloaded members that instantiated before this PR now hard-error, contrary to DRAFT.md.

The remaining comments are smaller (a misleading comment, `constexpr`/`noexcept` on the view constructor, dropped `static_assert`s, a duplicated enumeration pipeline, and the `volatile` constraint gap).

reflection/protocol.hh
philipcraig: `args...` is passed as an lvalue here (and at L176, L326, L338), so an interface parameter of type `T&&` or a move-only by-value type is a hard error, and copyable by-value parameters are copied three times per call:

```cpp
struct Interface { void take(std::string&&); };
p.take(std::string("x"));
// error: cannot bind rvalue reference of type 'std::string&&' to lvalue  (protocol.hh:166, :326)

struct Interface { void take(std::unique_ptr<int>); };
// error: use of deleted function unique_ptr(const unique_ptr&)
```

With a copy-counting type passed as an lvalue I measured copies=3 moves=0. `std::forward<Args>(args)...` fixes it, but it has to be applied in four near-identical bodies (two `operator()` overloads and two trampolines differing only in `const`); collapsing them into one helper / one trampoline template parameterised on the pointer type would make a partial fix impossible.

reflection/protocol.hh
```diff
@@ -113,55 +118,89 @@ constexpr inline auto protocol_interface_functions_of =
...
 
 // TODO(jbcoe): Extend this approach to handle lvalue and rvalue qualifiers.
-template <typename R, typename... Args, typename EnclosingType, bool IsConst,
-          bool IsNoexcept>
-struct method_thunk<R (*)(Args...), EnclosingType, IsConst, IsNoexcept> {
-  R operator()(Args... /*args*/) noexcept(IsNoexcept)
+template <typename R, typename... Args, typename EnclosingType,
+          typename ProtocolType, typename Vtable, std::meta::info Member,
+          bool IsConst, bool IsNoexcept>
+struct method_thunk<R (*)(Args...), EnclosingType, ProtocolType, Vtable, Member,
```
philipcraig, line 149: `method_thunk` is empty and freely copyable/default-constructible, but `operator()` only works when `this` is the `[[no_unique_address]]` subobject inside a `protocol_view`. Copying a named member out of the view compiles silently and then `reinterpret_cast`s an unrelated object:

```cpp
Conforming c; protocol_view<Interface> p(c);
auto getter = p.get_value;   // also: std::function<int()> f = p.get_value;
getter();                    // segfault; ASan: stack-buffer-overflow READ at protocol.hh:174
```

The pre-PR stub compiled this too but hit `std::unreachable()`; this PR makes the in-place path work while leaving the detached path reachable. Deleting the thunk's copy/move/default constructors turns all of these into compile errors.

reflection/protocol.hh
```diff
@@ -248,9 +410,17 @@ inline constexpr bool is_protocol_conformant_v =
...
+//
+// WARNING: Vtable layouts in `protocol` and
+// `detail::vtable_generator<I>::vtable` are currently inconsistent. Calling
+// member functions of interface `I` passes signature checks at compile time,
+// but triggers invalid `reinterpret_cast` operations, undefined behavior, and
+// potential segmentation faults at runtime.
 // ---------------------------------------------------------------------------
 template <typename I, typename Alloc = std::allocator<std::byte>>
-class protocol : public detail::protocol_stubs_t<I> {
+class protocol
```
philipcraig, line 419: Deriving `protocol` from `protocol_wrappers_t<I, protocol<I,Alloc>, vtable_generator<I>::vtable>` forces `define_aggregate` of `vtable_generator<I>::vtable` in the base clause even though `protocol` can't use it yet. That regresses interfaces with overloaded members, which DRAFT.md (L441-464, `protocol<OverloadedFunction>`) still lists as supported:

```cpp
struct Interface { int get() const; int get(int) const; };
protocol<Interface> p(Conforming{});   // never calls get()
// pre-PR: compiles and runs
// this PR: protocol.hh:302: name 'get' used in multiple data member descriptions
```

Suggest either not deriving `protocol` from the wrappers until it has a compatible vtable, or giving the thunk a uniform host accessor (`erased_object()` / `erased_vtable()`, or a traits struct) instead of reaching into `vtable_`/`object_` by name via friendship — then `protocol` and `protocol_view` can share one mechanism.

reflection/protocol.hh
philipcraig: `member_function_conforms_to` accepts (and requires matching) `&&`-qualified members, but the trampolines invoke the candidate on the lvalue `*static_cast<U*>(ptr)`, so an rvalue-ref-qualified interface member passes conformance and then fails here:

```cpp
struct Interface { int get() &&; };
struct Conforming { int get() && { return 3; } };
protocol_view<Interface> p(c); std::move(p).get();
// is_protocol_conformant_v == true, but:
// error: passing 'Conforming' as 'this' argument discards qualifiers (protocol.hh:326)
```

Either reject ref-qualified interface members up front with a clear `static_assert`, or call through `std::move(*static_cast<U*>(ptr))` when the member is `&&`-qualified. The TODO at L146 only mentions the thunk, not the trampoline.

reflection/protocol.hh
```diff
@@ -248,9 +410,17 @@ inline constexpr bool is_protocol_conformant_v =
 
 // ---------------------------------------------------------------------------
 // protocol<I, Allocator>
+//
+// WARNING: Vtable layouts in `protocol` and
```
philipcraig: This comment says a `protocol<I>` member call "passes signature checks at compile time, but triggers invalid reinterpret_cast operations, undefined behavior, and potential segmentation faults at runtime". It's actually ill-formed: the thunk reads `protocol`'s private `vtable_` (no friend declaration) whose type is `protocol`'s own nested `vtable`, not `vtable_generator<I>::vtable`, and reads `object_`, which `protocol` doesn't have (it has `obj_`):

```
protocol.hh:174: 'vtable_' is private within this context
protocol.hh:175: cannot convert 'const protocol<Interface>::vtable*' to 'const vtable_generator<Interface>::vtable*'
protocol.hh:176: has no member named 'object_'; did you mean 'obj_'?
```

Worth rewording so a reader doesn't think there's a runtime hazard. Note the surviving `static_assert(requires ...)` tests for `protocol` (protocol_test.cc L586-626) only check the thunk's signature, so they don't certify anything about `protocol`'s forwarding.

reflection/protocol.hh
```diff
@@ -528,11 +703,27 @@ class protocol_view : public detail::protocol_stubs_t<T> {
...
 
-  // Construct from any type U that conforms to the Interface T.
+  // Construct from any non-const type U that conforms to the Interface T.
   template <typename U>
     requires is_protocol_conformant_v<T, std::remove_cvref_t<U>> &&
-             (!is_protocol_view_v<std::remove_cvref_t<U>>)
-  explicit protocol_view(const U& object);
+                 (!is_protocol_view_v<std::remove_cvref_t<U>>) &&
+                 (!std::is_const_v<U>)
+  explicit protocol_view(U& object)
```
philipcraig, line 709: DRAFT.md's synopsis has `constexpr protocol_view(T& obj) noexcept;` (and `reference_interface_protocol.h` L687/L781 matches). This constructor is neither, though its body (`addressof` + address of a `constexpr` variable template) is trivially both, and the defaulted copy/move members on L700-703 are `noexcept`:

```cpp
static_assert(std::is_nothrow_constructible_v<protocol_view<Interface>, Conforming&>);  // fails
constexpr protocol_view<Interface> v(c);  // error: call to non-constexpr function
```

reflection/protocol.hh
```diff
@@ -528,11 +703,27 @@ class protocol_view : public detail::protocol_stubs_t<T> {
...
   ~protocol_view() = default;
 
-  // Construct from any type U that conforms to the Interface T.
+  // Construct from any non-const type U that conforms to the Interface T.
   template <typename U>
     requires is_protocol_conformant_v<T, std::remove_cvref_t<U>> &&
-             (!is_protocol_view_v<std::remove_cvref_t<U>>)
-  explicit protocol_view(const U& object);
+                 (!is_protocol_view_v<std::remove_cvref_t<U>>) &&
+                 (!std::is_const_v<U>)
```
philipcraig, line 708: The constraint only excludes `const U`, so a `volatile U` satisfies it and `is_constructible_v` reports true, but `static_cast<void*>(std::addressof(object))` is ill-formed for `volatile U*` — a hard error inside the body instead of a constraint failure:

```cpp
volatile Conforming c; protocol_view<Interface> p(c);
// is_constructible_v<protocol_view<Interface>, volatile Conforming&> == true, yet:
// error: invalid static_cast from type 'volatile Conforming*' to type 'void*' (protocol.hh:712)
```

`!std::is_const_v<U> && !std::is_volatile_v<U>` (or `std::is_same_v<U, std::remove_cv_t<U>>`) closes the gap.

reflection/protocol.hh
```diff
@@ -175,40 +214,163 @@ struct member_base_generator {
...
+//
+// Because C++ disallows two data members with the same name inside the same
+// class, overloaded methods will cause compile-time errors.
+// We will address this limitation in a follow-up PR.
+consteval std::vector<std::meta::info> generate_vtable_specs(
+    std::meta::info interface_type) {
+  std::vector<std::meta::info> function_pointer_specs;
+
+  std::ranges::range auto members =
+      std::define_static_array(members_of(
```
philipcraig: `generate_vtable_specs` re-implements the `members_of | is_function | !is_static_member | has_identifier` pipeline that `protocol_interface_functions_of<Type>` (L114-119) already provides, and `find_vtable_member`'s `std::unreachable()` (L133) relies on the two enumerations staying identical in membership and order. Applying the TODO at L112 (include static members) to one without mirroring it in the other makes `generate_wrapper_bases` and `vtable_generator` disagree — an opaque consteval hard error, or a thunk bound to the wrong slot if names still collide. Making `generate_vtable_specs` a template on `Type` that iterates `protocol_interface_functions_of<Type>` removes the duplication.

reflection/protocol_test.cc
```diff
@@ -427,43 +450,61 @@ TEST(ReflectionProtocolViewTest, SingleParameterMemberFunction) {
...
-  static_assert(requires(protocol_view<Interface>& p) {
-    { p.compute(0.0) } noexcept -> std::same_as<double>;
-  });
+  struct Conforming {
+    double compute(double input) noexcept { return input * 2.0; }
+  };
+
+  Conforming c;
+  protocol_view<Interface> p(c);
+  EXPECT_EQ(p.compute(21.0), 42.0);
```
philipcraig, line 477: Replacing the `static_assert(requires ... { p.compute(0.0) } noexcept -> std::same_as<double>)` blocks with `EXPECT_EQ` (here and in MultiParameterMemberFunction, VoidMemberFunction, MultipleMemberFunctions, ConstMemberFunction) drops every compile-time check that a `protocol_view` call is `noexcept` and returns exactly the interface's type; the only surviving ones (L586-626) are for `protocol<I>`. A regression in `IsNoexcept` (L205 -> L159/L169) or a merely-convertible return type would now pass the whole suite. The properties do currently hold (`static_assert(noexcept(p.compute(1.0))); static_assert(!noexcept(p.plain()));` compiles), so keeping the old `static_assert`s alongside the new runtime checks costs nothing.

reflection/protocol.hh
RyanJK5: Good catch.

reflection/protocol.hh
RyanJK5: This is already tracked as #97.

reflection/protocol.hh
RyanJK5: I believe volatility is being ignored for now. #162

reflection/protocol.hh
RyanJK5: > That regresses interfaces with overloaded members

The reflection implementation is not currently considering overloads, and did not before this PR.

jbcoe:
> Q: should constructing `protocol<Interface>` from a `const T` be allowed if `Interface` only has `const` member functions?

Let's handle this as follow-up. The DRAFT is somewhat underspecified.

reflection/protocol.hh
jbcoe: Good catch. 

Fixing this is noisy enough to be a follow-up PR.

reflection/protocol.hh
jbcoe: I think we're comfortable with understood limitations for now. This PR is relatively incremental and fits in my head.

jbcoe:
Thanks for this - lots of helpful insight here.

> Thanks — the vtable design is a clear improvement over the stubbed thunks.
> 
> I compiled probes against GCC trunk for each point below. Three are worth fixing before merge:
> 
> 1. **Argument forwarding** — the thunk and both trampolines pass `args...` as lvalues, so `T&&` and move-only parameters don't compile and by-value parameters are copied three times.

Fixed. Ideally we'd have tests for this too.

> 2. **Detached `method_thunk`** — `auto f = view.get_value; f();` compiles and then reads through an unrelated object (ASan: stack-buffer-overflow at protocol.hh:174).

This needs fixing but is messy and can be handled later.

> 3. **`protocol<I>` regression** — deriving `protocol` from `protocol_wrappers_t` forces `vtable_generator<I>::vtable` to be defined, so interfaces with overloaded members that instantiated before this PR now hard-error, contrary to DRAFT.md.

The earlier stubs did not support overloads, that's yet to be implemented so not a regression in this PR.

> 
> The remaining comments are smaller (a misleading comment, `constexpr`/`noexcept` on the view constructor, dropped `static_assert`s, a duplicated enumeration pipeline, and the `volatile` constraint gap).

philipcraig (approved):

Issue #39, opened 2026-03-29
clang-tidy CI checks are running checks that we don't need


jbcoe:
In logs I see:

/home/runner/work/cc-protocol/cc-protocol/build_default/_deps/google_benchmark-src/include/benchmark/benchmark.h:1855:5: warning: constructor does not initialize these fields: family_index, per_family_instance_index, repetition_index, repetitions, statistics [cppcoreguidelines-pro-type-member-init,hicpp-member-init]

That's a file we don't need to check and chekcs we don't need to run. This is not blocking.

https://productionresultssa19.blob.core.windows.net/actions-results/5f49eb59-89d5-42be-b279-c29d455f7166/workflow-job-run-1eacffd7-a62c-55db-8a49-f09780f59f9d/logs/job/job-logs.txt?rsct=text%2Fplain&se=2026-03-29T12%3A59%3A33Z&sig=H5SUMIHS5s6v8ut%2BoKfMm19ngsqQ%2BJmC%2BBSRFucvTVE%3D&ske=2026-03-29T16%3A01%3A53Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2026-03-29T12%3A01%3A53Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2026-03-29T12%3A49%3A28Z&sv=2025-11-05

philipcraig:
## Plan

The `_deps/google_benchmark` warnings quoted above were fixed by f77d58e (`HeaderFilterRegex: '^(?!.*_deps).*'`) and no longer appear. What remains is the underlying problem: `.clang-tidy` enables `*` and opts out, so each clang-tidy release adds new checks, and the run currently emits **~386 warnings** on `main` that nothing enforces (`WarningsAsErrors: ''`).

Baseline from the latest `main` run (clang-tidy 21):

| File | Warnings |
| --- | --- |
| `protocol_test.cc` | 227 |
| `tutorials/*.cc` | 145 |
| `protocol.cc` | 11 |
| `build/generated/protocol_A.h` | 2 |

[...]

PR #178, opened 2026-08-27
Reduce clang-tidy noise from checks the project does not need

philipcraig:
Step 1 of #39 (see the plan and checklist there). Configuration only, no source changes.

- Mark `googletest` and `google_benchmark` as `SYSTEM` in `FetchContent_Declare` so third-party headers are excluded regardless of `HeaderFilterRegex`.
- Extend `HeaderFilterRegex` to exclude `/generated/`; findings in generated code should be fixed in `scripts/protocol.j2`.
- Disable checks that contradict the design or are pure style noise, with rationale recorded in `.clang-tidy`:
  - `cppcoreguidelines-avoid-magic-numbers` (alias of the already-disabled `readability-magic-numbers`)
  - `readability-convert-member-functions-to-static` (interface structs and conforming types must have non-static members)
  - `google-explicit-constructor`, `hicpp-explicit-conversions` (implicit construction of `protocol` from a conforming type is by design)
  - `misc-const-correctness`, `modernize-use-nodiscard`
  - `llvm-prefer-static-over-anonymous-namespace`, `misc-use-internal-linkage` (conflict with Google style)
  - `readability-function-cognitive-complexity` (fires only on test bodies)
  - `clang-analyzer-cplusplus.Move` (duplicates `bugprone-use-after-move` but reports inside the generated callee, so it cannot be `NOLINT`ed at the test site)
  - `cppcoreguidelines-pro-bounds-avoid-unchecked-container-access` (new in clang-tidy 22; fires on every `operator[]` in tests)

## Verification

- `clang-tidy --verify-config` passes (the pre-existing `cert-dcl21-cpp` entry is unknown to clang-tidy 22 only; CI runs 21).
- Warnings on `main` drop from ~386 to ~100 measured locally with clang-tidy 22. Everything remaining is either a deliberate moved-from-object test or a genuine finding; both are step 2 of #39.
- `./scripts/cmake.sh --debug -DCLANG_TIDY_ENABLE=1`: builds, 87/87 tests pass.
- pre-commit clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01QHjiPBayDu7hLZyhcQK3tt

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,3 +1,23 @@
+# Checks are enabled with '*' and then opted out. Rationale for the less
+# obvious exclusions at the end of the list:
+#
+# - cppcoreguidelines-avoid-magic-numbers: alias of readability-magic-numbers,
+#   which is already disabled.
+# - readability-convert-member-functions-to-static: interface structs and
+#   conforming types must have non-static member functions; that is the whole
+#   point of structural subtyping.
+# - google-explicit-constructor, hicpp-explicit-conversions: implicit
+#   construction of protocol from a conforming type is by design.
+# - misc-const-correctness, modernize-use-nodiscard: style churn with no
+#   bug-finding value in a reference implementation.
+# - llvm-prefer-static-over-anonymous-namespace, misc-use-internal-linkage:
+#   conflict with Google style, which prefers anonymous namespaces.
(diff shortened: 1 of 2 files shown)


jbcoe (approved):
Thanks for this!

PR #179, opened 2026-08-27
Suppress intentional clang-tidy findings and fix the genuine ones

philipcraig:
Step 2 of #39, stacked on #178 (merge that first; this PR's base is `reduce-clang-tidy-noise`). Takes the clang-tidy run from ~70 warnings to zero.

## Suppressions

- Tests and tutorials that deliberately exercise the moved-from state are wrapped in `NOLINTBEGIN`/`NOLINTEND` for `bugprone-use-after-move`, so the check stays active everywhere else.
- New `tutorials/.clang-tidy` (`InheritParentConfig: true`) disables the checks that object to the non-`protocol` baseline the tutorials deliberately show: classic inheritance, hand-written vtables with raw casts, manual allocation.
- Remaining false positives are suppressed at the site with a reason: `unique_ptr<char[]>` (avoid-c-arrays), copies a test measures, `operator=(int)` mirroring `interface_D.h`, and move constructors that take resources with `std::exchange` rather than `std::move`.

## CMake fix (needed for the per-directory config)

`FindClangTidy.cmake` passed `--config-file=<root>/.clang-tidy`, which disables clang-tidy's per-directory lookup, so `tutorials/.clang-tidy` was never read. Dropping it exposed a second bug: with no extra arguments the command list ended in `;`, and clang-tidy treated the empty argument as a source file (`expected exactly one compiler job in ''`). Both fixed.

## Genuine fixes

`std::scoped_lock` in `protocol.cc`; `CopyCounter` gains a destructor and self-assignment guard; explicit pointer/integer comparisons instead of implicit bool conversions; `vector::reserve` before `emplace_back` loops; `emplace_back` for temporaries; member initialiser for `AnimalPtr::noise_func_`; `auto*` for pointer-typed `auto`.

## Verification

- clang-tidy 22 locally on `protocol.cc`, `protocol_test.cc`, `tutorials/*.cc`: **0 warnings**; `--verify-config` clean in both directories.
- `./scripts/cmake.sh --debug` with `CLANG_TIDY_ENABLE=1`: builds, 87/87 tests pass.
- pre-commit clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01QHjiPBayDu7hLZyhcQK3tt

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,3 +1,23 @@
+# Checks are enabled with '*' and then opted out. Rationale for the less
+# obvious exclusions at the end of the list:
+#
+# - cppcoreguidelines-avoid-magic-numbers: alias of readability-magic-numbers,
+#   which is already disabled.
+# - readability-convert-member-functions-to-static: interface structs and
+#   conforming types must have non-static member functions; that is the whole
+#   point of structural subtyping.
+# - google-explicit-constructor, hicpp-explicit-conversions: implicit
+#   construction of protocol from a conforming type is by design.
+# - misc-const-correctness, modernize-use-nodiscard: style churn with no
+#   bug-finding value in a reference implementation.
+# - llvm-prefer-static-over-anonymous-namespace, misc-use-internal-linkage:
+#   conflict with Google style, which prefers anonymous namespaces.
+# - readability-function-cognitive-complexity: fires only on test bodies.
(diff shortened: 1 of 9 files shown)


jbcoe (approved):
Thanks for this!

PR #190, opened 2026-08-28
Retarget the C++26 reflection build from gcc-16 to a GCC trunk snapshot

### LINKED ISSUE #122: Update gcc 16 to gcc-trunk
Reflection in GCC 16 is no longer receiving updates, since it was an experimental feature at the time. Now that it is being added to gcc-trunk, some bugs are being surfaced and fixed, e.g.:

```c++
#include <meta>

template <typename T>
struct S {};

// Fails on gcc 16, works on gcc-trunk
static_assert(can_substitute(^^S, {^^int&()}));
```

The docker container that uses 16 should be bumped to retarget trunk. To avoid building from source, https://jwakely.github.io/pkg-gcc-latest/ can be used.

philipcraig:
Fixes #122.

Ubuntu 26.04's `gcc-16` is a frozen experimental snapshot whose reflection support no longer receives fixes (e.g. `can_substitute(^^S, {^^int&()})` fails on gcc-16 and passes on trunk). This retargets the C++26 reflection build to Jonathan Wakely's prebuilt GCC trunk package (https://jwakely.github.io/pkg-gcc-latest/, currently GCC 17.0.0-20260816).

## Changes

- **`docker/Dockerfile`** — on amd64, installs `gcc-latest.deb` to `/opt/gcc-latest`. The package is x86_64-only, so arm64 keeps `gcc-16` as the fallback. `gcc-16` stays installed on all architectures.
- **`scripts/cmake.py`** — `--reflection` prefers `/opt/gcc-latest/bin/g++` when present, else `g++-16`; an existing `CXX`/`CC` in the environment overrides both.
- **`CMakeLists.txt`** — with reflection on, embeds the compiler's `libstdc++` directory (from `g++ -print-file-name=libstdc++.so`) as a build rpath, so executables linked against the snapshot's `/opt/gcc-latest/lib64` run without `LD_LIBRARY_PATH`.
- **`.github/workflows/cmake.yml`** — the reflection job now builds with GCC trunk and is the gating one. `gcc-16` is kept as a **non-blocking, informational** job (`continue-on-error`) because it is what a stock Ubuntu 26.04 user gets from `apt install gcc-16`; drop it once the backend needs trunk-only fixes. A comment notes how to temporarily pin a versioned snapshot if trunk regresses.
- `scripts/agentic-sandbox.py` — hint text.

## Verification (local, Ubuntu 26.04 devcontainer, `.deb` extracted with `dpkg-deb -x`)

- `CXX=…/gcc-latest/bin/g++ ./scripts/cmake.sh --release --reflection`: compiler identified as GNU 17.0.0, **162/162 tests pass** (including `reflection_protocol_test` and `reflection_tutorial`); `readelf -d` shows `RUNPATH` pointing at the snapshot's `lib64` and `ldd` resolves its `libstdc++.so.6`.
- `./scripts/cmake.sh --release --reflection` with no override (no `/opt/gcc-latest`): falls back to GNU 16.0.1, 162/162 pass.
- Trunk still requires `-freflection`, so the existing flags are unchanged.
- pre-commit (yaml, dockerfilelint, ruff, action-validator) and `pytest scripts` pass.

## Verification (after devcontainer rebuild, 2026-08-28)

- The Dockerfile layer installs correctly: `/opt/gcc-latest/bin/g++` reports GCC 17.0.0 20260816, and `./scripts/cmake.sh --release --reflection` now picks it up with no `CXX` override — GNU 17.0.0, 162/162 tests pass.
- All 47 CI checks pass, including both gating "Ubuntu GCC trunk (C++26 Reflection)" jobs (exercising the runner-side `.deb` install) and the two informational GCC-16 jobs.

## Notes

- The unversioned `.deb` URL tracks upstream weekly, so the image and the trunk CI job are intentionally unpinned.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -31,6 +31,7 @@ jobs:
   build:
     name: ${{ matrix.settings.name }} ${{ matrix.configuration }}
     runs-on: ${{ matrix.settings.os }}
+    continue-on-error: ${{ matrix.settings.informational || false }}
     strategy:
       fail-fast: false
       matrix:
@@ -238,7 +239,26 @@ jobs:
                 },
             }
           - {
-              name: "Ubuntu GCC-16 (C++26 Reflection)",
+              name: "Ubuntu GCC trunk (C++26 Reflection)",
+              os: ubuntu-26.04,
+              compiler:
+                {
+                  type: GCC_TRUNK,
+                  version: trunk,
+                  cc: "/opt/gcc-latest/bin/gcc",
+                  cxx: "/opt/gcc-latest/bin/g++",
+                  std: 26,
+                },
+              reflection: true,
+            }
+          # Ubuntu 26.04's gcc-16 is a frozen experimental snapshot whose
+          # reflection support no longer receives fixes (issue #122). It is kept
+          # here as a non-blocking, informational job because it is what a stock
+          # Ubuntu 26.04 user gets from `apt install gcc-16`; the GCC trunk job
+          # above is the gating one. Drop this entry once the reflection backend
+          # needs fixes that only exist on trunk.
+          - {
+              name: "Ubuntu GCC-16 (C++26 Reflection, informational)",
               os: ubuntu-26.04,
               compiler:
                 {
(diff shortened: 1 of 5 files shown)


jbcoe (commented):
Thanks for this.

Issue #29, opened 2026-03-26
Which reflection capabilities beyond C++ 26 are needed to implement this purely in C++


philipcraig:
# C++ Reflection Requirements for `cc-protocol`

An analysis of the reflection and code-injection capabilities that future C++ standards must provide in order to implement the `cc-protocol` library purely within the language, eliminating the need for external Python/Jinja2 code generation.

## What `cc-protocol` does

`cc-protocol` implements structural subtyping (duck typing) with value semantics for C++. Given a plain struct that declares an interface through its member function signatures (no virtual functions, no base classes), the library generates:

- **`protocol<I, Allocator>`**: An owning, type-erased wrapper with deep-copy value semantics, const-propagation, allocator awareness, and a valueless-after-move state. Any type that structurally matches the interface (i.e., provides member functions with compatible signatures) can be stored in it.

- **`protocol_view<I>`**: A non-owning, lightweight reference (analogous to `string_view`) for zero-allocation duck typing at function boundaries.

- **`protocol_view<const I>`**: A const-restricted view that only exposes const-qualified methods.

- **C++20 concepts**: `protocol_concept_I` and `protocol_const_concept_I` for compile-time structural validation with clear error messages.

The library provides two implementation strategies: one using virtual inheritance for dispatch, and one using manually constructed static vtables with function pointers. Both are generated from the same interface definition.

## How the code generator works today

The current pipeline is:

1. Parse an interface header (e.g., `interface_A.h`) using libclang via `py_cppmodel`
2. Extract class metadata: name, namespace, and for each member function: name, return type, parameter types and names, const-qualification
3. Generate MD5-based GUIDs for each method signature (to disambiguate overloads in the vtable)
4. Feed the metadata into a Jinja2 template that emits a complete C++ header file containing all the generated types

### 1. Member function introspection

**What:** Given a class type, enumerate all its declared member functions and extract for each: the function name, return type, each parameter's type, and whether the function is const-qualified.

**Why:** This is the foundational operation. Every generated construct — concepts, control blocks, forwarding functions, vtables — is driven by iterating over the interface's member functions.

**Current code-gen equivalent:**
```python
for m in target_class.methods:
    m.name            # function name
    m.return_type.name  # return type
    m.arguments         # parameter list
    m.is_const          # const-qualification
```

[...]

PR #191, opened 2026-08-28
Rewrite the paper's reflection appendix around the C++26 implementation

philipcraig:
Follow-up to the conclusion in #29 that C++26 reflection is sufficient and no post-C++26 code injection is needed.

- Replace the hypothetical-injection sketch in Appendix A with the design used by the reference implementation (`define_aggregate` vtable, spliced trampolines, `[[no_unique_address]]` thunk bases for call syntax), a worked example, and a remaining-work list. The appendix code, extracted verbatim, compiles with `g++ -std=c++26 -freflection -Wall -Wextra` on the GCC trunk snapshot and the example runs.
- Remove the code-injection requirement from the Abstract, Impact on the Standard and Reference Implementation sections; the latter now describes both implementations.
- Bump the header to P4148R3 with a placeholder date and add an R3 history entry.
- Remove a duplicated acknowledgements paragraph and a duplicated R2 history line.

Refs #29.


jbcoe:
Thanks for this. I'll hold off landing it but will use it for an impending end-to-end rewrite.

philipcraig:
> Thanks for this. I'll hold off landing it but will use it for an impending end-to-end rewrite.

Sounds good. I'll close so it doesn't skulk as an uneven PR

Issue #186, opened 2026-08-28
Change protocol_view interaction with const member functions


RyanJK5:
A `protocol_view<I>` should inject all the member functions of `I` as `const` member functions. It should NOT preserve the cvref-qualification of `I`. This is because `protocol_view` uses shallow `const` and could be cheaply copied to bypass the `const` qualification anyway.

`protocol_view<const I>` should be used instead, and should only inject the `const` interface of `I`.

philipcraig:
Agreed, and this is mostly a spec bug: `scripts/protocol.j2` already generates every member of `protocol_view<I>` as `const` and only the const members for `protocol_view<const I>` (see `int count() const` in the generated `protocol_view<A>`), so the Python implementation and its tests are already the oracle. `DRAFT.md` is self-inconsistent — the prose says views keep `I`'s const-qualification, the `protocol_view<I>` synopsis shows a mix, and the `protocol_view<const I>` synopsis lists a non-const member — which is why the reflection implementation went the other way.

Plan:
1. #195 / #196: fix the paper — state shallow const explicitly and make both synopses consistent.
2. #197: reflection implementation — `protocol_view<T>` stubs all `const`, new `protocol_view<const T>` specialisation exposing only const members, `protocol<T>` unchanged. Small and on `main` first so #174 can rebase onto it rather than re-deriving the rule.
3. #97 (ref-qualifiers) should follow the same "views don't propagate" principle; leaving it separate.

PR #196, opened 2026-08-28
Specify shallow const for protocol_view in DRAFT.md

### LINKED ISSUE #195: DRAFT.md: state shallow const for protocol_view and fix the view synopses
Sub-issue of #186 covering the paper.

`DRAFT.md` currently contradicts itself on `protocol_view` const-qualification:

- The "Generated structural subtyping" prose says `protocol` and `protocol_view` implement members "with identical constexpr, noexcept and const-qualification".
- The `protocol_view<I>` synopsis shows `func1(double) const` (non-const in `I`) next to `func2(int)` (non-const), so it matches neither reading.
- The `protocol_view<const I>` synopsis lists `func1(double) const`, which is not a const member of `I` and should not be there.

The Python reference implementation already generates every member of `protocol_view<I>` as `const` and only the const members for `protocol_view<const I>`; the reflection implementation followed the prose. The paper should state the rule from #186 explicitly (views have shallow const, like `span`/`string_view`; const-ness of the viewed object lives in the type) and make both synopses consistent with it.

philipcraig:
Fixes #195. Paper half of #186.

- Replace "identical constexpr, noexcept and const-qualification" with two paragraphs: `protocol` propagates const; `protocol_view` has shallow const, so every member of `protocol_view<I>` is const-qualified and `protocol_view<const I>` exposes only the const members of `I`, with the `span`/`string_view` rationale.
- `protocol_view<I>` synopsis: `func2` overloads are now `const`.
- `protocol_view<const I>` synopsis: drop `func1`, which is non-const in `I`.

This matches what `scripts/protocol.j2` already generates. No revision-history entry added since the document is still labelled R2; happy to add one under a new R3 heading if that is the convention.

### DIFF
--- a/DRAFT.md
+++ b/DRAFT.md
@@ -132,7 +132,22 @@ reference semantics to structural sub-types.
 
 For a given struct, the corresponding `protocol` and `protocol_view` will
 implement all the public non-virtual, non-template member functions with
-identical constexpr, noexcept and const-qualification.
+identical constexpr and noexcept qualification.
+
+`protocol` owns its object and propagates const: a member function of
+`protocol<I>` is const-qualified exactly when the corresponding member function
+of `I` is, so only the const member functions of `I` can be called through a
+`const protocol<I>`.
+
+`protocol_view` is a non-owning reference and, like `span` and `string_view`,
+has shallow const. Every member function of `protocol_view<I>` is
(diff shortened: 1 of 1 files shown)


jbcoe (approved):
Thanks for this. Shallow const, after sleep and a run, is the only approach that makes sense.

PR #197, opened 2026-08-28
Give protocol_view shallow const in the reflection implementation

### LINKED ISSUE #186: Change protocol_view interaction with const member functions

philipcraig:
Fixes #186. Implements the rule specified in #196.

- Add `detail::const_policy { preserve, all_const, const_only }` and thread it through `member_base_generator`, `generate_stub_bases` and `protocol_stubs_t` (default `preserve`, so `protocol<I>` is unchanged and still propagates const).
- `protocol_view<T>` now derives from `protocol_stubs_t<T, all_const>`: every stub is `const`, so a `const protocol_view<T>` exposes the full interface.
- New `protocol_view<const T>` partial specialisation derives from `protocol_stubs_t<T, const_only>` and is constructible from a `const U&`.
- Tests: `NonConstMemberFunctionNotInvocableFromConst` for the view is replaced by `NonConstMemberFunctionInvocableFromConstView`; added `ConstViewIsConstructibleFromConstObject`, `ConstViewExposesOnlyConstMemberFunctions`, `ConstViewOfInterfaceWithNoConstMembers`; `is_protocol_view_v` and special-member checks cover `protocol_view<const T>`. The `protocol` const-propagation test gains positive/negative `has_update` checks.

Verified with `./scripts/cmake.sh --release --reflection` on GCC trunk: 165/165 pass. Pre-commit clean.

Constructors are left as declarations, as on `main`. #174 will need a rebase: its `IsConst`-preserving thunk should take the policy-selected constness, and its "`protocol_view<const T>` needs to be added" note is covered here.

### DIFF
--- a/reflection/protocol.hh
+++ b/reflection/protocol.hh
@@ -143,9 +143,23 @@ struct method_thunk<R (*)(Args...), EnclosingType, IsConst, IsNoexcept> {
 template <typename R, typename... Args>
 using fn_ptr_t = R (*)(Args...);
 
+// How generated stubs treat the const-qualification of interface members.
+enum class const_policy {
+  // `protocol<I>`: as declared in `I`, so `const protocol<I>` exposes only the
+  // const member functions of `I` (const propagates).
+  preserve,
+  // `protocol_view<I>`: every stub is const-qualified regardless of `I`
+  // (shallow const, as for `std::span`).
+  all_const,
+  // `protocol_view<const I>`: only the const member functions of `I`.
+  const_only,
+};
+
 // A single-member base wrapping the thunk for one interface member function,
 // named after that method (giving the `p.method_name(args)` call syntax).
-template <std::meta::info Member>
+// `IsConst` selects the const-qualification of the generated stub, which may
+// differ from that of `Member` (see `const_policy`).
+template <std::meta::info Member, bool IsConst>
 struct member_base_generator {
   struct member_base;
   consteval {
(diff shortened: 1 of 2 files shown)


jbcoe:
I'm 90% sure that we want this but I'm not sure how shallow-const works for `protocol_view<T>` if `T` is:

```
struct T {
  int foo() const;
  int foo();
};
```

I'd like to land #174 first. We must then revisit this PR.

jbcoe (changes requested):
Needs rebasing onto a landed #174 after some design discussion.

I love _how_ this PR does what it does. I'm not sure _if_ we want to do it.

RyanJK5:
> I'm 90% sure that we want this but I'm not sure how shallow-const works for `protocol_view<T>` if `T` is:
> 
> ```
> struct T {
>   int foo() const;
>   int foo();
> };
> ```
> 
> I'd like to land #174 first. We must then revisit this PR.

My opinion (based on what `duck` does) is that `protocol_view<T>` should dispatch to `int foo()` but surface a `const` forwarder, while `protocol_view<const T>` should dispatch to `int foo() const` (and still use a `const` forwarder).

philipcraig:
Superseded by #204, rebuilt on top of #174.

jbcoe:
I'm down to 50% sure that we want this. Still loving how it works. 

On the one hand, `protocol_view` is a pointer (really) and pointers don't propagate constness. 

On the other hand, having `protocol` and `protocol_view` display surprisingly (maybe) different behaviour might be hard to stomach.

PR #204, opened 2026-08-28
Give protocol_view shallow const in the reflection implementation

### LINKED ISSUE #186: Change protocol_view interaction with const member functions

philipcraig:
Fixes #186 and implements the rule specified in #196.

`protocol_view<T>` has shallow const: a const view can still call non-const member functions on the object it refers to.

`protocol_view<const T>` is a new specialisation for viewing a const object, constructible only from a const reference and exposing only const member functions through its vtable.

A new const policy generates both cases in the wrapper/vtable generators without touching the existing behaviour used by `protocol<I>`.

Adds tests covering construction from const objects, const-only member exposure, and runtime dispatch through both kinds of view.

### DIFF
--- a/reflection/protocol.hh
+++ b/reflection/protocol.hh
@@ -188,7 +188,7 @@ using fn_ptr_t = R (*)(Args...) noexcept(Noexcept);
 enum class const_policy {
   // `protocol<I>`: as declared in `I`, so `const protocol<I>` exposes only the
   // const member functions of `I` (const propagates).
-  preserve,
+  propagate,
   // `protocol_view<I>`: every wrapper is const-qualified regardless of `I`
   // (shallow const, as for `std::span`).
   all_const,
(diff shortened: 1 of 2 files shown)


jbcoe:
I'm down to 50% sure that we want this. Still loving how it works.

On the one hand, protocol_view is a pointer (really) and pointers don't propagate constness.

On the other hand, having protocol and protocol_view display surprisingly (maybe) different behaviour might be hard to stomach.

RyanJK5:
> I'm down to 50% sure that we want this. Still loving how it works.
> 
> On the one hand, protocol_view is a pointer (really) and pointers don't propagate constness.
> 
> On the other hand, having protocol and protocol_view display surprisingly (maybe) different behaviour might be hard to stomach.

If I were to make the consistency argument, I would consider [`std::function_ref`](https://en.cppreference.com/cpp/utility/functional/function_ref) to be some prior art on this. Its `operator()` is always `const`, and the `const`-ness of the function stored within it is specified in the template parameter.

jbcoe:
I think that there will be sufficient educational differences between `protocol` and `protocol_view` anyway for different handling of const to be tolerable. It feels like the only sensible approach.

jbcoe:
50% will probably be 100% after a morning run 🏃‍♂️

jbcoe:
@philipcraig Can we clean up the PR description? It will become the squashed commit message and is currently a bit verbose.

jbcoe (changes requested):
The PR description, that will become the commit message is a bit too noisy. I suggest something like:

"
Give protocol_view shallow const in the reflection implementation

Fixes #186 and implements the rule specified in #196.

protocol_view<T> has shallow const: a const view can still call non-const member functions on the object it refers to.

protocol_view<const T> is a new specialisation for viewing a const object, constructible only from a const reference and exposing only const member functions through its vtable.
  
A new const policy generates both cases in the wrapper/vtable generators without touching the existing behaviour used by protocol<I>.

Adds tests covering construction from const objects, const-only member exposure, and runtime dispatch through both kinds of view.
"

jbcoe (approved):

PR #206, opened 2026-08-28
Fix delegating in-place constructors

RyanJK5:
Fixes a typo from the allocator implementation.

### DIFF
--- a/reflection/protocol.hh
+++ b/reflection/protocol.hh
@@ -529,7 +529,8 @@ class protocol
              is_protocol_conformant_v<I, T> &&
              std::default_initializable<Alloc>)
   constexpr explicit protocol(std::in_place_type_t<T>, Args&&... args)
-      : protocol(std::allocator_arg, Alloc{}, std::forward<Args>(args)...) {}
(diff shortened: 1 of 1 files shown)


jbcoe (approved):
Thanks for this. It might be nice to have tests that ensure that the right constructor is getting picked up. Not blocking though.

RyanJK5:
> Thanks for this. It might be nice to have tests that ensure that the right constructor is getting picked up. Not blocking though.

I agree, but am not sure the best place to put it. The allocator_tests file is currently only testing allocator-aware constructors, but it seems like protocol_test is more for the actual dispatch functions. Happy to put it wherever you think is best.

jbcoe:
> > Thanks for this. It might be nice to have tests that ensure that the right constructor is getting picked up. Not blocking though.
> 
> I agree, but am not sure the best place to put it. The allocator_tests file is currently only testing allocator-aware constructors, but it seems like protocol_test is more for the actual dispatch functions. Happy to put it wherever you think is best.

Let's leave it for now. It's a late-stage stress test and this PR fixes a clear typo.

PR #207, opened 2026-08-28
Forward member functions through an owning vtable in protocol

### LINKED ISSUE #201: Add method forwarding through the vtable to protocol
Add an owning / non-owning vtable to the reflection implementation so that `protocol` works with method forwarding.

philipcraig:
Adds an owning vtable to the reflection implementation so `protocol` forwards member function calls, as `protocol_view` already does.

`protocol`'s private `vtable` now inherits from `detail::vtable_generator<I>::vtable`, the generated per-member-function vtable `protocol_view` already uses, and adds the `destroy`/`copy`/`move` entries. Existing trampolines are reused as-is via the derived-to-base pointer conversion. Overloads and `protocol_view<const T>` remain out of scope.

Adds runtime forwarding tests for mixed const/mutating members, copy/move/swap, in-place construction, an allocator, and calls through a `const protocol`.

Closes #201.

### DIFF
--- a/reflection/protocol.hh
+++ b/reflection/protocol.hh
@@ -22,12 +22,10 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 // A C++26-reflection-based implementation of protocol and protocol_view.
 //
-// protocol_view has a minimal but working implementation: functions are
-// dispatched at runtime using a vtable.
-//
-// protocol currently supports only compile-time signature checks for member
-// functions; ownership (construction, copy, move, destruction) is
-// allocator-aware and implemented.
+// Both `protocol` and `protocol_view` dispatch member function calls at
+// runtime through a generated vtable; `protocol`'s vtable extends
+// `protocol_view`'s with destroy/copy/move entries used for allocator-aware
+// ownership.
 //
 // Neither implementation currently supports overloaded member functions or
 // operators.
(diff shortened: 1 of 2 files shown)


jbcoe (changes requested):
Can we clean up the PR description? It will become the squashed commit message and is currently a bit verbose.

RyanJK5:
I am not fully sure if using two different types is the best approach for the vtables, it will depend on how we implement narrowing. If we do narrowing by laying out pointers in the vtable ahead of time, the owning vtable will have to lay out pointers to the owning subsets as well, which means it would have to be generated via define_aggregate and therefore could not inherit from the view vtable.

jbcoe (approved):
Thanks for this.

jbcoe:
> I am not fully sure if using two different types is the best approach for the vtables, it will depend on how we implement narrowing. If we do narrowing by laying out pointers in the vtable ahead of time, the owning vtable will have to lay out pointers to the owning subsets as well, which means it would have to be generated via define_aggregate and therefore could not inherit from the view vtable.

This change is easy to follow and gives us functionality we need. I'd prefer not to block progress, there's some design work to do around narrowing and we can revisit this implementation as and when needed.

jbcoe (changes requested):
Can we clean up the PR description? It will become the squashed commit message and is currently a bit verbose.

Maybe something like:

"
Forward member functions through an owning vtable in protocol.

Extends `protocol`'s owning vtable so it also forwards member function calls.

`protocol`'s `vtable` now inherits from `detail::vtable_generator<I>::vtable`, the generated per-member-function vtable used elsewhere in the reflection implementation, adding member function dispatch alongside its existing `destroy`/`copy`/`move` entries. Existing trampolines are reused as-is via the derived-to-base pointer conversion.

Adds runtime forwarding tests for mixed const/mutating members, copy/move/swap, in-place construction, an allocator, and calls through a `const protocol`.

Closes #201.
"

jbcoe (approved):

PR #212, opened 2026-08-29
Make synthesised member thunks non-copyable

### LINKED ISSUE #202: Prevent callable members from being copied.
The reflection implementation uses callable members with operator() to match the signature of member functions. These members can currently be copied, calling copies will fail (UB) as they are no longer attached to a wrapper.

philipcraig:
Fixes #202.

The synthesised callable members in `reflection/protocol.hh` recover their enclosing `protocol`/`protocol_view` from `this`, so a copied member is detached and calling it is undefined behaviour.

- `detail::method_thunk`: delete the copy and move constructors and assignment operators (keeping a defaulted default constructor so the wrapper bases stay default-initialisable). `auto f = view.foo;` is now a compile error.
- `protocol_view`: replace the defaulted copy/move operations, which the change above would have deleted, with user-provided ones that copy `object_`/`vtable_` and leave the thunk bases default-initialised. Still `noexcept`.
- `protocol`: no change needed; its special member functions already never touch the wrapper bases.
- Tests: the member thunks of `protocol` and `protocol_view` are not copyable, movable or assignable; `protocol_view` copy/move remain `noexcept` and copies dispatch to the same viewed object; `protocol` copy/move still compile with the non-copyable bases.

Consequence: `protocol_view` is no longer trivially copyable, since a trivially copyable type must trivially copy every subobject.

### DIFF
--- a/reflection/protocol.hh
+++ b/reflection/protocol.hh
@@ -154,6 +154,16 @@ struct method_thunk<R (*)(Args...), EnclosingType, ProtocolType, Vtable, Member,
   static constexpr std::meta::info vtable_entry =
       find_vtable_member<^^Vtable, Member>();
 
+  // Copies detach the thunk from its enclosing object, so calling one is
+  // undefined behaviour; deleting the copy and move operations turns that
+  // misuse into a compile-time error.
+  method_thunk() = default;
+  method_thunk(const method_thunk&) = delete;
+  method_thunk(method_thunk&&) = delete;
+  method_thunk& operator=(const method_thunk&) = delete;
+  method_thunk& operator=(method_thunk&&) = delete;
+  ~method_thunk() = default;
+
   // Provides member-function call syntax. Recovers the EnclosingType pointer
   // through the vanishing-this-pointer cast, widens it to the enclosing
   // protocol/protocol_view object, then calls through its stored vtable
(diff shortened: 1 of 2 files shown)


jbcoe:
What are the consequences of "protocol_view is no longer trivially copyable, since a trivially copyable type must trivially copy every subobject"?

philipcraig:
> What are the consequences of "protocol_view is no longer trivially copyable, since a trivially copyable type must trivially copy every subobject"?

nothing in paper rules it out but it's a visible difference from span-like types

jbcoe:
The Itanium ABI will pass a class with a non-trivial copy constructor in memory, not in registers (like span and string view). I think this might be a deciding factor.

https://itanium-cxx-abi.github.io/cxx-abi/abi.html "non-trivial for the purposes of calls"

philipcraig:
> The Itanium ABI will pass a class with a non-trivial copy constructor in memory, not in registers (like span and string view). I think this might be a deciding factor.
> 
> https://itanium-cxx-abi.github.io/cxx-abi/abi.html "non-trivial for the purposes of calls"

For the avoidance of doubt, deciding factor *against* this approach? I listed the options I could see in #202. None are perfect except option 3 which we can't currently do.

jbcoe:
Thanks for clarifying. I think that making thunks non-copyable is not viable (this PR, option 1 from #202 ). 

I've no suggested way forward for now.

philipcraig:
Closing as, from discussion above, non-viable

PR #215, opened 2026-08-29
Make synthesised member thunks non-detachable while keeping protocol_view trivially copyable

### LINKED ISSUE #202: Prevent callable members from being copied.

philipcraig:
Implements @RyanJK5's suggestion from #202, replacing #212.

A copied or standalone `method_thunk` is undefined behaviour to call. The thunk's six special member functions are now `= default` but private, with the enclosing generated base as a friend, so `auto f = view.foo;` and `decltype(view.foo) f{};` are ill-formed while `protocol_view` keeps its defaulted special member functions and remains trivially copyable.

Closes #202.

### DIFF
--- a/reflection/protocol.hh
+++ b/reflection/protocol.hh
@@ -177,6 +177,15 @@ struct method_thunk<R (*)(Args...), EnclosingType, ProtocolType, Vtable, Member,
     return vtable->[:entry:](protocol_object->object_,
                              std::forward<Args>(args)...);
   }
+
+ private:
+  friend EnclosingType;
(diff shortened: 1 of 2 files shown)


jbcoe (changes requested):
Thanks for this.

Can the PR comment and code comment be condensed a bit? They are both rather verbose.

jbcoe (approved):
Thanks for this.

PR #216, opened 2026-08-29
Add vtable tutorial with name mangling for overloads and call operators

jbcoe:


philipcraig (commented):
Nice tutorial. A few places where the mangler produces colliding or unusable vtable names for inputs it claims to handle, plus some smaller things, inline.

tutorials/advanced_vtables.cc
```diff
@@ -0,0 +1,468 @@
...
+
+// Distinguishes overloads by folding each parameter's type and the
+// function's const-qualification into the name.
+consteval std::string_view mangle(std::meta::info fn) {
+  std::string name = "fn_" + mangle_atom(base_name_of(fn));
+  auto params = parameters_of(fn);
+  for (auto param : params) {
+    name += mangle_type(type_of(param));
+  }
+  if (is_const(fn)) name += "K";
```
philipcraig: `mangle` folds parameter types and `const` only, so ref-qualified and volatile-qualified overloads collide: `int f() &; int f() &&;` both give `fn_1f`, and `g() const` / `g() const volatile` both give `fn_1gK`, so `define_aggregate` fails with "name 'fn_1f' used in multiple data member descriptions". Folding the Itanium `R`/`O` ref-qualifier and `V` here would match what `same_signature_ignoring_const` in `protocol.hh` already distinguishes.

tutorials/advanced_vtables.cc
```diff
@@ -0,0 +1,468 @@
...
+// unambiguous because every atom is either a genuine C++ identifier, which
+// never starts with a digit, or one of this file's own non-digit tags, so
+// a decoder always knows where the length ends and the payload begins.
+consteval std::string mangle_atom(std::string_view text) {
+  return decimal(text.size()) + std::string(text);
+}
+
+// Length-prefixes each `::`-separated component of a qualified name
+// individually, so e.g. `A::B` can't collide with a hypothetical `A_B`.
+consteval std::string mangle_qualified_name(std::string_view name) {
```
philipcraig: Without nested-name bracketing (Itanium `N…E`), a single qualified type is indistinguishable from a sequence of unqualified ones: with `struct A { struct B {}; }; struct B {};`, `h(A::B)` and `h(A, B)` both mangle to `fn_1h1A1B` (likewise `h(A::B*)` vs `h(A*, B)` → `P1A1B`), so the comment above about `A::B` not colliding doesn't hold. Wrapping multi-component names in `N…E`, or walking `parent_of` structurally, fixes it.

tutorials/advanced_vtables.cc
```diff
@@ -0,0 +1,468 @@
...
+  if (is_const_type(type)) return "K" + mangle_type(remove_const(type));
+  if (is_pointer_type(type)) return "P" + mangle_type(remove_pointer(type));
+  if (is_lvalue_reference_type(type)) {
+    return "R" + mangle_type(remove_reference(type));
+  }
+  if (is_rvalue_reference_type(type)) {
+    return "O" + mangle_type(remove_reference(type));
+  }
+  if (auto builtin = builtin_type_code(type)) return std::string(*builtin);
+  return mangle_qualified_name(std::meta::display_string_of(type));
```
philipcraig: This splits `display_string_of` on every `::`, including inside template argument lists, and ties the golden at line 370 to GCC's current rendering of `std::string_view`. E.g. `std::vector<std::string>` becomes `3std10vector_std7__cxx1120basic_string_char___` (the ABI namespace leaks in, and `vector<std` isn't a scope component), and `Outer<A::B>::Inner` / `Outer_A::B_::Inner` both yield `7Outer_A2B_5Inner`. Walking `parent_of` / `identifier_of` / `template_arguments_of` removes both the split problem and the dependence on the implementation-defined string.

tutorials/advanced_vtables.cc
```diff
@@ -0,0 +1,468 @@
...
+
+  std::string_view noise(int) const { return "Purr"; }
+
+  std::string_view operator()() const { return "Meow"; }
+
+  std::string_view operator()(int) const { return "Purr"; }
+};
+
+template <typename R, typename... Args>
+using const_fn_ptr_t = R (*)(const void*, Args...);
```
philipcraig, line 444: Every generated slot takes `const void*`, but `member_functions_of` and `mangle` both admit non-const members (the name only gains `K` when const). For `struct NonConst { int set(int); }` the slot is `int (*fn_3seti)(const void*, int)` and the natural thunk fails with "passing 'const NonConst' as 'this' argument discards qualifiers", leaving `const_cast` as the only way through. `generate_vtable_specs` in `protocol.hh` picks `void*` / `const void*` from `is_const(fn)`; the earlier `Cat` in this file declares exactly such non-const `operator()` overloads.

tutorials/advanced_vtables.cc
```diff
@@ -0,0 +1,468 @@
...
+  if (is_const(fn)) name += "K";
+  return std::define_static_string(name);
+}
+
+// `Type`'s public, non-special member functions, in declaration order.
+template <std::meta::info Type>
+constexpr auto member_functions_of = std::define_static_array(
+    members_of(Type, std::meta::access_context::unprivileged()) |
+    filter(is_function) | filter(std::not_fn(is_static_member)) |
+    filter(std::not_fn(is_special_member_function)));
```
philipcraig: This range drops the `has_identifier || is_call_operator` guard that `conformance_candidate_infos` in `protocol.hh` and `filter(has_identifier)` in `reflection.cc` apply, so a user-declared constructor or a conversion function (`operator int() const`) reaches `identifier_of` and aborts constant evaluation with `std::meta::exception: reflection with has_identifier false`; `operator==` reaches `operator_code`'s default throw. Filtering `has_identifier(m) || is_operator_function(m)` and excluding constructors keeps the deliberate "for brevity" throws as the only failure mode.

tutorials/advanced_vtables.cc
```diff
@@ -0,0 +1,468 @@
...
+#include <charconv>
+#include <cstddef>
+#include <limits>
+#include <meta>
+#include <optional>
+#include <ranges>
+#include <stdexcept>
+#include <string>
+#include <string_view>
+#include <utility>
```
philipcraig: `<functional>` is missing: `std::not_fn` at line 338 only compiles because `<gtest/gtest.h>` pulls it in transitively (`protocol.hh` includes it explicitly for the same call), and the IWYU job never builds reflection targets so it won't catch this. `<utility>` looks unused.

tutorials/advanced_vtables.cc
```diff
@@ -0,0 +1,468 @@
...
+constexpr auto member_functions_of = std::define_static_array(
+    members_of(Type, std::meta::access_context::unprivileged()) |
+    filter(is_function) | filter(std::not_fn(is_static_member)) |
+    filter(std::not_fn(is_special_member_function)));
+
+TEST(TutorialsVtables, NameMangling) {
+  constexpr auto m_fns = member_functions_of<^^Cat>;
+
+  // operator()() const
+  XYZ_CONSTEVAL_CHECK(mangle(m_fns[0]) == "fn_2clK");
```
philipcraig: These `XYZ_CONSTEVAL_CHECK`s sit in a runtime `TEST` body, so only `mangle` runs at compile time and a wrong golden is a runtime failure rather than a build error, which is the opposite of what the surrounding prose implies. `reflection.cc` and `consteval_check_test.cc` wrap the same kind of checks in a `consteval { ... }` block.

tutorials/advanced_vtables.cc
```diff
@@ -0,0 +1,468 @@
...
+class AnimalPtr {
+  using vtable_t = vtable_generator<Cat>::vtable;
+  const void* data_;
+  const vtable_t* vtable_;
+
+ public:
+  template <typename T>
+  explicit AnimalPtr(const T* animal) : data_(animal) {
+    constexpr static vtable_t vtable_for_type{
+        .fn_5noiseK =
```
philipcraig: These designated initializers (and the accessors below) hand-copy the mangler's output with nothing tying them to `mangle(member_functions_of<^^Cat>[i])`, and the goldens in `NameMangling` test a different `Cat` with a different member order. Any change to `mangle` (e.g. the fixes above) breaks this section separately with "'struct vtable' has no member named 'fn_5noiseK'". A `static_assert(mangle(member_functions_of<^^Cat>[0]) == "fn_5noiseK")` beside them, or generating the thunks by splicing `nonstatic_data_members_of(^^vtable_t)` the way `make_view_vtable` does, would remove the drift.

tutorials/advanced_vtables.cc
jbcoe: The mangling is good enough for a tutorial.

philipcraig (approved):

tutorials/advanced_vtables.cc
jbcoe: I'm happy with this being brittle for brevity's sake.

jbcoe: We can tolerate the brittle nature of this example code for brevity and simplicity.

PR #220, opened 2026-08-29
Support overloaded member functions in the reflection implementation

### LINKED ISSUE #209: Support overloaded member functions in the reflection implementation

philipcraig:
Closes #209.

- Interface member functions are grouped by name; each name gets one `member_thunk` deriving from a `method_thunk` per overload, with `using ...::operator()...` forming the overload set.
- Vtable entries are addressed by index (named `<identifier>_<index>`), so overloads get distinct slots.
- `protocol_view<I>` (shallow const): for a pair `R f() const; R f();` the const overload is dropped so `f()` resolves to `R f()`, as through `I&`. `protocol_view<const I>` keeps only the const overloads; `protocol<I>` propagates const as before.

### DIFF
--- a/reflection/protocol_test.cc
+++ b/reflection/protocol_test.cc
@@ -1053,10 +1053,32 @@ TEST(ReflectionProtocolViewTest, MemberThunksCannotBeDetachedForOverloads) {
   static_assert(!std::is_default_constructible_v<decltype(p.compute)>);
   static_assert(!std::is_destructible_v<decltype(p.compute)>);
   static_assert(std::is_trivially_copyable_v<decltype(p.compute)>);
+}
 
+TEST(ReflectionProtocolViewTest, OverloadsThroughThunkReference) {
+  struct Interface {
+    int compute(int x);
+    double compute(double x);
+    std::string compute(const std::string& x) const;
+  };
+
+  struct Conforming {
+    int compute(int x) { return x * 2; }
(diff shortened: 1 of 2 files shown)


reflection/protocol_test.cc
```diff
@@ -1176,4 +1394,129 @@ TEST(ReflectionProtocolTest, ConstMemberFunctionOnConstProtocol) {
...
+  static_assert(!std::is_move_constructible_v<decltype(p.compute)>);
+  static_assert(!std::is_copy_assignable_v<decltype(p.compute)>);
+  static_assert(!std::is_move_assignable_v<decltype(p.compute)>);
+  static_assert(!std::is_default_constructible_v<decltype(p.compute)>);
+  static_assert(!std::is_destructible_v<decltype(p.compute)>);
+  static_assert(std::is_trivially_copyable_v<decltype(p.compute)>);
+
+  // Const propagates through `protocol`: the non-const overloads need a
+  // non-const reference to the thunk.
+  auto& compute = p.compute;
```
jbcoe, line 1537: I think that this block is misplaced. It belongs in another test as it's not associated with member thunk detachment.

jbcoe (changes requested):
Looks great. One small change needed to tests.

jbcoe (approved):
Thanks for this.

PR #221, opened 2026-08-29
Support call operators in the reflection implementation

### LINKED ISSUE #210: Support call operators in the reflection implementation

philipcraig:
Closes #210.

- `protocol_interface_functions_of` now includes `operator()`; `same_name` treats all call operators as one name, so conformance and overload grouping work as for named members. Lambdas conform (`protocol<Function> p([](int x) { ... })`, `protocol_view<Function>` as `function_ref`).
- Call operators get a `call_operator_base` wrapper base deriving from the `member_thunk` overload set, so `p(args)` works directly; protected special members keep it copyable by `protocol`/`protocol_view` but not sliceable.
- `method_thunk::enclosing` now uses `static_cast` when `EnclosingType` derives from the thunk, keeping the vanishing-this `reinterpret_cast` only for the `member_base` data-member case.
- Vtable entries for call operators are named `call_operator_<index>`.
- Const propagation, shallow const and `protocol_view<const I>` behave as for named overloads.

### DIFF
--- a/reflection/protocol.hh
+++ b/reflection/protocol.hh
@@ -74,6 +74,16 @@ inline constexpr bool is_protocol_view_v = is_protocol_view<T>::value;
 
 namespace detail {
 
+// Per ISO C++ ([expr.prim.lambda.closure]), closure types are unique, unnamed,
+// non-union class types.
+// This concept will also match an unnamed class type with a single
+// `operator()`.
+// TODO(jbcoe): Refine this concept to match only lambdas.
+template <typename T>
+concept is_maybe_lambda =
+    is_class_type(dealias(^^T)) && !has_identifier(dealias(^^T)) &&
+    requires { &T::operator(); };
+
 consteval bool is_call_operator(std::meta::info function) {
   return is_operator_function(function) &&
          operator_of(function) == std::meta::operators::op_parentheses;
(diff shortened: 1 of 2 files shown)


jbcoe (approved):
Thanks for this.

reflection/protocol.hh
```diff
@@ -108,19 +136,36 @@ consteval bool member_function_conforms_to(std::meta::info candidate,
...
+consteval auto protocol_interface_function_infos() {
+  auto named = members_of(Type, std::meta::access_context::unprivileged()) |
+               std::views::filter(std::meta::is_function) |
+               std::views::filter(std::not_fn(std::meta::is_static_member)) |
+               std::views::filter([](std::meta::info member) consteval {
+                 return has_identifier(member) || is_call_operator(member);
+               });
+  std::vector<std::meta::info> result(std::ranges::begin(named),
+                                      std::ranges::end(named));
+  // Per [meta.reflection.member.queries], a closure type's function call
```
jbcoe, line 155: @philipcraig are you happy with this?

philipcraig: the workaround is ok. Should it specify that the workaround is needed for GCC 16 but not trunk?

jbcoe: 100% yes. As worded the comment will persist for too long.

PR #224, opened 2026-08-29
Add **/*.hh to workflow path filters

### LINKED ISSUE #223: Workflow path filters miss .hh headers
The `paths` filters in `cmake.yml`, `sanitizers.yml` and `iwyu.yml` (both triggers) and `clang-tidy.yml` (`push`) list `**/*.h` but not `**/*.hh`, so a change touching only `reflection/protocol.hh` runs no build, sanitizer, IWYU or clang-tidy job. #220 and #221 only ran them because they also change `protocol_test.cc`.

Add `**/*.hh` to each filter.

philipcraig:
The `paths` filters in `cmake.yml`, `sanitizers.yml`, `iwyu.yml` and `clang-tidy.yml` list `**/*.h` but not `**/*.hh`, so a change touching only `reflection/protocol.hh` ran no build, sanitizer, IWYU or clang-tidy job.

Fixes #223


jbcoe (approved):
Thanks for this!

PR #226, opened 2026-08-29
Support static member functions on conformant types in the reflection implementation

### LINKED ISSUE #211: Support static functions for conformant types (not the interface type) in the reflection implementation

philipcraig:
Closes #211.

- `conformance_candidates_of<Type>` lists the named and call-operator member functions of a type, static or not; `is_protocol_conformant` and `find_conforming_member` search it. `protocol_interface_functions_of` is the non-static subset and remains the interface list, so an interface's own statics are still ignored.
- `member_function_conforms_to` skips the const and ref-qualifier checks for a static candidate, which has no object parameter; name, return type, parameter types and the noexcept rule still apply.
- The vtable trampolines are unchanged: `object->static_member(args)` is already valid C++.

### DIFF
--- a/reflection/protocol.hh
+++ b/reflection/protocol.hh
@@ -27,7 +27,7 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // `protocol_view`'s with destroy/copy/move entries used for allocator-aware
 // ownership.
 //
-// Neither implementation currently supports operators.
+// Neither implementation currently supports operators other than operator().
 
 #include <algorithm>
 #include <cassert>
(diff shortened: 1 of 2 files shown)


reflection/protocol.hh
```diff
@@ -93,54 +93,71 @@ consteval std::string_view member_name(std::meta::info member) {
...
+// interface member function.
+template <std::meta::info Type>
+constexpr inline auto conformance_candidates_of = std::define_static_array(
+    members_of(Type, std::meta::access_context::unprivileged()) |
+    std::views::filter(std::meta::is_function) |
+    std::views::filter([](std::meta::info member) consteval {
+      return has_identifier(member) || is_call_operator(member);
+    }));
+
+// The non-static members of `conformance_candidates_of<Type>`: the member
```
jbcoe: This comment needs some amendment. I don't think that "an entry's index identifies its vtable slot", nor is this function the right place to be documenting vtable naming or behaviour.

jbcoe (changes requested):
Thanks for this!

Changes to comments needed but code changes look good.

jbcoe (changes requested):
The PR description (and subsequent commit message) would benefit from rewording to make it more concise.

jbcoe (approved):
Thanks for this.

PR #228, opened 2026-08-29
Add member function forwarding tests to the reflection implementation

### LINKED ISSUE #145: Add tests to the reflection implementation for member function forwarding

philipcraig:
Closes #145.

New `reflection/forwarding_test.cc`, each case for both `protocol` and `protocol_view`:

- Arguments: rvalue-reference parameters are moved from, move-only by-value parameters, lvalue references mutate the caller's object, const references are not copied, by-value parameters are copied once from an lvalue and not at all from an rvalue, a braced-init-list and a user-defined conversion are accepted as for a direct call, six mixed parameters.
- Return values: lvalue/const references alias the stored object, move-only returns, by-value returns are not copied, void.
- Exceptions propagate through the thunk and the object stays usable.
- Implementations see their own `this`.

### DIFF
--- a/reflection/forwarding_test.cc
+++ b/reflection/forwarding_test.cc
@@ -6,6 +6,7 @@
 #include <cstddef>
 #include <format>
 #include <memory>
+#include <numeric>
 #include <stdexcept>
 #include <string>
 #include <string_view>
@@ -47,48 +48,56 @@ struct CopyCounter {
 // Argument forwarding.
 // ---------------------------------------------------------------------------
 
+// A moved-from std::unique_ptr is specified to be null, unlike the
+// valid-but-unspecified state of a moved-from standard container.
 TEST(ReflectionProtocolTest, RvalueReferenceParameterIsMovedFrom) {
   struct Interface {
-    void take(std::string&&);
-    const std::string& taken() const;
+    void take(std::unique_ptr<int>&&);
+    int taken() const;
   };
 
   struct Conforming {
-    std::string taken_value;
+    int taken_value = 0;
 
-    void take(std::string&& value) { taken_value = std::move(value); }
+    void take(std::unique_ptr<int>&& pointer) {
+      std::unique_ptr<int> owned = std::move(pointer);
+      taken_value = *owned;
+    }
 
-    const std::string& taken() const { return taken_value; }
+    int taken() const { return taken_value; }
   };
 
   protocol<Interface> p(Conforming{});
 
-  std::string source(40, 'x');
+  std::unique_ptr<int> source = std::make_unique<int>(5);
   p.take(std::move(source));
 
-  EXPECT_TRUE(source.empty());
-  EXPECT_EQ(p.taken(), std::string(40, 'x'));
+  EXPECT_EQ(source, nullptr);
+  EXPECT_EQ(p.taken(), 5);
 }
 
 TEST(ReflectionProtocolViewTest, RvalueReferenceParameterIsMovedFrom) {
   struct Interface {
-    void take(std::string&&);
+    void take(std::unique_ptr<int>&&);
   };
 
   struct Conforming {
-    std::string taken;
+    int taken_value = 0;
 
-    void take(std::string&& value) { taken = std::move(value); }
+    void take(std::unique_ptr<int>&& pointer) {
+      std::unique_ptr<int> owned = std::move(pointer);
+      taken_value = *owned;
+    }
   };
 
   Conforming implementation;
   protocol_view<Interface> view(implementation);
 
-  std::string source(40, 'x');
+  std::unique_ptr<int> source = std::make_unique<int>(5);
   view.take(std::move(source));
 
-  EXPECT_TRUE(source.empty());
-  EXPECT_EQ(implementation.taken, std::string(40, 'x'));
+  EXPECT_EQ(source, nullptr);
+  EXPECT_EQ(implementation.taken_value, 5);
 }
 
 TEST(ReflectionProtocolTest, MoveOnlyByValueParameter) {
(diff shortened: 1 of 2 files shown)


reflection/forwarding_test.cc
```diff
@@ -0,0 +1,691 @@
...
+    void take(std::string&& value) { taken = std::move(value); }
+  };
+
+  Conforming implementation;
+  protocol_view<Interface> view(implementation);
+
+  std::string source(40, 'x');
+  view.take(std::move(source));
+
+  EXPECT_TRUE(source.empty());
```
jbcoe: I'm not sure that this is a safe check. A moved-from string with SSO could still hold the previous content. 
`std::vector` has no SSO so is guaranteed to be empty.

(https://eel.is/c++draft/string.cons#2)

```
constexpr basic_string(const basic_string& str);
constexpr basic_string(basic_string&& str) noexcept;
```

Effects: Constructs an object whose value is that of str prior to this call.
Remarks: In the second form, str is left in a *valid but unspecified state*.

reflection/forwarding_test.cc
```diff
@@ -0,0 +1,691 @@
...
+
+  struct Conforming {
+    int consume(CopyCounter counter) { return counter.value; }
+  };
+
+  Conforming implementation;
+  protocol_view<Interface> view(implementation);
+
+  CopyCounter::reset();
+  CopyCounter counter(9);
```
jbcoe, line 234: Q: Mostly curiousity, why `9`?

reflection/forwarding_test.cc
```diff
@@ -0,0 +1,691 @@
...
+
+  Conforming implementation;
+  protocol_view<Interface> view(implementation);
+
+  CopyCounter::reset();
+  view.consume(CopyCounter{1});
+  EXPECT_EQ(CopyCounter::copies, 0);
+}
+
+TEST(ReflectionProtocolTest, ImplicitConversionAtCallSite) {
```
jbcoe: Q: What is this testing? How can we change code that would invalidate this test?

reflection/forwarding_test.cc
```diff
@@ -0,0 +1,691 @@
...
+    std::size_t length(std::string_view text) const { return text.size(); }
+  };
+
+  protocol<Interface> p(Conforming{});
+
+  EXPECT_EQ(p.scale(3), 6.0);
+  EXPECT_EQ(p.length("four"), 4u);
+}
+
+TEST(ReflectionProtocolViewTest, ImplicitConversionAtCallSite) {
```
jbcoe: Q: Similalry, what is this testing? How can we change code that would invalidate this test?

reflection/forwarding_test.cc
```diff
@@ -0,0 +1,691 @@
...
+  };
+
+  Conforming implementation;
+  protocol_view<Interface> view(implementation);
+
+  EXPECT_EQ(view.describe(1, 2.5, 'a', "text", true, 7L),
+            "1,2.5,a,text,true,7");
+}
+
+TEST(ReflectionProtocolTest, ProtocolViewParameter) {
```
jbcoe: Q: What is this testing? How can we change code that would invalidate this test?

reflection/forwarding_test.cc
```diff
@@ -0,0 +1,691 @@
...
+
+    int apply(int value) const { return inner.apply(value) + 1; }
+  };
+
+  protocol<Unary> outer(Forwarder{protocol<Unary>(Doubler{})});
+
+  EXPECT_EQ(outer.apply(4), 9);
+}
+
+TEST(ReflectionProtocolViewTest, ViewOfImplementationHoldingAProtocol) {
```
jbcoe: Q: What is this testing? How can we change code that would invalidate this test?

jbcoe (approved):
Thanks for this.

Some of these tests feel a bit odd - I don't see that we can change our implementation in any way that would make them fail as they rely on C++ language properties.

None does any harm. The forwarding tests are very much needed!

philipcraig:
> Thanks for this.
> 
> Some of these tests feel a bit odd - I don't see that we can change our implementation in any way that would make them fail as they rely on C++ language properties.
> 
> None does any harm. The forwarding tests are very much needed!

I think you were right and this is addressed

reflection/forwarding_test.cc
philipcraig: Agreed — switched both tests to `std::unique_ptr<int>&&`, whose moved-from state is specified to be null ([unique.ptr.single.ctor]). `std::vector` gives the same "valid but unspecified" wording as `std::string` in [container.reqmts], so I avoided it too.

reflection/forwarding_test.cc
philipcraig: No significance. The test now asserts `view.consume(counter) == 9` so the value has a purpose; same for `1` in the rvalue tests.

reflection/forwarding_test.cc
philipcraig: It pins the generated wrapper's signature to the interface's declared parameter types (`method_thunk::operator()(Args...)`) rather than deducing them, so a call site converts exactly as a direct call to the interface would. `scale(3)` didn't demonstrate that, since a perfect-forwarding wrapper would also convert `int` to `double` at the trampoline. Replaced with `ArgumentsConvertAsForADirectCall`: `p.sum({1, 2, 3})` against `int sum(const std::vector<int>&)`, which a `template <typename... Ts> operator()(Ts&&...)` wrapper cannot deduce, plus `const char*` → `std::string_view` as a user-defined conversion.

reflection/forwarding_test.cc
philipcraig: Same as above; replaced with the braced-init-list check.

reflection/forwarding_test.cc
philipcraig: Only that a `protocol_view` can be passed by value through the vtable, which reduces to `protocol_view` being copyable — already covered by the static_asserts in `protocol_test.cc`. No change to the call path would fail this test alone, so dropped.

reflection/forwarding_test.cc
philipcraig: Nothing specific: an implementation holding a `protocol<Unary>` just makes a second, independent vtable call, so this only re-exercised the plain call path. Dropped it and its `protocol` twin.

PR #232, opened 2026-08-29
Grant the codespace actions: write so failed CI jobs can be re-run with gh

philipcraig:
`gh run rerun --failed` from a codespace fails with "Resource not accessible by integration": the Codespaces `GITHUB_TOKEN` has no `actions: write`. Declaring the permission in `devcontainer.json` makes GitHub request it when the codespace is created, so failed jobs (e.g. the transient apt failure on #228) can be re-run from the terminal.


jbcoe (approved):
Thanks for this!

PR #235, opened 2026-08-30
Define a separate default build-dir for clang-tidy

### LINKED ISSUE #231: `scripts/cmake.sh --clang-tidy` fails and reconfigures `build/` when the configured compiler is GCC trunk
The manual `clang-tidy` pre-commit hook runs `scripts/cmake.sh --debug
--clang-tidy`, which reuses `build/`. If `build/` was configured with
`/opt/gcc-latest/bin/g++` (the `--reflection` setup), two things go wrong:

- GCC trunk defaults to C++20, so CMake emits no `-std=` for `cxx_std_20`
  targets; `clang-tidy-22` then parses `protocol.cc` as gnu++17 and reports
  errors such as "defaulted comparison operators are a C++20 extension" and
  "no template named 'remove_cvref_t'".
- The hook reconfigures `build/` to the Debug preset with `CLANG_TIDY_ENABLE`,
  so the next `--reflection` build needs `--clean`.

CI is unaffected (it uses `clang++-22`, which gets an explicit `-std`).
Suggested fix: when `--clang-tidy` is given, default `-B` to
`build-clang-tidy` and `CMAKE_CXX_COMPILER` to `clang++` unless `CXX` is set,
so the hook matches CI and never touches `build/`.

jbcoe:
Closes #231

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -16,6 +16,7 @@
       },
       {
         "name": "default",
+        "binaryDir": "${sourceDir}/build",
         "cacheVariables": {
           "CMAKE_INSTALL_PREFIX": "${sourceDir}/install/${presetName}",
           "CMAKE_INTERPROCEDURAL_OPTIMIZATION": true
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -16,7 +16,6 @@
       },
       {
         "name": "default",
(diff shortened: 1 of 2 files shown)


philipcraig (approved):
tested locally and with worktrees, all good

Issue #239, opened 2026-08-30
Remove the --reflection build option


philipcraig:
With the reflection implementation as the only implementation (#214), the opt-in to build it is redundant. Remove:

- `--reflection` from `scripts/cmake.py`; the GCC trunk / gcc-16 compiler selection it gates should become the default when `CXX` is unset
- `XYZ_PROTOCOL_BUILD_REFLECTION` and the `XYZ_PROTOCOL_REFLECTION_SUPPORTED` check in `CMakeLists.txt`; make the P2996 check unconditional and fail configuration if the compiler lacks reflection
- the `if(XYZ_PROTOCOL_BUILD_REFLECTION)` guards in `CMakeLists.txt`, `tutorials/CMakeLists.txt` and `reflection/CMakeLists.txt`
- `settings.reflection` and the `--reflection` argument in `.github/workflows/cmake.yml`
- `--reflection` from the build instructions in `CONTRIBUTING.md`, `GEMINI.md`, `CLAUDE.md` and `Agents.md`

`-freflection` itself is still required on GCC trunk; it should be applied unconditionally rather than per-target.

Related: #214.

Issue #237, opened 2026-08-30
Drop all CI builds except GCC 16 and GCC trunk


philipcraig:
The reflection implementation is becoming the sole implementation (#214) and only builds on GCC with P2996 support. Once the Python-generated implementation is gone, the other compilers in `.github/workflows/cmake.yml` have nothing to build.

Remove from the `cmake.yml` matrix:

- Ubuntu GCC-11, GCC-12, GCC-13, GCC-14, GCC-14 (C++23)
- Ubuntu Clang-17 through Clang-22 (+ libc++, C++20 and C++23 variants)
- Visual Studio 2019, 2022 v17/v18/v19
- MacOS Apple Clang 15

Keep:

- Ubuntu GCC trunk (C++26 Reflection) — gating
- Ubuntu GCC-16 (C++26 Reflection) — informational, as today

Also decide what happens to the Clang-based workflows, which cannot compile the reflection code: `sanitizers.yml` (Clang 22; GCC trunk can run ASAN/UBSAN/TSAN but not MSAN), `clang-tidy.yml` (a required check, see #213), and `iwyu.yml`.

Related: #214, #122.

PR #240, opened 2026-08-30
Remove the Python code-generation implementation

### LINKED ISSUE #238: Remove the Python code-generation implementation
Once the reflection implementation is the sole implementation (#214), remove the Python backend and everything that exists only to support it:

- `scripts/generate_protocol.py`, `scripts/protocol.j2`, `scripts/regenerate_reference_interface_protocol.py`, `scripts/test_generate_protocol.py`, `scripts/test_concept_errors.py`, `scripts/run_concept_errors_test.sh`
- `cmake/xyz_generate_protocol.cmake` and the `xyz_generate_protocol(...)` calls, `generate_protocols` target, and Python3/jinja2/clang/xyz.cppmodel discovery in `CMakeLists.txt`
- generated-implementation sources and inputs: `protocol.h`, `protocol.cc`, `protocol_test.cc`, `protocol_benchmark.cc`, `interface_*.h`, `reference_interface.h`, `reference_interface_protocol.h`
- `.github/workflows/pytest.yml`
- `jinja2`, `clang`, `pyyaml` and `xyz-cppmodel` from `pyproject.toml` / `uv.lock` (keep `uv` and the tooling deps used by `scripts/cmake.py`, pre-commit, clang-format, cmake-format)
- the code-generation sections of `README.md`, `CONTRIBUTING.md`, `implementation-notes.md`, `GEMINI.md` and `CLAUDE.md`

Move `reflection/` contents to the repository root so `reflection/protocol.hh` becomes the implementation, and drop the `reflection_` prefix from target names.

Related: #214.

philipcraig:
Closes #238.

Removes the Python/Jinja2 code-generation implementation and everything that existed only to support it, and makes `protocol.hh` (moved from `reflection/`) the implementation. Target names drop the `reflection_` prefix: `protocol`, `xyz_protocol::protocol`, `protocol_test`.

Beyond the issue's list:

- The `protocol_benchmark` target, `run_benchmark`, the google-benchmark fetch and the `benchmark` mode of `scripts/cmake.py` are removed with `protocol_benchmark.cc`.
- `pytest` is dropped from `pyproject.toml` with `pytest.yml`; `pyrefly` and its config stay.
- `reflection/tagged_allocator.h` and `reflection/tracking_allocator.h` were identical to the root copies and are dropped rather than moved.
- `implementation-notes.md` is rewritten for the reflection implementation.

Left as is: the `xyz::reflection` namespace and include guard, the `[py_cppmodel]` reference in `DRAFT.md`, `test_helpers.h`, and the `--reflection` option (#239) and CI matrix (#237).


implementation-notes.md
jbcoe, line 1: This is too detail-heavy. Lots of this is irrelevant to a reader unless they look into code in which case it is clearly documented in code. Low level detail in high-level docs will result in churn/confusion.

jbcoe (changes requested):
Thanks for this.

`implementation-notes.md` needs some human authorship (remove bits), otherwise we're good to go.

implementation-notes.md
philipcraig: Removed in 78d7b2e rather than trimmed; CONTRIBUTING.md is now the only prose about the implementation and the README points there.

jbcoe (approved):
Thanks for this!

PR #241, opened 2026-08-30
Remove the --reflection build option

### LINKED ISSUE #239: Remove the --reflection build option

philipcraig:
Closes #239.

The reflection implementation is the only one, so the opt-in goes away


jbcoe (changes requested):
Please amend the PR description.

jbcoe (approved):
Thanks for this.

PR #242, opened 2026-08-30
Drop all CI builds except GCC 16 and GCC trunk

### LINKED ISSUE #237: Drop all CI builds except GCC 16 and GCC trunk

philipcraig:
Closes #237.

Builds on #240; this branch carries its commit and should land after it. #241 then rebases onto this.

Clang-based workflows:

- `sanitizers.yml` moves to GCC trunk with `asan`+`ubsan` and `tsan`. GCC has no MemorySanitizer, so the `msan` job goes, and with it the `--msan`/`ENABLE_MSAN` options that would otherwise silently build unsanitized.
- `clang-tidy.yml` and `iwyu.yml` are deleted, along with `FindClangTidy.cmake`, `FindIWYU.cmake`, the IWYU mapping file, both `.clang-tidy` configs, the `--clang-tidy` option in `scripts/cmake.py` and the manual pre-commit hook. clang-tidy 22 cannot parse the reflection sources.


jbcoe (approved):
Thanks for this!

jbcoe (changes requested):
Please simplify the PR description

jbcoe (approved):

PR #244, opened 2026-08-30
Let a protocol be viewed through protocol_view

### LINKED ISSUE #229: `protocol<I>` cannot be viewed through `protocol_view<I>`
`is_protocol_conformant<I, protocol<I>>()` is `false`: the synthesised members
of `protocol<I>` are data members holding `member_thunk` objects with an
`operator()`, not member functions, so `conformance_candidates_of` never sees
them. As a result

    protocol<I> p(Impl{});
    protocol_view<I> view(p);   // ill-formed

does not compile, and a `protocol<I>` cannot be passed where an
implementation of `I` is expected (e.g. an interface member taking
`protocol_view<I>`, or as the stored object of an outer `protocol<I>`).

The code-generation implementation supports `protocol_view<I>` from
`protocol<I>`. Options:

1. Extend the conformance check to accept a public data member whose type is
   callable with the interface signature (a `member_thunk` overload set).
2. Special-case `protocol<I>` / `protocol_view<I>` in `protocol_view<I>`'s
   constructor and in the conformance check.
3. Document the limitation and leave it.

Found while writing the forwarding tests for #145 (#228).

philipcraig:
Closes #229.

`protocol_view<I>` and `protocol_view<const I>` can view the object a `protocol<I>` owns, sharing its vtable; `protocol_view<const I>` can also be constructed from a `protocol_view<I>`.


jbcoe:
This needs to be rebased onto #256 .

jbcoe (changes requested):
This needs rebasing onto https://github.com/jbcoe/cc-protocol/pull/256

I shall do the rebasing.

philipcraig:
> This needs rebasing onto #256
> 
> I shall do the rebasing.

I had a go at the rebase

philipcraig:
If you'd rather have strict noexcept matching for protocol candidates (exact-name-only lookup), the `mangle(function, as_noexcept)` overload and two-name lookup can be deleted and the one assertion in `NoexceptInterfaceRequiresNoexceptProtocol` flipped — an isolated ~15-line simplification.

protocol.hh
```diff
@@ -204,6 +207,27 @@ consteval std::meta::info find_vtable_entry() {
...
 }
 
+// The entry of `VtableType` named by the mangled signature of the interface
+// member function `member`, if it declares one. A `member` that may throw
+// also matches an entry mangled as noexcept: a noexcept entry can stand in
+// for one that is allowed to throw, mirroring `member_function_conforms_to`.
+template <std::meta::info VtableType>
+consteval std::optional<std::meta::info> find_vtable_entry(
+    std::meta::info member) {
+  std::string name = xyz::name_mangling::mangle(member);
```
jbcoe: Can this be cached? We will be calculating the mangling a lot of times as written.

Use of `mangled_name_of` allows caching of values.

protocol.hh
```diff
@@ -570,8 +620,39 @@ struct const_view_trampoline<R (*)(const void*, Args...) noexcept(Noexcept), U,
...
+  static R call(void* ptr, Args... args) noexcept(Noexcept) {
+    auto* stored = static_cast<Protocol*>(ptr);
+    return stored->vtable_->[:Entry:](stored->object_,
+                                      std::forward<Args>(args)...);
+  }
+};
+
+template <typename R, typename... Args, bool Noexcept, typename Protocol,
+          std::meta::info Entry>
+struct protocol_trampoline<R (*)(const void*, Args...) noexcept(Noexcept),
```
jbcoe: This will do a multiple-lookup through multiple vtables. I don't know how we can do better (yet) but we don't want to do this as it can chain in an unbounded way.

protocol.hh
```diff
@@ -589,19 +670,28 @@ consteval typename vtable_generator<T>::vtable make_view_vtable() {
...
+    if constexpr (is_const(member) || ConstPolicy != const_policy::const_only) {
+      constexpr std::meta::info vtable_member =
+          find_vtable_entry<^^Vtable, member>();
+      using FnPtrType = typename[:type_of(vtable_member):];
+      if constexpr (is_protocol_v<U>) {
+        using StoredVtable =
+            typename vtable_generator<protocol_interface_t<U>>::vtable;
+        constexpr std::meta::info entry =
+            *find_vtable_entry<^^StoredVtable>(member);
+        result.[:vtable_member:] = &protocol_trampoline<FnPtrType, U,
```
jbcoe: This will do a multiple-lookup through multiple vtables. I don't know how we can do better (yet) but we don't want to do this as it can chain in an unbounded way.

protocol.hh
```diff
@@ -1036,6 +1158,19 @@ class protocol_view<const T> : public detail::protocol_wrappers_t<
...
+  // moved from, as the owned object moves with it.
+  //
+  // Precondition: `p` is not valueless.
+  template <typename Alloc>
+  explicit protocol_view(const protocol<T, Alloc>& p) noexcept
+      : object_(p.object_), vtable_(p.vtable_) {}
+
+  // A view of a temporary would dangle.
+  template <typename Alloc>
+  protocol_view(const protocol<T, Alloc>&&) = delete;
```
jbcoe, line 1093: This is very much needed.

protocol.hh
```diff
@@ -1036,6 +1158,19 @@ class protocol_view<const T> : public detail::protocol_wrappers_t<
         vtable_(
             &detail::view_vtable_for<T, U, detail::const_policy::const_only>) {}
 
+  // Views the object a `protocol<T>` owns, sharing its vtable. The view is
+  // invalidated when `p` is destroyed or assigned to, but not when it is
+  // moved from, as the owned object moves with it.
+  //
+  // Precondition: `p` is not valueless.
+  template <typename Alloc>
+  explicit protocol_view(const protocol<T, Alloc>& p) noexcept
```
jbcoe, line 1088: This is good and does not need changes to conformance to allow construction.

protocol_test.cc
```diff
@@ -2166,4 +2167,368 @@ TEST(ReflectionProtocolTest, ConstValuelessCall) {
 
 #endif
 
+// ---------------------------------------------------------------------------
+// Protocols as conforming types: a `protocol<J>` conforms to `I` when its
```
jbcoe: I think that this is beyond scope.

We want construction of:

`protocol_view<T>` from `protocol<T>`
`protocol_view<const T>` from `protocol<T>`
`protocol_view<const T>` from `protocol_view<T>`

jbcoe (changes requested):
We must avoid the multiple jumping through chained vtables.

jbcoe:
> If you'd rather have strict noexcept matching for protocol candidates (exact-name-only lookup), the `mangle(function, as_noexcept)` overload and two-name lookup can be deleted and the one assertion in `NoexceptInterfaceRequiresNoexceptProtocol` flipped — an isolated ~15-line simplification.

Construction of

`protocol_view<T>` from `protocol<T>`
`protocol_view<const T>` from `protocol<T>`
`protocol_view<const T>` from `protocol_view<T>`

Does not allow for differing noexcept specifications.

I think that the conformance question can be wholly sidestepped as we require `T` to be the same.

`protocol_view<T>` is not constructible from `protocol<U>` unless `U` and `T` are the same type.

jbcoe:
I think that this issue is misnamed - we don't need conformance or vtable naming alignment to check for conformance, just new constructors.

protocol_test.cc
```diff
@@ -2164,4 +2165,329 @@ TEST(ReflectionProtocolTest, ConstValuelessCall) {
...
+// vtable has an entry for every member function of `I`.
+// ---------------------------------------------------------------------------
+
+TEST(ConformsToTest, ProtocolConformsToItsOwnInterface) {
+  struct Interface {
+    int get() const;
+    void update(int value);
+  };
+
+  static_assert(is_protocol_conformant<Interface, protocol<Interface>>());
```
jbcoe: I'm not sure that we want or need this.

Might be worth a quick chat?

jbcoe (changes requested):

protocol.hh
philipcraig: Gone in 4152857: the conformance changes were removed (see the scope thread below), and this overload with them; entry lookup is back to the cached `mangled_name_of` path only.

protocol.hh
philipcraig: Removed in 4152857: a protocol can no longer be stored in or viewed through another protocol. Only same-interface views remain, and they share the protocol's vtable.

protocol.hh
philipcraig: Removed in 4152857 along with `protocol_trampoline`; `make_view_vtable` is back to main's form.

protocol_test.cc
philipcraig: Scoped down in 4152857: the conformance changes are gone and the PR now adds exactly these three constructions. `protocol_view<const T>` from `protocol_view<T>` is new — implicit and by value, per DRAFT.md.

protocol_test.cc
philipcraig: Removed in 4152857 with the rest of the conformance changes; `is_protocol_conformant` is untouched by this PR now. Happy to chat if you want to revisit conformance for protocols separately.

protocol.hh
```diff
@@ -1013,7 +1018,25 @@ class protocol_view
         vtable_(
             &detail::view_vtable_for<T, U, detail::const_policy::all_const>) {}
 
+  // Views the object a `protocol<T>` owns, sharing its vtable. The view is
+  // invalidated when `p` is destroyed or assigned to, but not when it is
```
jbcoe: The invalidation comment here feels out of place. It’s true but maybe not needed.

philipcraig: Trimmed here and on the protocol_view<const T> counterpart.

protocol_test.cc
```diff
@@ -2204,4 +2204,210 @@ TEST(ReflectionProtocolTest, ConstValuelessCall) {
...
+  protocol<Interface> p(Conforming{});
+  protocol_view<Interface> view(p);
+  protocol_view<const Interface> const_view(view);
+
+  view.update(7);
+  EXPECT_EQ(const_view.get(), 7);
+  static_assert(!has_update<protocol_view<const Interface>>);
+}
+
+TEST(ReflectionProtocolViewTest, ViewOfProtocolPassedToInterfaceMember) {
```
jbcoe: I think this test is more complicated than it needs to be. What do we gain by having sink as a protocol?

jbcoe (changes requested):
This is in great shape. Some tests feel a little convoluted and unfocused. I'm happy to fix them up tonight.

jbcoe (approved):
Thanks for this. 

There was significant scope change as the PR was being drafted; the end result is awesome.

PR #247, opened 2026-08-30
Give the vscode devcontainer user UID 1000 and document credential sharing

philipcraig:
`ubuntu:24.04` and later ship a default `ubuntu` user at 1000:1000, so `useradd vscode` in the Dockerfile landed on UID 1001. On Linux and WSL hosts, directories bind-mounted from the host are owned by UID 1000, and VS Code's `updateRemoteUserUID` cannot remap `vscode` onto an ID that is already taken. Remove the stock user before creating `vscode` so it gets 1000:1000.

Docker Desktop on macOS and Codespaces are unaffected either way.

`devcontainer.json` gains a comment showing the optional bind mounts for reusing a host `gh auth login` and Claude Code login inside the container (`~/.config/gh`, `~/.claude`, `~/.claude.json`). They are not enabled by default because Docker refuses to start the container when a bind source is missing.

Also recommends the Claude Code VS Code extension, matching the `CLAUDE.md` in the repository.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -9,6 +9,11 @@
 		"type=bind,source=${localEnv:HOME}${localEnv:USERPROFILE}/.ssh,target=/home/vscode/.ssh,readonly"
 	],
 	"customizations": {
+		"vscode": {
+			"extensions": [
+				"anthropic.claude-code"
+			]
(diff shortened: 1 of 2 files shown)


jbcoe (approved):
Thanks for this!

Issue #213, opened 2026-08-29
Follow clang-tidy checks approach to allow checks to be required and skipped if not needed


jbcoe:
See: https://github.com/jbcoe/cc-protocol/pull/188

philipcraig:
wait, which checks do you want to block merging to main? all checks or some subset and if a subset, which subset?

jbcoe:
Let's upgrade the GCC-16 (latest) ones, IWYU, pre-commit. Others will be removed soon enough.

philipcraig:
> Let's upgrade the GCC-16 (latest) ones, IWYU, pre-commit. Others will be removed soon enough.

and code coverage

philipcraig:
The workflow side of this is done: #248 added the detect-changes skip to the CMake, sanitizer and uv-lock workflows, #271 added the fixed-name summary jobs (`cmake`, `bazel`, `sanitizers`) that report even when a matrix is skipped, and #293 moved the sanitizers to Bazel under the same name. Of the subset asked for here: `pre-commit` is already required, the GCC-16 builds are covered by `cmake` and `bazel`, IWYU was removed in #242, and code coverage stays advisory (decided on #257: the consteval coverage job takes about 15 minutes against under 4 for every other check). The only step left is the ruleset edit, which #249 tracks with the exact entries and the run-history evidence that they are safe. Closing in favour of #249.

PR #188, opened 2026-08-28
Make clang-tidy a required status check for merging to main

philipcraig:
Refs #185.

Workflow-side preparation for making the `clang-tidy` job a required status check on `main`.

- Drop the `paths:` filter from the `pull_request` trigger of `.github/workflows/clang-tidy.yml`. A workflow filtered out by `paths:` never reports a check, so once the check is required, PRs that touch no C++ (e.g. `DRAFT.md`-only changes) would sit blocked as "Expected — Waiting for status to be reported".
- Add a `detect-changes` job (`dorny/paths-filter@v3`, same glob list as the `push` trigger) that gates the `clang-tidy` job. A job skipped by an `if:` condition counts as a passing check. If change detection fails, the build runs rather than silently passing.
- `push` and `schedule` triggers are unchanged.
- `CONTRIBUTING.md` documents the required check.

**Merge this before enabling the required check** in the `main` ruleset (needs admin; instructions in #185), otherwise the next non-C++ PR is stuck.

Verification: `uv run pre-commit run --files .github/workflows/clang-tidy.yml CONTRIBUTING.md` passes (including `action-validator`). The `Clang Tidy` run on this PR exercises the new `detect-changes` → `clang-tidy` path.

https://claude.ai/code/session_01TGTHbSozmsGbTecSZkuPSt

PR #248, opened 2026-08-30
Prepare build, sanitizer and uv-lock jobs to be required status checks

philipcraig:
Refs #213.

Workflow-side preparation for making the build, sanitizer and lockfile jobs required status checks on `main`, following the approach from #188: the `pull_request` triggers lose their `paths:` filters so every pull request reports these checks, and a `detect-changes` job (`dorny/paths-filter`) skips the expensive jobs when nothing relevant changed. A job skipped by `if:` counts as a passing check. `uv-lock` is cheap enough to always run.

Job names are shortened so the ruleset contexts are short and stable. After this merges, add these contexts to the `main` ruleset under *Require status checks to pass* (needs admin):

- `GCC trunk Release`
- `GCC trunk Debug`
- `GCC-16 Release`
- `GCC-16 Debug`
- `asan`
- `tsan`
- `uv-lock`
- `pre-commit` (already required)

**Enable the ruleset entries only after this merges**: with the current `paths:` filters a required check that never reports would block every non-C++ pull request.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -13,21 +13,55 @@ on:
       - "**/CMakeLists.txt"
       - "CMakePresets.json"
       - "**/*.cmake"
+      - "scripts/cmake.py"
+      - "scripts/cmake.sh"
+  # No `paths:` filter here. The `GCC trunk` build jobs are required status
+  # checks for merging to main, and a workflow that is filtered out by
+  # `paths:` never reports a check, which leaves unrelated pull requests
+  # blocked as "Expected - Waiting for status to be reported". Instead, the
+  # `detect-changes` job below decides whether the build needs to run, and a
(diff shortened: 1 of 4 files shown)


.github/workflows/cmake.yml
```diff
@@ -55,7 +89,7 @@ jobs:
           # above is the gating one. Drop this entry once the reflection backend
           # needs fixes that only exist on trunk.
           - {
-              name: "Ubuntu GCC-16 (C++26 Reflection, informational)",
+              name: "GCC-16 (informational)",
```
jbcoe: AHA! This is where "informational" comes from!

jbcoe (changes requested):
Thanks for this.

We can remove the informational label for GCC-16 and then we're good to go!

jbcoe (approved):

PR #250, opened 2026-08-30
Adds Bazel build and test targets (#243)

### LINKED ISSUE #243: Add support for building with Bazel
See https://bazel.build/

jbcoe:
Adds `.github/workflows/bazel.yml` to test GCC trunk and GCC-16 in CI.

Modifies `scripts/cmake.py` to confine CMake build directories to a single (bazel) ignored parent.
The default CMake build directory is nested under `build` and distinct per OS and host.

Follow-up work is needed to get Bazel builds working with homebrew install GCC 16 on MacOS.

closes #243


scripts/bazel.py
```diff
@@ -0,0 +1,191 @@
...
+    if system_gxx_path and system_gcc_path:
+        return system_gcc_path, system_gxx_path
+
+    # Fallback to standard gcc/g++ if version 16 is not explicitly named
+    default_gxx_path = shutil.which("g++") or "g++"
+    default_gcc_path = shutil.which("gcc") or "gcc"
+    return default_gcc_path, default_gxx_path
+
+
+def create_compiler_wrappers(real_cc_path: str, real_cxx_path: str) -> Tuple[str, str]:
```
philipcraig: The wrapper scripts make the real compiler invisible to Bazel's cache keys: `--repo_env` CC/CXX are the constant wrapper paths, and Bazel never digests the binary behind them. Switching `CXX=gcc-16` after a trunk build, or replacing the `/opt/gcc-latest` snapshot in place, rewrites the wrapper without invalidating anything, so stale objects are replayed. With bazel.yml's disk cache keyed only by workflow name, both matrix jobs share one cache — the nightly "GCC trunk" job can go green without compiling the new snapshot. Consider embedding the resolved compiler path/version in the cache key (e.g. via `--repo_env` of the real path, or a version stamp in the wrapper content plus `--action_env`), and splitting the CI cache per matrix job.

.bazelignore
```diff
@@ -0,0 +1,15 @@
+build
```
philipcraig, line 1: `.bazelignore` entries are literal paths — no globs — so `bazel-*` matches nothing and this list only covers the six named build dirs. Any other `-B` name (e.g. `./scripts/cmake.sh -B build-tsan`) leaves a `_deps/googletest-src/BUILD.bazel` that breaks `bazel build //...` analysis. Since `.gitignore` uses `build*/`, the free-form names are expected. One option: have scripts/cmake.py confine build dirs to a single ignored parent, or generate `.bazelignore` entries.

CONTRIBUTING.md
```diff
@@ -16,25 +16,28 @@ Python dependencies and execute build scripts.
...
 
 ```bash
-./scripts/cmake.sh --help
+./scripts/bazel.sh
+```
+
+or directly using Bazel:
+
+```bash
+bazel test //...
```
philipcraig: Documenting bare `bazel test //...` drops both scripts/bazel.py's compiler selection and CMake's configure-time reflection probe (CMakeLists.txt FATAL_ERROR with install guidance). On stock distro GCC this fails with `unrecognized command-line option '-freflection'` deep in sandboxed actions. Worth either removing the bare invocation or adding a capability check on the Bazel path.

scripts/bazel.py
```diff
@@ -0,0 +1,191 @@
...
+    }
+    target_mode = mode_map[parsed_arguments.mode]
+
+    def log_message(message_content: Any) -> None:
+        if parsed_arguments.verbose:
+            print(message_content)
+
+    bazel_environment = os.environ.copy()
+
+    if "CC" in bazel_environment and "CXX" in bazel_environment:
```
philipcraig: This requires both CC and CXX to be set; exporting only `CXX` is silently ignored and `/opt/gcc-latest` wins — diverging from scripts/cmake.py, which honors `CXX` alone and derives CC. The workflow comment ("Set CXX/CC explicitly so scripts/bazel.py does not pick a stray /opt/gcc-latest") shows the trap. Suggest keying on CXX alone like cmake.py.

scripts/bazel.py
```diff
@@ -0,0 +1,191 @@
...
+        existing_ld_path = bazel_environment.get("LD_LIBRARY_PATH", "")
+        full_ld_path = (
+            f"{libstdcxx_directory}:{existing_ld_path}"
+            if existing_ld_path
+            else libstdcxx_directory
+        )
+        bazel_command.append(f"--test_env=LD_LIBRARY_PATH={full_ld_path}")
+        bazel_environment["LD_LIBRARY_PATH"] = full_ld_path
+
+    bazel_command.append("//...")
```
philipcraig, line 119: Passthrough arguments land after the hardcoded `//...`, and Bazel unions target patterns, so `./scripts/bazel.sh test //tutorials:reflection_tutorial` still builds and runs everything. Append extras before `//...`, or only default to `//...` when no target was given.

BUILD.bazel
```diff
@@ -0,0 +1,31 @@
...
+    name = "protocol",
+    hdrs = ["protocol.hh"],
+)
+
+cc_library(
+    name = "consteval_check",
+    hdrs = ["consteval_check.h"],
+)
+
+cc_test(
```
philipcraig, line 18: The Bazel build compiles without the `-Werror;-Wall` that cmake/xyz_add_test.cmake applies to every test and tutorial, so a locally green Bazel iteration can be rejected by the required CMake CI job. Add the equivalent copts (e.g. in `.bazelrc`).

MODULE.bazel
```diff
@@ -0,0 +1,7 @@
+module(
+    name = "protocol",
+    version = "0.0.1",
+)
+
+bazel_dep(name = "rules_cc", version = "0.2.22")
+bazel_dep(name = "googletest", version = "1.18.0")
```
philipcraig, line 7: googletest 1.18.0 here vs the v1.15.2 FetchContent pin in CMakeLists.txt — the two builds test against different GoogleTest versions, inviting CMake-green/Bazel-red splits. Pin the same version in both.

scripts/bazel_compiler_wrapper.py
```diff
@@ -0,0 +1,139 @@
...
+    seen_libraries: Set[str] = set()
+    for argument in raw_arguments:
+        if argument.startswith("@") and os.path.isfile(argument[1:]):
+            response_file_path = argument[1:]
+            new_response_path = process_response_file(response_file_path)
+            target_command.append("@" + new_response_path)
+        else:
+            transformed = transform_argument(argument)
+            if transformed is not None:
+                if transformed.startswith("-l"):
```
philipcraig: The `-l` dedup (here and in `process_response_file`) is unconditional, but the rationale is Apple-ld-only. GNU ld resolves static archives left-to-right, so a legitimately repeated library gets silently rewritten and later-needed symbols go unresolved. Gate it on `sys.platform == "darwin"`.

scripts/bazel_compiler_wrapper.py
```diff
@@ -0,0 +1,139 @@
...
+import os
+import shutil
+import subprocess
+import sys
+from typing import List
+from typing import Optional
+from typing import Set
+from typing import Tuple
+
+CLANG_ONLY_FLAG_PREFIXES: Tuple[str, ...] = (
```
philipcraig: The wrapper solves a macOS-only problem (Homebrew symlinks, Clang-only flags) but is interposed on every platform, adding interpreter startup plus a response-file rewrite to each Linux action for a filter that matches nothing there. Create the wrappers only on darwin.

scripts/bazel.py
```diff
@@ -0,0 +1,191 @@
...
+    if cxx_compiler_path.startswith(gcc_latest_root):
+        for candidate_subpath in ("lib64", "lib"):
+            candidate_directory = os.path.join(gcc_latest_root, candidate_subpath)
+            if os.path.isdir(candidate_directory):
+                return candidate_directory
+
+    return None
+
+
+def find_system_compilers() -> Tuple[str, str]:
```
philipcraig: `find_system_compilers` and `find_libstdcxx_directory` re-implement scripts/cmake.py's compiler discovery and CMakeLists.txt's libstdc++ rpath logic, and have already diverged (bare `g++` fallback cmake.py deliberately omits; the CC/CXX handling above). Suggest a shared helper module both entry points import so toolchain policy lives in one place.

philipcraig (changes requested):
Bazel support is a welcome addition, but a few gaps let the two build systems diverge in ways that undermine the nightly trunk job and day-to-day use. Requesting changes primarily for the cache-correctness, `.bazelignore`, and toolchain-parity items below.

jbcoe:
I will remove mac OS bazel support from this PR as it is significantly complicated and best addressed independently.

CONTRIBUTING.md
jbcoe: Happy to handle this in a follow-up PR. 

`bazel {build|test} //...` works on CI and in the devcontainer and agentic-sandbox image.

scripts/bazel.py
jbcoe: `./scripts/bazel.sh test //tutorials:reflection_tutorial` is not intended usage.

Like the cmake script, bazel.sh is a quick way to build and test all the things.

scripts/bazel_compiler_wrapper.py
jbcoe: The compiler wrapper is now removed.

scripts/bazel_compiler_wrapper.py
jbcoe: This is now removed.

.bazelignore
jbcoe: I've modified scripts/cmake.py to confine build dirs to a single (bazel) ignored parent.

scripts/bazel.py
jbcoe: `//tutorials:reflection_tutorial` and the like are rejected - this is not intended use

philipcraig (commented):
Re-review at 53d78fa. The blocking items from my previous review are resolved: build directories are confined under `build/`, the compiler wrapper is gone and concrete compiler paths now feed the cache keys, `.bazelrc` matches xyz_add_test.cmake's warning set, the googletest pins agree at 1.18.0, compiler selection is shared via scripts/compiler_discovery.py, and CONTRIBUTING.md warns about bare invocations. Three follow-ups below; the paths-filter one affects the required CMake checks.

.github/workflows/bazel.yml
```diff
@@ -0,0 +1,140 @@
...
+      - "**/*.cc"
+      - "**/*.h"
+      - "**/*.hh"
+      - "**/*.bazel"
+      - "BUILD.bazel"
+      - "MODULE.bazel"
+      - "MODULE.bazel.lock"
+      - ".bazelrc"
+      - ".bazelignore"
+      - "scripts/bazel.py"
```
philipcraig, line 19: `scripts/compiler_discovery.py` is missing from this list and the detect-changes filter below — and from the equivalent lists in cmake.yml and sanitizers.yml, whose entry points now import it. A PR touching only that file skips the required `GCC trunk Debug`/`Release` checks (a job skipped by detect-changes counts as passing) and never runs the Bazel or sanitizer builds.

scripts/bazel.py
```diff
@@ -0,0 +1,99 @@
...
+    # than bare names resolved via PATH. If none was found, leave CC/CXX
+    # unset so Bazel falls through to its own default toolchain and reports
+    # that it lacks reflection support, matching scripts/cmake.py's policy.
+    if cxx_path is not None:
+        cxx_target_path = shutil.which(cxx_path) or cxx_path
+        bazel_environment["CXX"] = cxx_target_path
+        bazel_command.append(f"--action_env=CXX={cxx_target_path}")
+        bazel_command.append(f"--repo_env=CXX={cxx_target_path}")
+
+        if cc_path is not None:
```
philipcraig, line 86: Bazel's autoconfigured C++ toolchain selects the compiler from `CC` (or `BAZEL_CC`) and never reads `CXX`, so for a caller following scripts/cmake.py's convention of exporting `CXX` alone, everything here except the rpath logic is inert and Bazel silently builds with the default `/usr/bin/gcc`. When `cc_path` is None, derive it from `cxx_target_path` (the sibling `gcc` next to `g++`) rather than leaving `CC` unset.

scripts/bazel.py
```diff
@@ -0,0 +1,99 @@
...
+
+    # Bazel's repository rules need concrete, absolute compiler paths rather
+    # than bare names resolved via PATH. If none was found, leave CC/CXX
+    # unset so Bazel falls through to its own default toolchain and reports
+    # that it lacks reflection support, matching scripts/cmake.py's policy.
+    if cxx_path is not None:
+        cxx_target_path = shutil.which(cxx_path) or cxx_path
+        bazel_environment["CXX"] = cxx_target_path
+        bazel_command.append(f"--action_env=CXX={cxx_target_path}")
+        bazel_command.append(f"--repo_env=CXX={cxx_target_path}")
```
philipcraig, line 72: The concrete path fixes the cross-compiler half of the earlier cache concern. The in-place half remains: `/opt/gcc-latest` is replaced weekly at a constant path, the compiler binary is not an action input, and bazel.yml restores the disk cache across nightly runs — so the scheduled `GCC trunk` job can replay objects compiled by last week's snapshot without exercising the new one. Embedding the compiler's identity in every action key closes it, e.g. `--action_env=XYZ_PROTOCOL_COMPILER_VERSION=` set to the first line of `{cxx_target_path} --version` (plus the matching `--repo_env`).

.bazelignore
```diff
@@ -0,0 +1,2 @@
...
```
philipcraig, line 1: Optional: existing checkouts still holding pre-confinement directories at the root (`build-asan/`, `build-reflection/`, …) will fail `bazel build //...` analysis on the `_deps` BUILD files inside them until deleted. Keeping the old common names as extra literal lines here is cheap backward compatibility.

jbcoe: I'll keep .bazelignore clean. The value_types project worked fine with a minimal ignore file:

https://github.com/jbcoe/value_types/blob/main/.bazelignore

philipcraig (approved):
All points from both rounds are addressed — shared compiler discovery module, compiler-version stamping in the action cache keys, CC derived from CXX's sibling, warning-flag and googletest parity with CMake, and compiler_discovery.py in all three workflows' paths filters. Nice simplification dropping the wrapper scripts.

PR #256, opened 2026-08-30
Use Itanium-ABI-like name mangling for vtable entries

### LINKED ISSUE #254: Use name mangling for the vtable
Use the name mangling approach from the advanced_vtable turtorial for the vtable in protocol to support narrowing conversions.

jbcoe:
Closes #254 

The name mangler intentionally duplicates and extends the name mangling logic from the tutorial `tutorials/advanced_vtables.cc`.

Some parameter types are intentionally unsupported. Using unsupported types as parameters will fail noisily rather than result in failed conformance checks or odd behaviour at runtime.

### DIFF
--- a/name_mangling.h
+++ b/name_mangling.h
@@ -20,20 +20,12 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 #ifndef XYZ_REFLECTION_NAME_MANGLING_H_
 #define XYZ_REFLECTION_NAME_MANGLING_H_
 
-// Names a member function by mangling its signature, following the Itanium
-// C++ ABI where it has an encoding: the function name (or `cl` for the call
-// operator), wrapped in `N...E` with its cv- and ref-qualifiers when it has
-// any, followed by the parameter types. The name depends only on the
-// signature, so a member function of one class names the same entry as a
-// member function of the same signature on another class; this lets a
-// vtable's entries be found by the signature of the interface member function
-// they implement, rather than by declaration order.
-//
-// Parameter types may be fundamental, pointers, references, arrays,
-// functions, and (possibly cv-qualified) named class, union and enumeration
-// types, including class-template specialisations with type, integral or
-// enumeration template arguments. Naming an entry for any other parameter
-// type is ill-formed.
+// Names a member function with a string that's a valid C++ identifier: only
+// letters, digits and underscore, none of the characters (`:`, `(`, `)`,
+// `<`, `>`) that a plain rendering of a signature would contain. The
+// encoding reuses the Itanium C++ ABI's member-mangling scheme, which
+// already solves this problem for linker symbol names, rather than
+// inventing a new one.
 
 #include <array>
 #include <charconv>
@@ -248,11 +240,10 @@ consteval std::string base_name_of(std::meta::info function) {
 
 }  // namespace detail
 
-// Names a vtable entry for the member function `function`. Distinguishes
-// overloads by folding the function's cv- and ref-qualification, then each
-// parameter's type, into the name. Matches the Itanium ABI's own placement:
-// cv-/ref-qualifiers sit inside `N...E`, wrapping the function name, ahead of
-// the parameter list.
+// Names the member function `function`, distinguishing overloads by folding
+// its cv- and ref-qualification, then each parameter's type, into the name.
+// Matches the Itanium ABI's own placement: cv-/ref-qualifiers sit inside
+// `N...E`, wrapping the function name, ahead of the parameter list.
 consteval std::string mangle(std::meta::info function) {
   std::string qualifiers;
   if (is_volatile(function)) qualifiers += "V";
--- a/name_mangling.h
+++ b/name_mangling.h
@@ -1,0 +1,274 @@
+/* Copyright (c) 2025 The XYZ Protocol Authors. All Rights Reserved.
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of
+this software and associated documentation files (the "Software"), to deal in
+the Software without restriction, including without limitation the rights to
+use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+the Software, and to permit persons to whom the Software is furnished to do so,
+subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+==============================================================================*/
+#ifndef XYZ_REFLECTION_NAME_MANGLING_H_
+#define XYZ_REFLECTION_NAME_MANGLING_H_
+
+// Names a member function by mangling its signature, following the Itanium
+// C++ ABI where it has an encoding: the function name (or `cl` for the call
+// operator), wrapped in `N...E` with its cv- and ref-qualifiers when it has
+// any, followed by the parameter types. The name depends only on the
+// signature, so a member function of one class names the same entry as a
+// member function of the same signature on another class; this lets a
+// vtable's entries be found by the signature of the interface member function
+// they implement, rather than by declaration order.
+//
+// Parameter types may be fundamental, pointers, references, arrays,
+// functions, and (possibly cv-qualified) named class, union and enumeration
+// types, including class-template specialisations with type, integral or
+// enumeration template arguments. Naming an entry for any other parameter
+// type is ill-formed.
+
+#include <array>
+#include <charconv>
+#include <limits>
+#include <meta>
+#include <optional>
+#include <ranges>
+#include <stdexcept>
+#include <string>
+#include <string_view>
+#include <type_traits>
+#include <vector>
+
+namespace xyz::name_mangling {
+
+namespace detail {
+
+consteval std::string decimal(std::size_t value) {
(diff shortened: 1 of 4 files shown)


RyanJK5:
I think #244 should be built on this.

name_mangling.h
```diff
@@ -0,0 +1,265 @@
...
+// its cv- and ref-qualification, then each parameter's type, into the name.
+// Matches the Itanium ABI's own placement: cv-/ref-qualifiers sit inside
+// `N...E`, wrapping the function name, ahead of the parameter list.
+consteval std::string mangle(std::meta::info function) {
+  std::string qualifiers;
+  if (is_volatile(function)) qualifiers += "V";
+  if (is_const(function)) qualifiers += "K";
+  if (is_lvalue_reference_qualified(function)) qualifiers += "R";
+  if (is_rvalue_reference_qualified(function)) qualifiers += "O";
+  std::string atom = detail::mangle_atom(detail::base_name_of(function));
```
philipcraig: `base_name_of` returns the Itanium `<operator-name>` `cl`, but passing it through `mangle_atom` length-prefixes it to `2cl` — identical to the atom for a member function literally named `cl`. With `struct I { int operator()(int) const; int cl(int) const; };` both members mangle to `fn_NK2clEi`, and since `same_name` treats them as distinct members, `generate_vtable_specs` emits two `data_member_spec`s with the same name. The Itanium ABI leaves operator codes un-prefixed precisely so `cl` can never alias the source-name `2cl` — skip `mangle_atom` for operator names.

name_mangling.h
```diff
@@ -0,0 +1,265 @@
...
+    for (std::meta::info parameter : parameters) {
+      out += mangle_type(is_type(parameter) ? parameter : type_of(parameter));
+    }
+    return out + "E";
+  }
+  if (auto code = builtin_type_code(type)) return std::string(*code);
+  if (is_class_type(type) || is_union_type(type) || is_enum_type(type)) {
+    return mangle_qualified_name(type);
+  }
+  throw std::runtime_error("name mangling: unsupported parameter type");
```
philipcraig, line 233: `mangle_type` has no case for pointer-to-member types (`is_pointer_type` is false for them; Itanium spells them `M<class><member type>`), and `builtin_type_code` covers no extended fundamentals, so `void bind(void (Widget::*)(int))` and `void f(__int128)` — both accepted under the index-based names — now fail at `define_aggregate` time.

name_mangling.h
```diff
@@ -0,0 +1,265 @@
...
+  }
+  if (is_enum_type(type)) {
+    std::vector<std::meta::info> enumerators = enumerators_of(type);
+    for (std::meta::info enumerator : enumerators) {
+      if (constant_of(enumerator) == argument) {
+        return out + mangle_atom(identifier_of(enumerator)) + "E";
+      }
+    }
+  }
+  throw std::runtime_error(
```
philipcraig, line 112: Only `bool`, the 15 listed integral types, and enum values exactly equal to an enumerator get through. Every other legal template argument in a parameter type makes the whole `protocol<I>` ill-formed: floating-point (`Gain<1.5>`), class-type (`Named<fixed_string("x")>`), pointer (`PtrHolder<&g>` — `is_value` is true but it falls off the end of this function), reference (`RefHolder<g>` — reflects as an object, hits the `else` in `mangle_scope_component`), template-template arguments, and enum values matching no enumerator (`Opt<Flags(3)>` bitmasks, and any `std::byte` argument since it has no enumerators). All of these compile on `main`.

name_mangling.h
```diff
@@ -0,0 +1,265 @@
...
+
+// Mangles one scope level (a namespace, class, enumeration or function),
+// folding in template arguments when `scope` is a class-template
+// specialisation.
+consteval std::string mangle_scope_component(std::meta::info scope) {
+  if (is_namespace(scope) && !has_identifier(scope)) return "12_GLOBAL__N_1";
+  std::meta::info named =
+      has_template_arguments(scope) ? template_of(scope) : scope;
+  if (!has_identifier(named)) {
+    throw std::runtime_error("name mangling: unnamed types are not supported");
```
philipcraig, line 124: A closure-type parameter lands on this throw: `is_namespace(scope)` is false and closures have no identifier. So `auto cmp = [](int a, int b) { return a < b; };` `struct I { void sort_with(decltype(cmp) c); };` no longer compiles — awkward given `is_maybe_lambda` exists to support lambda-shaped types. A typedef'd anonymous enum likely hits it too, since `mangle_type` dealiases first.

protocol.hh
```diff
@@ -201,17 +205,18 @@ consteval std::meta::info vtable_entry_at() {
...
 template <typename R, typename... Args, typename EnclosingType,
-          typename ProtocolType, typename Vtable, std::size_t Index,
+          typename ProtocolType, typename Vtable, std::meta::info Member,
           bool IsConst, bool IsNoexcept>
-struct method_thunk<R (*)(Args...), EnclosingType, ProtocolType, Vtable, Index,
+struct method_thunk<R (*)(Args...), EnclosingType, ProtocolType, Vtable, Member,
                     IsConst, IsNoexcept> {
   static constexpr std::meta::info vtable_entry =
-      vtable_entry_at<^^Vtable, Index>();
+      *find_vtable_entry<^^Vtable>(Member);
```
philipcraig: Unchecked `*` on the optional, here and in `make_view_vtable`. nullopt is unreachable only while `generate_vtable_specs` and both lookups run the same mangler over the same member list — the invariant `vtable_entry_at` enforced by construction. If they diverge (e.g. a const-only vtable), this surfaces as "not a constant expression" pointing into `<optional>`. Throwing a message that names the member (as `name_mangling.h` does) and returning `std::meta::info` keeps the failure legible and drops the `<optional>` include.

name_mangling.h
```diff
@@ -0,0 +1,265 @@
...
+
+template <typename T>
+consteval std::string signed_decimal(std::meta::info argument) {
+  T value = extract<T>(argument);
+  if constexpr (std::is_signed_v<T>) {
+    if (value < 0) {
+      return "n" + decimal(std::size_t{0} - static_cast<std::size_t>(value));
+    }
+  }
+  return decimal(static_cast<std::size_t>(value));
```
philipcraig: Funnelling through `std::size_t` truncates 64-bit values modulo 2^32 where `size_t` is 32 bits: `Tagged<0LL>` and `Tagged<4294967296LL>` both mangle their argument as `0` (and the negative branch gives `n1` for both `-1LL` and `-4294967297LL`), so overloads distinguished only by such arguments produce duplicate entry names. Casting through `std::make_unsigned_t<T>` keeps the width tied to `T`; LP64 output is unchanged.

name_mangling.h
```diff
@@ -0,0 +1,265 @@
...
+  return std::string(identifier_of(function));
+}
+
+}  // namespace detail
+
+// Names the member function `function`, distinguishing overloads by folding
+// its cv- and ref-qualification, then each parameter's type, into the name.
+// Matches the Itanium ABI's own placement: cv-/ref-qualifiers sit inside
+// `N...E`, wrapping the function name, ahead of the parameter list.
+consteval std::string mangle(std::meta::info function) {
```
philipcraig, line 262: The key excludes return type and `noexcept`, but `same_name_and_parameters` compares dealiased return types and `member_function_conforms_to` compares `noexcept` — so for #244's name-based conformance, `J::get() const -> long` and `I::get() const -> int` both mangle to `fn_NK3getE`: the lookup matches, then the fn-ptr types mismatch at vtable assignment instead of a clean non-conformance. Harmless within a single vtable today (overloads can't differ only in return type), but `mangle_type` already emits `F<ret><params>E`, so folding return type and `noexcept` in now avoids re-keying every entry later.

protocol.hh
```diff
@@ -178,20 +176,26 @@ constexpr inline auto conformance_candidates_of =
...
-template <std::meta::info VtableType, std::size_t Index>
-consteval std::meta::info vtable_entry_at() {
-  return nonstatic_data_members_of(
-      VtableType, std::meta::access_context::unprivileged())[Index];
+// The entry of `VtableType` named by the mangled signature of the interface
+// member function `member`, if it declares one.
+template <std::meta::info VtableType>
+consteval std::optional<std::meta::info> find_vtable_entry(
+    std::meta::info member) {
+  std::string name = xyz::name_mangling::mangle(member);
```
philipcraig: Every call re-mangles `member` (recursive string concatenation, no substitution compression) and linearly compares against all entries, and `make_view_vtable` does this for each member per (interface, T) instantiation — strings `generate_vtable_specs` already computed. Memoizing per member, e.g. `template <std::meta::info M> constexpr inline auto mangled_name_of = std::define_static_string(mangle(M));`, mirrors the existing `define_static_array` caching above.

philipcraig (changes requested):
The index-based naming this replaces was total and injective by construction; the mangler as it stands is neither — several interfaces that compile on `main` become ill-formed on this branch, and one encoding collision is possible. Details inline.

One point with no diff line to anchor to: the mangler was lifted from `tutorials/advanced_vtables.cc` without switching the tutorial over, leaving ~180 near-identical lines that have already diverged in both directions (the tutorial keeps the general `operator_code` table but has 3 builtins and no template-argument support; this header has 21 builtins and template arguments but only `operator()`). Suggest the tutorial include `name_mangling.h`, lifting its `operator_code` table into `base_name_of` on the way — that also gives non-`()` operators a clean diagnostic instead of an `identifier_of` failure.

name_mangling.h
jbcoe: We will leave this unsupported for now. Using pointer-to-member types as parameters will fail noisily rather than result in failed conformance checks or odd behaviour at runtime.

name_mangling.h
jbcoe: We will leave these as unsupported for now. Using unsupported types as parameters will fail noisily rather than result in failed conformance checks or odd behaviour at runtime.

name_mangling.h
jbcoe: We will leave these as unsupported for now. Using unsupported types as parameters will fail noisily rather than result in failed conformance checks or odd behaviour at runtime.

philipcraig (approved):
All eight points are addressed:

- The `cl`/`2cl` collision is gone (operator codes now bypass `mangle_atom`), the 64-bit truncation is fixed via `make_unsigned_t`, the optional dereference is now a throwing `find_vtable_entry` that names the missing entry, and `mangled_name_of<Member>` caches the mangling per member. All with tests where they count.
- Folding return type and noexcept-ness in now, with the comment explaining the cross-interface motivation, saves #244 a re-keying later.
- On the intentionally unsupported parameter types: agreed, and the failure is genuinely noisy — every unsupported path ends in a consteval throw whose message reaches the diagnostic. One correction to my earlier comment: the pointer-NTTP case never flowed off the end of `mangle_template_argument_value`; the trailing throw was already there. `__int128`/`unsigned __int128` joining the builtin table is a bonus.

Keeping the tutorial's copy of the mangler separate is your call; the PR description saying so is enough.

PR #260, opened 2026-08-31
Reject ref-qualified interface member functions

### LINKED ISSUE #230: Ref-qualified interface member functions conform but are not forwarded in the reflection implementation
`member_function_conforms_to` requires the candidate's `&`/`&&` qualifier to
match the interface's (see the `LvalueRefQualifierMustMatch` /
`RvalueRefQualifierMustMatch` tests), but `method_thunk` only synthesises an
unqualified `operator()`, and the view trampolines call the member on an
lvalue. So for

    struct I { int take() &&; };
    struct Impl { int take() && { return 1; } };
    protocol<I> p(Impl{});
    std::move(p).take();

conformance passes but the call is ill-formed: the trampoline calls
`static_cast<Impl*>(ptr)->take()` on an lvalue. `&`-qualified members work
by accident (the trampoline always has an lvalue), which also means
`std::move(p).f()` calls an `&`-qualified `f` through an rvalue `protocol`.

Marked `TODO(jbcoe)` on `method_thunk` in `reflection/protocol.hh`. Options:

1. Generate `&` / `&&` qualified `operator()` on the thunk and forward the
   value category through the vtable (`static_cast<U&&>(*ptr).f(...)` for
   `&&`).
2. Reject ref-qualified interface members in `protocol_interface_functions_of`
   with an `XYZ_CONSTEVAL_CHECK`, until 1 is done.

Found while writing the forwarding tests for #145 (#228).

philipcraig:
Ref-qualified interface members conformed, but the synthesised thunks call the target on an lvalue and cannot forward the protocol object's value category: `std::move(p).take()` would call `take() &&` on an lvalue. Reject them when enumerating interface member functions, until the thunks forward value categories.

`is_protocol_conformant` throws the rejection from the function itself so a consteval caller can catch it; instantiating `protocol<I>`/`protocol_view<I>` with a ref-qualified interface member remains a hard error.

Closes #230.

### DIFF
--- a/protocol.hh
+++ b/protocol.hh
@@ -177,12 +177,30 @@ constexpr inline auto conformance_candidates_of =
 // The non-static members of `conformance_candidates_of<Type>`: the member
 // functions an interface `Type` requires. Overloads appear as separate
 // entries, each naming its own vtable entry (see
-// `xyz::name_mangling::mangle`).
+// `xyz::name_mangling::mangle`). Ref-qualified members are rejected: the
+// synthesised thunks call the target on an lvalue and cannot forward the
+// protocol object's value category (see the TODO on `method_thunk`).
+template <std::meta::info Type>
+consteval std::vector<std::meta::info> protocol_interface_function_infos() {
+  std::vector<std::meta::info> result;
+  for (std::meta::info member : conformance_candidates_of<Type>) {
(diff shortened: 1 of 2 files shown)


jbcoe (approved):
Thanks for this.

PR #263, opened 2026-08-31
Document the Bazel entrypoint in the agent context file

philipcraig:
GEMINI.md (symlinked as Agents.md and CLAUDE.md) still described scripts/cmake.sh as the only build entrypoint. Add scripts/bazel.sh alongside it and note that both entrypoints select a reflection-capable compiler, so a bare cmake/bazel invocation fails on stock GCC.


jbcoe (approved):
Thanks for this.

Issue #257, opened 2026-08-30
Measure and improve test coverage


jbcoe:
Measuring test coverage for `consteval` code is probably a relatively new problem. That shall not stand in our way! 🚀

## Tasks

- [x] Add coverage measurement — runtime gcov plus consteval trap probing: #264
- [x] Set a real `codecov/project` target in `codecov.yml` once `main` has a baseline from the new setup
- [ ] Consider making the Codecov statuses required checks
- [x] Increase coverage from the measured runtime gaps: #265 (the remaining unevaluated consteval line, the `find_vtable_entry` throw at protocol.hh:221, is a defensive internal check that is unreachable by construction)

philipcraig:
Measurement half implemented; PR to follow. gcov cannot see consteval code (it marks it non-executable), and the usual constexpr workaround — running tests at both compile time and runtime — does not apply to immediate functions, so `scripts/consteval_coverage.sh` measures it by compile-time trap probing: instrument a copy of protocol.hh, then recompile the test translation units once per probe point with one trap armed; a compile failure proves the test suite evaluated that line. Current numbers: consteval 71/72 probe points — only the internal `find_vtable_entry` throw (protocol.hh:221) is unevaluated; runtime 92% overall, protocol.hh 86%, with the gaps in `create`'s exception path and some instantiations' copy/move thunks. If P2758 compile-time printing lands in GCC, the probing loop can collapse to a single compile.

philipcraig:
Three of the four items are done. Holding the last — making the Codecov statuses required checks — until the new project gate (target: auto, 1% threshold) has run cleanly on a few more pull requests.

philipcraig:
Considered and decided against for now. The `target: auto` / 1% gate has proved sound (40 pull-request heads since #289, both failures explained), but the `consteval coverage` job takes about 15 minutes, 11 of them in the probe loop, while every other warm check finishes in under 4. Requiring `codecov/project` and `codecov/patch` would put coverage on the critical path of every pull request, and coverage.yml's `paths:` filter would also have to go so non-C++ pull requests get a check at all. The statuses stay advisory. Revisit if the probe loop gets down to build-job length, for example by splitting the probes across a job matrix.

PR #264, opened 2026-08-31
Measure test coverage of runtime and consteval code

philipcraig:
`scripts/cmake.sh test --coverage` builds the Debug preset with gcov
instrumentation, runs the tests and reports with gcovr, writing an LCOV
trace for Codecov. gcov marks consteval code as non-executable: constant
evaluation happens inside the compiler, and running tests at both
compile time and runtime - the usual constexpr workaround - is
impossible for immediate functions.

`scripts/consteval_coverage.sh` measures the consteval code in
protocol.hh instead, by compile-time trap probing: it instruments a copy
of the header with trap calls at block entries and return/throw
statements (line numbers preserved), then recompiles the
protocol-instantiating test translation units once per probe point with
-fsyntax-only, arming one trap at a time; a compile failure proves the
test suite evaluated that line, because a constant evaluation has no
other observable side effect. Runtime thunk lambdas inside the vtable
builders are excluded: they execute at runtime, where gcov measures
them.

The Coverage workflow runs both measurements and uploads them to Codecov
under the runtime and consteval flags.

Towards #257.


jbcoe:
Great progress! Thanks for this.

philipcraig:
`codecov/project` is red only because Codecov compares against a pre-existing upload at 9531f89 made before this measurement setup existed, so the -4.37% is a methodology artifact rather than a regression (`codecov/patch` passes, and the status is not a required check). Once this merges, the Coverage workflow establishes a fresh baseline on `main` and subsequent pull requests compare like-for-like.

scripts/consteval_coverage.py
jbcoe, line 1: This is innovative and I wonder if there is an existing solution that we can reuse or reference? 

Innovation is best avoided where possible.

jbcoe (approved):
Thanks for this. 

Data is good and this will improve our understanding of our code.

scripts/consteval_coverage.py
philipcraig: I searched for prior art and didn't find it

philipcraig: There's no existing tool to reuse: gcov and llvm-cov only count runtime execution, and consteval bodies compile to nothing runnable — gcov marks them non-executable (verified on GCC trunk). The established workaround for constexpr code is to run every test at both compile time and runtime so runtime instrumentation sees it (Catch2's `CATCH_CONFIG_RUNTIME_STATIC_REQUIRE`, catchorg/Catch2#2350; doctest/doctest#250), which cannot apply to immediate functions — and P2996 metafunction calls make every caller immediate. Coco (commercial) instruments source but still measures runtime execution only. Aborting compilation from an armed trap is the one observable a constant evaluation has; when P2758 compile-time printing lands in GCC, the probing loop can collapse to a single compile.

PR #265, opened 2026-08-31
Add tests for unexercised allocator and assignment paths

philipcraig:
Re-enable the four commented-out exception-safety tests, and add tests
for the paths the new coverage measurement reported unexercised: the
unequal-allocator move slow paths, self-assignment, the release of the
owning allocation when in-place construction throws, and the null
vtable's copy, move and destroy entries.

Towards #257.


jbcoe (approved):
Thanks for this.

PR #266, opened 2026-08-31
Set a real codecov project target against the new baseline

philipcraig:
Replace the interim informational project status with target: auto and
a 1% threshold now that main uploads a like-for-like baseline, and
extend the ignore list to the *_tests.cc files it was always meant to
cover.

Towards #257.


jbcoe (approved):
Thanks for this.

Issue #249, opened 2026-08-30
Add the required status checks to the `main` ruleset once #248 merges


philipcraig:
Follow-up to #213 / #248. Once #248 is merged, the workflows report these checks on every pull request (skipping via `detect-changes` when nothing relevant changed), so they can be made required without blocking non-C++ pull requests.

In *Settings → Rules → Rulesets → main → Require status checks to pass*, add:

- `GCC trunk Release`
- `GCC trunk Debug`
- `GCC-16 Release`
- `GCC-16 Debug`
- `asan`
- `tsan`
- `uv-lock`

`pre-commit` is already required. Leave the two `detect-changes` jobs out.

Do not enable these before #248 merges: with the current `paths:` filters a required check that never reports leaves every non-C++ pull request stuck on "Expected — Waiting for status to be reported".

#213 can be closed once the ruleset is updated.

philipcraig:
The list above can't be enabled as written. A matrix job skipped by its job-level `if:` reports a single check under the unexpanded job name — on #263 (docs-only) the skipped checks were literally `${{ matrix.settings.name }} ${{ matrix.configuration }}` and `${{ matrix.sanitizer }}` — so `GCC trunk Release`, `asan`, etc. never report and a docs-only pull request would be stuck on "Expected — Waiting for status to be reported". The skipped-job approach only works for non-matrix jobs.

#271 adds a fixed-name summary job to each build workflow that always reports and fails when a build it depends on fails. Once it merges, the entries to add under *Settings → Rules → Rulesets → main → Require status checks to pass* are:

- `cmake`
- `bazel`
- `sanitizers`
- `uv-lock`

`pre-commit` stays required; leave the `detect-changes` jobs out.

philipcraig:
Checked the run history since #271 merged on 2026-09-01. The four entries above are safe to require now.

- Every pull request head since then reports `cmake`, `bazel`, `sanitizers` and `uv-lock`: 26 pull requests, 12 of them touching no C++ or build files (#289, #291, #299, #301, #302, #303, #308, #309, #313, #316, #319 among them). None waited on "Expected".
- The summary jobs go red when a build fails. The three build failures in roughly 140 pull-request runs per workflow were all real (#276's `is_maybe_lambda` typo, a branch built on top of it, and work in progress on #306), and `cmake`, `bazel` and `sanitizers` failed each time. The two `uv-lock` failures were real lockfile mismatches, fixed before merge.
[...]

PR #271, opened 2026-08-31
Add fixed-name summary jobs for required status checks

philipcraig:
A matrix job skipped by its job-level `if:` reports a single check under its unexpanded name, so the per-configuration checks from #248 never report on pull requests where `detect-changes` skips the build; requiring them would leave docs-only pull requests waiting on checks that never arrive (see the checks on #263 for an example). Add a fixed-name, non-matrix `summary` job to each build workflow that passes when the build was skipped and fails when it fails. The ruleset entries for #249 become `cmake`, `bazel`, `sanitizers`, and `uv-lock`.

Towards #213.


jbcoe (approved):
Thanks for this.

Issue #208, opened 2026-08-28
Decide how a valueless protocol should behave when a member function is called


philipcraig:
PR #207 (for #201) gives `protocol` a vtable whose member function entries are null in `null_vtable`, the stand-in vtable installed after a move. Calling a member function on a valueless `protocol` is therefore a precondition violation that currently manifests as a call through a null function pointer.

This matches `std::polymorphic`'s contract for valueless objects, but nothing exercises or documents it beyond a comment in `reflection/protocol.hh`.

Options to consider:

- Leave it as undefined behaviour and document the precondition in the paper's wording for `protocol`.
- Point `null_vtable`'s member function entries at trampolines that `assert`/`std::terminate` so misuse is diagnosed in debug builds, and add a death test.
- Use a C++26 contract precondition on the synthesised thunks once the toolchain supports it.

jbcoe:
I'm happy to leave this as UB and let contract checks for `valueless_after_move` handle `polymorphic` and `protocol` at the same time.

We can modify our implementation to make the UB be a call to `std::terminate` if it does not complicate code significantly.

RyanJK5:
> Point null_vtable's member function entries at trampolines that assert/std::terminate so misuse is diagnosed in debug builds, and add a death test.

I think a simple null check in the dispatch function would suffice, which would then be disabled on release builds. The null vtable is a good optimization trick but less critical for a debug build IMO.

I'll add the assertion branch since it should be fairly simple. We can change that to a contract assertion later.

PR #272, opened 2026-08-31
Allows protocol_view (for T and const T) to be constructed implicitly.

jbcoe:
Constructing a `protocol_view` is now implicit, so a `protocol` or a conforming object converts to a view wherever one is expected without an explicit cast.

Adds deleted constructors to avoid constructing views from temporaries.

### DIFF
--- a/protocol.hh
+++ b/protocol.hh
@@ -1008,21 +1008,25 @@ class protocol_view
   protocol_view& operator=(protocol_view&&) noexcept = default;
   ~protocol_view() = default;
 
-  // Construct from any non-const type U that conforms to the Interface T.
+  // Views a non-const object that conforms to the Interface T. Implicit
+  // because the lvalue-reference parameter already rejects temporaries, so a
+  // view can never outlive what it views.
   template <typename U>
     requires is_protocol_conformant_v<T, std::remove_cvref_t<U>> &&
                  (!is_protocol_view_v<std::remove_cvref_t<U>>) &&
                  (!std::is_const_v<U>)
-  explicit protocol_view(U& object)
+  protocol_view(U& object)
       : object_(static_cast<void*>(std::addressof(object))),
         vtable_(
             &detail::view_vtable_for<T, U, detail::const_policy::all_const>) {}
 
-  // Views the object a `protocol<T>` owns, sharing its vtable.
+  // Views the object a `protocol<T>` owns, sharing its vtable. Implicit
+  // because a view is a non-owning handle and the deleted rvalue overload
+  // below rejects temporaries, so nothing can dangle.
   //
   // Precondition: `p` is not valueless.
   template <typename Alloc>
-  explicit protocol_view(protocol<T, Alloc>& p) noexcept
+  protocol_view(protocol<T, Alloc>& p) noexcept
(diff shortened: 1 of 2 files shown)


jbcoe:
@philipcraig @RyanJK5 

This is simple code but there is a design question being addressed here too. 

All are in scope for review.

RyanJK5:
> @philipcraig @RyanJK5
> 
> This is simple code but there is a design question being addressed here too.
> 
> All are in scope for review.

I think implicit is good for `protocol_view`. Making `protocol` constructors implicit is a more difficult question. I will take a look now.

jbcoe:
`protocol` constructors should not be implicit as they allocate memory.

RyanJK5 (approved):
This looks good - is there any coverage of a `protocol_view` from non-`const` `protocol&&` or `T&&`? Is that meant to be allowed?

EDIT: I see now, it just wasn't touched in the diff. LGTM!

RyanJK5:
> `protocol` constructors should not be implicit as they allocate memory.

I am inclined to agree. I have heard some others say they should be implicit because `std::function` has converting constructors.

jbcoe:
`protocol_view` from non-`const` `protocol&&` or `T&&` should be deleted.

`std::function` is an old type and not one I'd base future design upon.

jbcoe:
> This looks good - is there any coverage of a `protocol_view` from non-`const` `protocol&&` or `T&&`? Is that meant to be allowed?
> 
> EDIT: I see now, it just wasn't touched in the diff. LGTM!

Always good to check! Thanks.

philipcraig (commented):
Implicit is right for `protocol_view`: it's a non-owning parameter type, and the precedents line up — C++26 `std::function_ref` converts implicitly, `std::span` from an lvalue range, while `std::polymorphic` (allocating, like `protocol`) stays explicit. Keeping `protocol`'s constructors explicit matches P3019.

Deleting the rvalue overloads is the `span` choice rather than the `string_view`/`function_ref` one: it also rejects the safe same-expression call `read(Conforming{5})`, where the temporary outlives the call. I think that's the right trade — naming the object is a one-line fix and a dangling view is a miserable bug. Worth a sentence in DRAFT.md recording that the divergence from `function_ref` is deliberate, plus the two traps implicitness keeps: a view of a soon-to-die lvalue is now invisible at the call site, and overloading on `protocol_view<I>`/`protocol_view<const I>` is ambiguous for a non-const conforming lvalue.

Two asks:
- The from-protocol constructors now convert a moved-from `protocol` with no visible syntax, and the method_thunk assert is gated on `is_protocol_v`, so it's compiled out for views — a member call through such a view invokes a null function pointer even in debug builds. `assert(!p.valueless_after_move())` in both would catch it (#208 territory).
- DRAFT.md's synopsis lists the deleted `protocol_view(const T&&)` on the mutable view too; adding it keeps synopsis and implementation aligned and gives the same "would dangle" diagnostic the const view now has.

jbcoe:
DRAFT is due a complete rewrite. We'll get onto this very soon.

jbcoe:
Improved handling for valueless views is out of scope for this PR. It should be fixed though.

PR #274, opened 2026-08-31
Add deducing this support

### LINKED ISSUE #164: Consider handling of deducing-this in interface conformance

RyanJK5:
Closes #164


RyanJK5:
NOTE: This does not add support for a templated deducing-this function, for the same reasons we don't currently match templated member functions.

protocol.hh
```diff
@@ -129,6 +129,33 @@ consteval bool same_signature_ignoring_const(std::meta::info candidate,
...
+consteval bool is_explicit_object_function(std::meta::info candidate) {
+  const auto params = parameters_of(candidate);
+  return !params.empty() && is_explicit_object_parameter(params.front());
+}
+
+// Returns `true` if the member functions `candidate` and `interface` have the
+// same name, de-aliased return type and parameter types, and const / reference
+// qualification. `candidate` is assumed to be a function with an explicit
+// object parameter
+consteval bool same_name_and_params_with_explicit_object(
```
jbcoe: I would prefer to have the check for an explicit this folded into this function. It makes chaining functions correctly unnecessary.

jbcoe (changes requested):
Thanks for this.

One request for a mostly cosmetic change.

jbcoe (approved):
Thanks for this.

PR #275, opened 2026-08-31
Add interface constraints

### LINKED ISSUE #162: Consider handling of volatile for interface conformance

RyanJK5:
Adds a constraint to `protocol` and `protocol_view` on the types of allowed interfaces. In particular, we disallow interfaces with `volatile` member functions.

Closes #162


jbcoe (approved):
Thanks for this.

PR #276, opened 2026-09-01
Exclude types with base classes from the maybe-lambda fallback

philipcraig:
clang-tidy fails on `main` since #244: its view-constructor constraint evaluates `is_protocol_conformant_v<I, protocol<I>>`, and when `I` has a single call operator, `protocol<I>` matches `is_maybe_lambda` (a class template specialization has no identifier), so the lambda fallback reflects `protocol`'s wrapper-base `operator()` — a using-declarator reflection clang-p2996 rejects where GCC trunk accepts it. The rejection looks like rejects-valid and is reported upstream as bloomberg/clang-p2996#342, tracked here by #277.

Closure types have no base classes, so requiring an empty `bases_of` keeps lambdas matched and excludes `protocol` and `protocol_view` from the fallback. No conformance results change, and the exclusion remains a correct refinement of the concept even once the upstream fix lands.

### DIFF
--- a/protocol.hh
+++ b/protocol.hh
@@ -79,19 +79,18 @@ namespace detail {
 
 // Per ISO C++ ([expr.prim.lambda.closure]), closure types are unique, unnamed,
 // non-union class types.
+// The closure type is not an aggregate type.
 // This concept will also match an unnamed class type with a single
 // `operator()`.
-// TODO(jbcoe): Refine this concept to match only lambdas.
 //
-// Base classes are excluded: a closure type has none, and a class template
-// specialization also has no identifier, so without this `protocol<I>` and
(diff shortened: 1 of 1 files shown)


protocol.hh
```diff
@@ -82,9 +82,17 @@ namespace detail {
 // This concept will also match an unnamed class type with a single
 // `operator()`.
 // TODO(jbcoe): Refine this concept to match only lambdas.
+//
+// Base classes are excluded: a closure type has none, and a class template
```
jbcoe: I think that this comment can be made a LOT shorter.

jbcoe (approved):
Thanks for this. A clear improvement.

Issue #193, opened 2026-08-28
Reflection protocol: constrain constructors so a T less copyable/movable than I is a constraint failure, not a hard error


philipcraig:
Constructing a `protocol<I>` from a `T` whose special member functions are more restrictive than `I`'s produces a hard error deep inside `allocator_traits::construct` rather than a constraint failure.

`vtable_for<T>` (`reflection/protocol.hh:314-321`) instantiates both the `.copy` and `.move` lambdas whenever `protocol` is constructed from `T`. The `.copy` lambda is guarded by `if constexpr (is_copy_constructible_v<I>)`, but nothing guards against `T` being less copyable/movable than `I`:

| `I` | `T` | Result |
|---|---|---|
| copyable | non-copyable | hard error in `.copy` (`:314`) |
| movable, non-copyable | immovable | hard error in `.move` (`:321`) |

Both reproduce on `main` with `std::allocator`, so this is independent of the allocator question in #187.

```c++
struct Interface { void foo(); };
struct NonCopyable { NonCopyable(const NonCopyable&) = delete; NonCopyable(NonCopyable&&) = default; void foo() {} };
xyz::reflection::protocol<Interface> p{std::allocator_arg, std::allocator<std::byte>{}, std::in_place_type<NonCopyable>};
// alloc_traits.h: error: no matching function for call to 'construct_at(NonCopyable*&, const NonCopyable&)'
```

P3019 puts `is_copy_constructible_v<U>` in the *Constraints* of every `polymorphic` constructor that takes a `U`. `protocol` deliberately admits non-copyable interfaces, so the generalised form is: `T` must support every special member that `I` does.

Proposed: add to every constructor taking a `T` (from-value, in-place, and initializer-list, with and without allocator):

```c++
requires (!std::is_copy_constructible_v<I> || std::is_copy_constructible_v<T>) &&
         (!std::is_move_constructible_v<I> || std::is_move_constructible_v<T>)
```

and guard the `.move` lambda with `if constexpr (is_move_constructible_v<I>)` to mirror `.copy`. The design of which operations `protocol` itself offers when `I` is immovable is #187; this issue is only about turning the hard error into a constraint so that overload resolution and `static_assert(!std::is_constructible_v<...>)` tests work.

Related paper change: `DRAFT.md` should state this rule explicitly (#194).

PR #280, opened 2026-09-01
Let interfaces control movability

### LINKED ISSUE #187: Revisit move construction design
The current reflection implementation of `protocol` assumes that any type coming in is movable. This is OK if a custom allocator is not supplied, but if one is, it's possible that this will result in unexpected compilation failure. For example:

```c++
// Some allocator where is_always_equal = false_type
struct MyAlloc { ... };

struct Interface {
    void foo();
};

struct MyType {
    MyType(MyType&&) = delete;
    MyType(const MyType&) = delete;
    MyType& operator=(MyType&&) = delete;
    MyType& operator=(const MyType&) = delete;
    ~MyType() = delete;
    void foo();
};

static_assert(is_protocol_conformant_v<Interface, MyType>);

// Likely ERROR from instantiating .move in the vtable!
protocol<Interface, MyAlloc> p{MyType{}};
```

The two approaches to handle this I can think of are: allow interfaces to explicitly control their move construction requirements, OR require that an immovable type only be used with an `always_equal` allocator.

RyanJK5:
Left in draft because this poses some important design questions.

Closes #187 and #193.


RyanJK5:
We could allow move construction unconditionally and just prevent allocator-aware move construction and move assignment. Do we want to do this?

jbcoe:
Polymorphic (owning) is unconditionally move constructible, I think we need to do the same for protocol for consistency with other standard library types.

RyanJK5:
> Polymorphic (owning) is unconditionally move constructible, I think we need to do the same for protocol for consistency with other standard library types.

Are you leaning more towards gating movability requirements based on the allocator than the interface?

An alternative approach we could take is allow unconditional move construction and assignment, UNLESS the user-provided allocator explicitly needs to know how to generate the move() function (i.e. POCMA or always_equal is false)

jbcoe:
Any U in polymorphic<T> must be copyable. We don't have this constraint in protocol so have more freedom than I'd initially thought. 

We should enumerate and evaluate our options.

RyanJK5:
> Any U in polymorphic<T> must be copyable. We don't have this constraint in protocol so have more freedom than I'd initially thought. 
> 
> We should enumerate and evaluate our options.

Happy to discuss further. As I see it, the choices are to allow the interface to control this capability, or to make it dependent on the allocator.

jbcoe:
Do we want to flesh these options out in the comments and show code that would work and not work under the various options? GitHub markdown is pretty powerful.

RyanJK5:
> Do we want to flesh these options out in the comments and show code that would work and not work under the various options? GitHub markdown is pretty powerful.

One approach is to let the interface control movability, like this PR adds:

```c++
struct Movable {};

protocol<Movable> p(SomeCopyableType{});
auto p2 = std::move(p); // moving allowed
auto p3 = p2; // copying allowed
protocol<Movable> p4(SomeMoveOnlyType{}); // ERROR

struct Immovable {
    Immovable(const Immovable&) = delete;
    Immovable(Immovable&&) = delete;
};

protocol<Immovable> p4(SomeMoveOnlyType{}); // OK
protocol p5(std::move(p4)); // ERROR (moving)
auto p6 = p5; // ERROR (copying)
p4 = SomeCopyableType{}; // OK
```

We can also recognize, however, that the only time we need to actually call the move constructor of the object is when we dispatch to the `move()` function in the vtable. This only occurs when we need to reconstruct the object from an unequal allocator, so we can guarantee the object won't be moved in the same situations where we can guarantee move construction / assignment is `noexcept`.

So this alternative approach would mean:

- move constructor: available always.
- allocator-aware move construction: available if the interface is movable, OR if the allocator `is_always_equal`.
- move assignment: available if the interface is movable, OR if the allocator `is_always_equal` OR if the allocator has `pocma`.

I think we still need the interface to constrain movability in this case, as that is the user opting in to say "I care about allocator-aware move construction and move assignment" and therefore requiring movability from a conforming `T` in the case where the allocator is non-standard.

An example:

```c++
// pretend there's allocator types PocmaFalseAlwaysEq and PocmaTrueNotAlwaysEq

protocol<Movable> p(SomeMoveOnlyType{}); // No ERROR, because std::allocator is_always_equal

protocol<Movable, PocmaFalseAlwaysEq> p1(std::allocator_arg, /*whatever*/, SomeMoveOnlyType{}); // No ERROR, because always equal

protocol<Movable, PocmaTrueNotAlwaysEq> p2(std::allocator_arg, /*whatever*/, SomeMoveOnlyType{}); // ERROR, because allocator-aware move construction would not work.

protocol<Immovable> p3(SomeMoveOnlyType{}); // OK
protocol p4(std::move(p3)); // OK
p3 = std::move(p4); // OK

protocol<Immovable, PocmaTrueNotAlwaysEq> p5(SomeMoveOnlyType{}); // OK
protocol p6(std::move(p5)); // OK
p5 = std::move(p6); // OK because pocma is true.
protocol p7(std::allocator_arg, /* a new allocator */, std::move(p5)); // ERROR because always_equal is false

// If we had PocmaFalseNotAlwaysEq, both move assignment AND alloc-aware move construction would be disabled.
```

I don't think these are necessarily the only methods, but these are the ones I thought of.

RyanJK5:
This also has some implications for SBO. If the interface requires movability, the SBO can only be used if the type is movable. If the interface does not, SBO can be used for immovable types.

jbcoe:
I think that protocol should be movable when the type is movable and copyable when the type is copyable.
This is simple to reason about.

Allocators will complicate things (as you mentioned) as they may require copies which in some cases can't be performed. We can add requirements to the relevant constructors to prevent move construction when we cannot prove that no copy of a copyable type might need to take place.

I wonder if we need to go this far though? `vector<NonCopyable>` fails in compilation when copies are made but is copy-constructible according to traits.

jbcoe (approved):
This looks good. We might need a few more tests around allocators but this is a solid foundation.

RyanJK5:
> I think that protocol should be movable when the type is movable and copyable when the type is copyable. This is simple to reason about.
> 
> Allocators will complicate things (as you mentioned) as they may require copies which in some cases can't be performed. We can add requirements to the relevant constructors to prevent move construction when we cannot prove that no copy of a copyable type might need to take place.
> 
> I wonder if we need to go this far though? `vector<NonCopyable>` fails in compilation when copies are made but is copy-constructible according to traits.

Understood. If I understand correctly, then, I believe that approach maps onto what this PR adds.

jbcoe:
I’m happy to merge this once clang-tidy runs without issues.

jbcoe:
I pushed a fix for the clang-tidy issue (missing rule-of-five destructor).

RyanJK5:
> I pushed a fix for the clang-tidy issue (missing rule-of-five destructor).

Thanks!

jbcoe:
I'm happy to merge this (or for you to merge this) if you mark it as non-draft.

PR #288, opened 2026-09-02
Mark Bazel tests as small

philipcraig:
Every test finishes in under a second; the default medium size triggered Bazel's oversized-test warning.


jbcoe (approved):
Thanks for this.

PR #282, opened 2026-09-01
Extend operator support

jbcoe:

PR #289, opened 2026-09-02
Wait for both coverage uploads before notifying

philipcraig:
Codecov posts the project status after the first upload, so the runtime-only report is compared against a base that also carries the consteval flag and fails the 1% threshold (seen on #282). Set `after_n_builds: 2` so the comparison waits for both uploads.


jbcoe (approved):
Thanks for this. The previous setup did confuse me a bit.

PR #290, opened 2026-09-02
Delete rvalue constructor on protocol_view<T>

### LINKED ISSUE #285: protocol_view<I> lacks the deleted rvalue constructor DRAFT.md's synopsis specifies
DRAFT.md's synopsis declares a deleted rvalue constructor on both
`protocol_view` specializations:

```c++
// Construction from an rvalue conforming object is deleted.
template <typename T>
protocol_view(const T&&) = delete;
```

The implementation has it only on `protocol_view<const T>` (`protocol.hh`,
"A view of a temporary would dangle"). `protocol_view<T>` relies on its `U&`
constructor failing to bind an rvalue.

Both views reject a temporary, but the diagnostics differ. With GCC trunk:

```c++
struct Shape { double area() const; };
struct Circle { double area() const { return 1; } };
void f(protocol_view<Shape>);
void g(protocol_view<const Shape>);

g(Circle{});
// error: use of deleted function 'protocol_view<const T>::protocol_view(const U&&) [with U = Circle; T = Shape]'
// note: declared here  (points at the "would dangle" comment)

f(Circle{});
// error: could not convert 'Circle()' from 'Circle' to 'protocol_view<Shape>'

protocol_view<Shape> v(Circle{});
// error: no matching function for call to 'protocol_view<Shape>::protocol_view(Circle)'
// note: cannot bind non-const lvalue reference of type 'Circle&' to an rvalue of type 'Circle'
```

The const view names the deleted constructor and its comment; the mutable view
reports a binding failure with no pointer to the lifetime reason.

Fix: add the deleted overload to `protocol_view<T>` with the same constraints
as its `U&` constructor, and a `!is_constructible_v<protocol_view<Interface>,
Conforming>` static_assert alongside the existing lvalue ones. If the
synopsis is meant to differ, drop the declaration from `protocol_view<I>` in
the DRAFT.md rewrite instead.

Raised in the review of #272; deferred there as a documentation matter.

philipcraig:
DRAFT.md's synopsis declares a deleted rvalue constructor on both `protocol_view` specialisations, but the implementation had it only on `protocol_view<const T>`. `protocol_view<T>` relied on its `U&` constructor failing to bind, so viewing a temporary produced a different diagnostic for the mutable and const views.

Adds the deleted overload with the same constraints as the `U&` constructor, so both views report use of a deleted function pointing at the "would dangle" comment.

Closes #285

### DIFF
--- a/protocol.hh
+++ b/protocol.hh
@@ -1061,6 +1061,13 @@ class protocol_view
         vtable_(
             &detail::view_vtable_for<T, U, detail::const_policy::all_const>) {}
 
+  // A view of a temporary would dangle.
+  template <typename U>
+    requires is_protocol_conformant_v<T, std::remove_cvref_t<U>> &&
+                 (!is_protocol_view_v<std::remove_cvref_t<U>>) &&
+                 (!std::is_const_v<U>)
+  protocol_view(const U&&) = delete;
+
   // Views the object a `protocol<T>` owns, sharing its vtable.
   //
   // Precondition: `p` is not valueless.


jbcoe (approved):
Thanks for this.

PR #291, opened 2026-09-02
Prefer Bazel for building and testing

philipcraig:
Bazel builds this project faster than CMake, both from scratch and after
editing `protocol.hh`. Measured on a two-core codespace from fresh
directories, with Bazel given its own output root and repository cache so
nothing was reused:

| Scenario | CMake | Bazel |
|---|---|---|
| Full build | 223 s | 140 s |
| One-character comment change in `protocol.hh` | 81 s | 45 s |

Both systems rebuild the same three test translation units and relink after
the comment change.

Reword the agent instructions (`GEMINI.md`, which `CLAUDE.md` and
`Agents.md` link to), the developer guide and the sandbox hint to prefer
`scripts/bazel.sh`. CMake remains the path for sanitizers, coverage and
clang-tidy, and is still required when CMake files or `scripts/cmake.py`
change.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,19 +6,35 @@ This document explains how to set up, build, and understand the internals of the
 
 ### Prerequisites
 
-Before building, ensure you have [CMake](https://cmake.org/download/) 3.25 or
-later, a GCC with C++26 reflection (P2996) support, and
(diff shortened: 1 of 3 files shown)


jbcoe (approved):
Nicely spotted!

PR #293, opened 2026-09-02
Use Bazel for the sanitizer CI jobs

### LINKED ISSUE #292: Use bazel for sanitizer runs as the build is faster
CMake builds are slower than equivalent bazel builds. Using bazel for sanitizers builds will give us faster turnaround on CI.

philipcraig:
Closes #292.

The sanitizer CI jobs now build and test with Bazel instead of CMake. The branch first ran both side by side to compare them; the findings are below.

- `.bazelrc` gains `asan`, `ubsan`, and `tsan` configs with the same flags `cmake/sanitizers.cmake` applies. They use `--copt` so googletest is instrumented too, and `--compilation_mode=dbg` so reports carry symbolised stack traces.
- `scripts/bazel.sh` takes `--asan`, `--ubsan`, and `--tsan`, mirroring `scripts/cmake.sh`; `--asan` with `--tsan` is rejected. The CMake flags stay for local use.
- `sanitizers.yml` sets up Bazel with a per-sanitizer disk cache and runs `./scripts/bazel.sh --asan --ubsan` and `./scripts/bazel.sh --tsan`. Its paths filters track the Bazel inputs instead of the CMake ones.
- The change-detection skip moves from job level to step level in `cmake.yml`, `bazel.yml`, and `sanitizers.yml`, so skipped matrix jobs no longer report under the unexpanded name `${{ matrix.settings.name }} ${{ matrix.configuration }}`.
- `.bazelignore` also ignores `.claude`, where Claude Code keeps agent worktrees whose `build/` directories otherwise make `bazel test //...` fail to load.
- Docs updated.

The `asan`, `tsan`, and `sanitizers` check names are unchanged, so the required-check ruleset needs no change.

### Findings from four runs with both build systems

Build-and-test step, seconds:

| Run | cmake asan | cmake tsan | bazel asan | bazel tsan |
|---|---|---|---|---|
| Cold cache | 85 | 78 | 91 | 92 |
| Warm rerun, no change | 88 | 78 | 21 | 16 |
| Push after main changed `protocol.hh` | 88 | 78 | 61 | 64 |
| After rebase | 89 | 68 | 77 | 65 |

- Bazel is never slower than CMake and drops to about 20 s when nothing changed. Whole-job times are equal within noise because the "Install GCC trunk" step (81 to 149 s, mostly the package's `mkheaders` postinst) dominates both; #297 tracks caching it.
- The setup-bazel disk cache saves only on an exact key miss, so it froze at the first run's contents and later runs stayed partially warm; #298 tracks saving it per run.

### DIFF
--- a/.bazelignore
+++ b/.bazelignore
@@ -1,2 +1,3 @@
 build
 _deps
+.claude
--- a/.bazelrc
+++ b/.bazelrc
@@ -19,3 +19,22 @@ build --cxxopt=-Wno-self-move
 
 # Show test output on failure
 test --test_output=errors
+
+# Sanitizer builds, matching the flags cmake/sanitizers.cmake applies. Select
+# one with `--config=asan`, `--config=ubsan`, or `--config=tsan`
+# (`scripts/bazel.sh --asan/--ubsan/--tsan`). `--copt` rather than `--cxxopt`
+# so external dependencies (googletest) are instrumented too, and `dbg` so
+# sanitizer reports carry symbolised stack traces. ASAN and TSAN cannot be
+# combined; ASAN and UBSAN can.
+build:asan --compilation_mode=dbg
+build:asan --copt=-fsanitize=address
+build:asan --copt=-fno-omit-frame-pointer
+build:asan --linkopt=-fsanitize=address
+
+build:ubsan --compilation_mode=dbg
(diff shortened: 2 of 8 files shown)


jbcoe (approved):
Thanks for this. Everything will be a little bit faster to do now.

PR #296, opened 2026-09-02
Cosmetic cleanup and comment fix

jbcoe:


protocol.hh
```diff
@@ -183,7 +183,7 @@ consteval bool member_function_conforms_to(std::meta::info candidate,
     // No additional conditions.
   } else {
     if (!same_signature_ignoring_const(candidate, interface)) return false;
-    // If interface is `const`, `candidate` must be const.
+    // `const` qualifiers must match.
```
jbcoe, line 186: This caused me some confusion and I'm not 100% sure it's what we want. For now, the comment and code are consistent.

protocol.hh
```diff
@@ -1164,7 +1165,7 @@ class protocol_view<const T> : public detail::protocol_wrappers_t<
   // Non-owning pointer to the viewed object.
   const void* object_ = nullptr;
 
-  const typename detail::vtable_generator<T>::vtable* vtable_;
```
jbcoe, line 1167: `typename` is not needed as a type must appear here.

protocol.hh
RyanJK5: I think a `!interface_is_const || candidate_is_const` relationship makes more sense here, but that's out of scope for this PR. the comment makes it clearer.

RyanJK5 (approved):

PR #299, opened 2026-09-03
Cache the installed GCC trunk tree between CI runs

### LINKED ISSUE #297: Cache the installed GCC trunk tree in CI
Every GCC trunk job spends 81 to 149 s in its "Install GCC trunk" step, more than the build it then runs. Breakdown from PR #293's runs: wget of the 460 MB .deb 18 to 22 s, `apt-get update` 7 to 11 s, dpkg unpack 12 to 21 s, and the package's postinst running `mkheaders` 38 to 91 s.

Caching the .deb would save under 20 s. Instead, cache the installed `/opt/gcc-latest` tree after a successful install, keyed by the versioned file name the unversioned URL redirects to (currently `gcc-latest_17.0.0-20260816git8df013222dff.deb`), and fall back to the download and apt install on a miss. Expected cost afterwards: roughly 20 to 30 s per trunk job, across the five workflows that install it.

`mkheaders` must not be skipped: the .deb ships the build host's fixed `bits/floatn.h`, which fixincludes discards on Ubuntu 26.04, and it generates `include-fixed/limits.h`, which the package does not contain.

A composite action under `.github/actions/` would replace the five copies of the step that the "Keep in sync" comments point at.

philipcraig:
Closes #297.

Every GCC trunk job spent 81 to 149 s installing the compiler, more than the build it then ran, and most of that was the package's `mkheaders` postinst regenerating fixed headers.

- `.github/actions/install-gcc-trunk` resolves the versioned file the unversioned URL redirects to, restores `/opt/gcc-latest` from the Actions cache under a key made of that file name and its `Last-Modified` date, and falls back to the download and apt install on a miss. If the server ever stops providing either identifier, it installs without caching rather than key blindly.
- The five copies of the install step in `cmake.yml`, `bazel.yml`, `sanitizers.yml`, and `coverage.yml` now use the action, and the action file is in each workflow's paths filter. Pinning a snapshot after an upstream regression is the action's `package-url` input.

### DIFF
--- a/.github/actions/install-gcc-trunk/action.yml
+++ b/.github/actions/install-gcc-trunk/action.yml
@@ -1,0 +1,72 @@
+name: Install GCC trunk
+description: >-
+  Install the prebuilt GCC trunk snapshot from
+  https://jwakely.github.io/pkg-gcc-latest/ into /opt/gcc-latest, restoring
+  the installed tree from the Actions cache while the snapshot is unchanged.
+  scripts/cmake.py and scripts/bazel.py find /opt/gcc-latest themselves, so
+  callers need no PATH changes.
+
+inputs:
+  package-url:
+    description: >-
+      URL of the snapshot .deb. The default tracks upstream weekly; if a
+      snapshot regression breaks a job, pin the versioned file linked from
+      https://jwakely.github.io/pkg-gcc-latest/, e.g.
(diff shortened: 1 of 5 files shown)


jbcoe (approved):
Thanks for this. This should speed up CI.

PR #301, opened 2026-09-05
Persist the agentic-sandbox uv cache between runs

jbcoe:
Previously the agentic-sandbox downloaded uv packages on every run. 

A named volume now stores uv's download cache and is used by the agentic-sandbox.

Running `scripts/agentic-sandbox.py --no-cache-volumes` will disable persistent cache.

Disable dockerfile linter pending replacement as it does not support the extended Dockerfile.


jbcoe:
This PR needs manual testing
- [ ] Windows
- [x] Mac
- [ ] Linux / WSL

RyanJK5:
I'm seeing this, but I'm certain I'm doing something wrong. What steps did you follow to test this?

image

```
vscode@0582cb25bca4:/workspace$ ./scripts/bazel.sh
error: Failed to initialize cache at `/home/vscode/.cache/uv`
  Caused by: failed to open file `/home/vscode/.cache/uv/CACHEDIR.TAG`: Permission denied (os error 13)
```

EDIT: Was able to get it working, but I have to type `export UV_CACHE_DIR=some/directory/` every time I launch the sandbox.

jbcoe:
@RyanJK5 you're very likely doing nothing wrong. This should just work. 

I'll make the cache macOS only as it's the only thing I can directly check. I'll split the PR up.

philipcraig:
@RyanJK5 the permission error is the volume, not the image: before f658e29 the image had no `vscode`-owned mount point, so Docker created `cc-protocol-uv-cache` root-owned on your first run, and a volume keeps the ownership it was created with. The fix only applies to a volume created after it. Try

```
docker volume rm cc-protocol-uv-cache
./scripts/agentic-sandbox.py --rebuild-docker --cache-volumes
```

and you should no longer need `UV_CACHE_DIR`.

@jbcoe with that, the cache volume worked first time for me on Ubuntu under WSL2 (cold run 1m20s, warm run 2.9s), so the default could be on for Linux too rather than macOS-only.

jbcoe:
I've modified the PR and description to enable cache volumes by default for all platforms.

RyanJK5:
> @RyanJK5 the permission error is the volume, not the image: before [f658e29](https://github.com/jbcoe/cc-protocol/commit/f658e29f05b73164e60012fb261a9a4f72b57afa) the image had no `vscode`-owned mount point, so Docker created `cc-protocol-uv-cache` root-owned on your first run, and a volume keeps the ownership it was created with. The fix only applies to a volume created after it. Try
> 
> ```
> docker volume rm cc-protocol-uv-cache
> ./scripts/agentic-sandbox.py --rebuild-docker --cache-volumes
> ```
> 
> and you should no longer need `UV_CACHE_DIR`.
> 
> @jbcoe with that, the cache volume worked first time for me on Ubuntu under WSL2 (cold run 1m20s, warm run 2.9s), so the default could be on for Linux too rather than macOS-only.

Thanks. It looks to be working now for me - it is still supposed to reinstall after exiting and re-entering the sandbox, right?

scripts/agentic-sandbox.py
```diff
@@ -147,6 +157,8 @@ def log(msg: str) -> None:
             [
                 "docker",
                 "build",
+                "--target",
+                "sandbox",
```
philipcraig, line 161: The image check above keys only on the tag, so anyone who built the sandbox before this PR keeps the old image: `docker image inspect` succeeds, no rebuild happens, and the volume is mounted onto `/home/vscode/.cache/uv`, which that image never created. Docker creates the mount point and the volume root as `root:root`, and the first `uv run` fails with permission denied. Nothing warns that the image is stale and `--rebuild-docker` is opt-in.

Suggest always running `docker build --target sandbox` (the layer cache makes it near-free), or keying the tag on a hash of the Dockerfile.

.devcontainer/devcontainer.json
```diff
@@ -2,7 +2,8 @@
 	"name": "xyz.cc-protocol",
 	"build": {
 		"dockerfile": "../docker/Dockerfile",
-		"context": ".."
+		"context": "..",
+		"target": "base"
```
philipcraig, line 6: `target: base` also drops `npm install -g @google/gemini-cli @anthropic-ai/claude-code`, the Antigravity installer, `NPM_CONFIG_PREFIX` and the `.npm-global/bin` PATH entry, all of which now live below `FROM base AS sandbox`. After a rebuild `which claude gemini agy` returns nothing, and without `NPM_CONFIG_PREFIX` a manual `npm install -g` as `vscode` fails with EACCES because the prefix is `/usr`. The VS Code extension keeps working from its bundled binary, but the terminal CLIs the `~/.claude` mount comment below refers to are gone.

Either keep the devcontainer on the sandbox stage, or move the CLI install into `base` and say in the PR description that the devcontainer loses them.

docker/Dockerfile
```diff
@@ -76,44 +81,49 @@ RUN ARCH=$(dpkg --print-architecture) && \
...
-ENV UV_PROJECT_ENVIRONMENT="/home/vscode/.venv"
-
 # Agent config dirs (~/.gemini / ~/.claude) are mounted from the host at runtime
 # to persist auth state between sessions. Default config files are seeded by the launch
 # script before the mount, so image-side preconfiguration is not needed here.
+
+# Pre-create the cache dirs owned by vscode: Docker otherwise creates a
+# volume's mount point as root on first use, and these tools cannot write
+# into a root-owned directory.
+RUN mkdir -p /home/vscode/.cache/uv
```
philipcraig: This is the third copy of the path, after `ENV UV_CACHE_DIR` at line 107 and `CACHE_VOLUMES` in `agentic-sandbox.py`, and it sits one stage away from the ENV it has to match. If `UV_CACHE_DIR` moves, the mkdir stays behind and the volume root goes back to being root-owned with nothing to catch it.

`RUN mkdir -p "$UV_CACHE_DIR"` directly under the ENV in `base` (USER is already `vscode` there) removes the drift and is harmless for the devcontainer.

docker/Dockerfile
```diff
@@ -76,44 +81,49 @@ RUN ARCH=$(dpkg --print-architecture) && \
...
-# Configure npm for user-local global installs.
+ENV PATH="/home/vscode/.local/bin:/usr/local/bin:${PATH}"
+
+# Set up uv and the Python environment. UV_CACHE_DIR names the download cache;
+# in the devcontainer it lives in the container's writable layer, and the
+# sandbox stage mounts a persistent volume there instead.
+RUN curl -LsSf https://astral.sh/uv/install.sh | sh
+ENV UV_PROJECT_ENVIRONMENT="/home/vscode/.venv"
+ENV UV_CACHE_DIR="/home/vscode/.cache/uv"
+ENV UV_LINK_MODE="copy"
```
philipcraig: Only the sandbox needs `UV_LINK_MODE=copy`: its cache is a volume on a different filesystem from the venv. In the devcontainer `~/.venv` and `~/.cache/uv` share the overlayfs, so hardlinks work and this forces uv to copy every package on each sync instead. uv already falls back clone -> hardlink -> copy on its own, so the sandbox only needs this to silence the warning.

Move it below `FROM base AS sandbox`, or pass `-e UV_LINK_MODE=copy` next to the volume mount in `agentic-sandbox.py`.

docker/.dockerfilelintrc
```diff
@@ -0,0 +1,5 @@
+# dockerfilelint's missing_tag check predates multi-stage builds: it reads the
+# `FROM base` in the sandbox stage as an untagged external image and flags it.
+# The real base image is pinned to a tag at the top of Dockerfile, so disable it.
+rules:
+  missing_tag: off
```
philipcraig: This turns the untagged-FROM check off for the whole file, not just the `FROM base` line: an appended `FROM node` stage passes `pre-commit run dockerfilelint` with this rc in place. dockerfilelint has no per-line suppression, so within that tool this is the only lever; hadolint understands multi-stage builds and has a pre-commit hook, and would be the proper fix. If the rule stays off, worth saying so in the comment.

scripts/agentic-sandbox.py
```diff
@@ -119,6 +123,12 @@ def main() -> None:
     parser.add_argument(
         "--rebuild-docker", action="store_true", help="Rebuild the Docker image."
     )
+    parser.add_argument(
+        "--cache-volumes",
```
philipcraig, line 127: The `agentic-sandbox.sh` Options table in CONTRIBUTING.md needs `--cache-volumes` / `--no-cache-volumes`, and something should name the `cc-protocol-uv-cache` volume and how to reset it (`docker volume rm cc-protocol-uv-cache`). The volume grows without bound since uv never prunes it on its own.

docker/Dockerfile
```diff
@@ -76,44 +81,49 @@ RUN ARCH=$(dpkg --print-architecture) && \
...
-# always owned by UID 1000, and VS Code's `updateRemoteUserUID` cannot remap
-# `vscode` onto an ID that is already taken. Remove the stock user first so
-# `vscode` gets 1000:1000 and the mounted files are readable and writable.
+# Install Node.js. The devcontainer's Claude Code VS Code extension needs it,
+# and the sandbox stage installs the agent CLIs on top of it.
+RUN curl -fsSL https://deb.nodesource.com/setup_20.x | bash - \
+    && apt-get install -y --no-install-recommends nodejs \
+    && rm -rf /var/lib/apt/lists/*
+
+# Create the non-root user `vscode` at UID 1000. Ubuntu ships a stock `ubuntu`
```
philipcraig: The rationale for UID 1000 was dropped from here and from devcontainer.json: the host bind mounts (`~/.ssh`, `~/.config/gh`, `~/.claude`) are UID 1000, and VS Code's `updateRemoteUserUID` cannot remap onto a UID that is already taken. Without it a later edit that bumps the base image or drops `--uid 1000` / `userdel` has no in-tree warning that every bind mount becomes unreadable. This text came from #247; suggest keeping one sentence of it.

scripts/agentic-sandbox.py
```diff
@@ -13,6 +13,10 @@
 
 IMAGE_NAME = "cc-protocol-sandbox"
 
+# Docker named volumes persisting each tool's cache across constainer instances.
```
philipcraig: Typo: `constainer`. Also "each tool's cache" here and "uv cache volumes" in the `--cache-volumes` help describe a dict with one entry; the npm and Bazel caches were deliberately dropped, so singular wording avoids implying they are persisted.

philipcraig (changes requested):
Two blocking items: a pre-PR `cc-protocol-sandbox` image is reused as-is and gets a root-owned uv cache, and the devcontainer's switch to `target: base` drops the `claude`/`gemini`/`agy` CLIs. The rest are smaller. Details inline.

docker/Dockerfile
jbcoe: Fixed.

scripts/agentic-sandbox.py
jbcoe: Rebuilding every time needs internet access which we need to be able to work without.

Given that, we have no automated way to rebuild the docker image as required, I'm happy with a forced rebuild.

philipcraig (commented):
The uv side does what the PR says: with a warm cache, `uv sync --offline` into a fresh venv takes 0.2 s and `uv run --offline` works, so once the volume is warm the sandbox's Python tooling needs no network. Both earlier blocking items are resolved: `FROM base AS sandbox` now sits below the CLI install, so `target: base` keeps `claude`/`gemini`/`agy`, and the mount point is vscode-owned.

The remaining points are all about the version tag, and they matter most for the offline case. Details inline; the short version is that a content hash of the Dockerfile stored as an image label, with a fixed image name and a fixed volume name, gives the forced rebuild you want without a manual bump, without pyproject/uv.lock churn, and with a working fallback when the rebuild cannot run.

scripts/agentic-sandbox.py
```diff
@@ -9,9 +9,23 @@
...
+import tomllib
 from typing import TypedDict
 
-IMAGE_NAME = "cc-protocol-sandbox"
+PROJECT_ROOT = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+PYPROJECT_PATH = os.path.join(PROJECT_ROOT, "pyproject.toml")
+with open(PYPROJECT_PATH, "rb") as f:
+    DEVTOOLS_VERSION = tomllib.load(f)["project"]["version"]
+
+IMAGE_NAME = f"cc-protocol-sandbox-{DEVTOOLS_VERSION}"
```
philipcraig: A versioned name makes the rebuild automatic, but the rebuild is the one step that needs the network. Pull a commit that bumps the version before a flight, and `docker image inspect` misses, `docker build` fails at the first `apt-get`, and the script exits: there is no `--image`, no `--no-build`, and the previous image is untagged from this name's point of view. The bump is also manual, so a Dockerfile change without a bump (#302, if it lands after this) reproduces the stale-image case this is meant to fix, and every bump drags `uv.lock` along.

Suggest hashing the inputs instead and keeping one name:

```python
DOCKERFILE_HASH = hashlib.sha256(
    Path(PROJECT_ROOT, "docker/Dockerfile").read_bytes()
).hexdigest()[:16]
# build: docker build --label cc-protocol.dockerfile=<hash> -t cc-protocol-sandbox ...
# check: docker image inspect -f '{{index .Config.Labels "cc-protocol.dockerfile"}}' cc-protocol-sandbox
```

Rebuild when the label differs or the image is missing. If the build fails and an image exists, warn that it is stale and run it anyway, which is exactly the offline behaviour. Re-tagging leaves the old image dangling for `docker image prune`, where versioned names accumulate several GB per bump with nothing removing them. Add `.bazelversion` to the hash once #302 lands, since it is a `COPY` input.

scripts/agentic-sandbox.py
```diff
@@ -9,9 +9,23 @@
...
+
+# Docker named volumes persisting each tool's cache across container instances.
+# The devcontainer is long-lived and does not use these, it keeps its cache locally.
+_UNVERSIONED_CACHE_VOLUMES: dict[str, str] = {"uv-cache": "/home/vscode/.cache/uv"}
+CACHE_VOLUMES = {
```
philipcraig: Versioning the volume discards the uv cache on every bump, so the first run after one needs the network to re-download every wheel; on a flight that run fails. The cache content does not depend on the image: uv versions its own cache layout (`wheels-v*`, `archive-v*`) and handles upgrades itself. The root-owned volume was created by pre-fix builds of this branch, so a single rename, or the `docker volume rm` already documented, covers it; a fixed name is the right steady state. Old `cc-protocol-uv-cache-<version>` volumes also pile up with nothing pruning them.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "cc-protocol-dev-tools"
-version = "0.1.0"
+version = "2026.9.7.1708"  # Used for agentic-sandbox Docker image name.
```
philipcraig: This turns the dev-tools package version into a Docker tag, so every image bump also rewrites `uv.lock` and trips the uv-lock CI check if forgotten. With the label approach above this line can go back to `0.1.0`.

.pre-commit-config.yaml
```diff
@@ -32,11 +32,12 @@ repos:
   hooks:
   - id: clang-format
     types_or: [c++, c]
-- repo: https://github.com/pryorda/dockerfilelint-precommit-hooks
-  rev: v0.1.0
-  hooks:
-  - id: dockerfilelint
-    stages: [pre-commit]
+# TODO(jbcoe): Use hadolint (or similar) for improved Dockerfile linting.
```
philipcraig, line 35: `https://github.com/AleksaC/hadolint-py` (hook id `hadolint`) installs the hadolint binary through pip, so it runs under pre-commit without Docker and understands multi-stage builds.

jbcoe:
I've pulled the versioning out. It's too complicated and I'd rather limit the scope of this PR. 

People can remove and rebuild containers as documented.

.pre-commit-config.yaml
jbcoe: We can handle this later.

pyproject.toml
jbcoe: Versioning is now dropped.

scripts/agentic-sandbox.py
jbcoe: Versioning is now dropped.

scripts/agentic-sandbox.py
jbcoe: Versioning is now dropped.

scripts/agentic-sandbox.py
```diff
@@ -13,6 +13,10 @@
 
 IMAGE_NAME = "cc-protocol-sandbox"
 
+# Docker named volumes persisting each tool's cache across container instances.
+# The devcontainer is long-lived and does not use these, it keeps its cache locally.
+CACHE_VOLUMES: dict[str, str] = {"cc-protocol-uv-cache": "/home/vscode/.cache/uv"}
```
philipcraig, line 18: A `cc-protocol-sandbox` image built before this PR passes the image-exists check below, and the first run then creates `cc-protocol-uv-cache` root-owned, which is what @RyanJK5 hit. Nothing in CONTRIBUTING says to run `--rebuild-docker` once after pulling this. If you want the script to catch it without network access: the new image is the one whose environment carries `UV_LINK_MODE=copy`, so `docker image inspect -f '{{.Config.Env}}' cc-protocol-sandbox` not containing it means the image predates this change, and a warning that names `--rebuild-docker` and `docker volume rm cc-protocol-uv-cache` is three lines. A sentence in CONTRIBUTING would also do.

philipcraig (approved):
With the versioning gone this is the 921c8ed change plus the fixes for both earlier blocking items: fixed image and volume names, `pyproject.toml` and `uv.lock` untouched, the agent CLIs stay in `base`, and the volume mount point is vscode-owned. The uv side works offline once warm (`uv sync --offline` into a fresh venv from a warm cache takes 0.2 s). Good to merge from my side; one optional note inline.

scripts/agentic-sandbox.py
jbcoe: We explain in `CONTRIBUTING ` how to remove caches and update stale images. I think that suffices.

PR #302, opened 2026-09-05
Pre-warm bazelisk's Bazel download in the Docker image

jbcoe:
Pin the Bazel version via .bazelversion and have the image fetch that version during build, so sandbox containers reuse the baked-in binary instead of downloading it on every run.

The docker image will (currently) need explicit rebuilding with `--rebuild-docker`. Modifying `agentic-sandbox.sh` to build by default (and pick up changes to bazel version) is deferred to a later PR.

This PR is part of a sequence of changes to enable a workflow that does not rely on internet access.


docker/Dockerfile
```diff
@@ -115,5 +115,11 @@ RUN curl -LsSf https://astral.sh/uv/install.sh | sh
...
-# to persist auth state between sessions. Default config files are seeded by the launch
-# script before the mount, so image-side preconfiguration is not needed here.
+# to persist auth state between sessions. Default config files are seeded by the
+# launch script before the mount, so image-side preconfiguration is not needed
+# here.
+
+# Pre-warm bazelisk's download of the pinned Bazel binary (~60MB) to avoid
+# fetching it on every run.
+COPY .bazelversion /workspace/.bazelversion
+RUN bazel version
```
philipcraig: At build time `/workspace` holds only `.bazelversion`, with no `MODULE.bazel` or `WORKSPACE` marker, so bazelisk has no workspace root and ignores the pin: it resolves "latest" and matches 9.2.0 only by coincidence today. A directory containing only `.bazelversion` set to `0.0.0-nonexistent` still prints `bazel 9.2.0` with this bazelisk; the same file beside a `MODULE.bazel` makes it try to download 0.0.0-nonexistent. Once 9.3.0 ships, any uncached image build bakes the wrong version and every sandbox run downloads 9.2.0 again.

```suggestion
RUN USE_BAZEL_VERSION="$(cat /workspace/.bazelversion)" bazel version
```

`USE_BAZEL_VERSION` is honoured without a workspace root and is scoped to this RUN. Copying `MODULE.bazel` alongside would also work.

docker/Dockerfile
```diff
@@ -115,5 +115,11 @@ RUN curl -LsSf https://astral.sh/uv/install.sh | sh
 ENV UV_PROJECT_ENVIRONMENT="/home/vscode/.venv"
 
 # Agent config dirs (~/.gemini / ~/.claude) are mounted from the host at runtime
-# to persist auth state between sessions. Default config files are seeded by the launch
-# script before the mount, so image-side preconfiguration is not needed here.
+# to persist auth state between sessions. Default config files are seeded by the
+# launch script before the mount, so image-side preconfiguration is not needed
+# here.
+
+# Pre-warm bazelisk's download of the pinned Bazel binary (~60MB) to avoid
```
philipcraig: `bazel version` also extracts the 188 MB install base, so this layer is about 250 MB rather than 60 MB. That is useful, since the install base is keyed by binary hash and skips "Extracting Bazel installation" on each run, but the comment should say so. If only the download was intended, `bazel --version` is client-only and leaves just the 63 MB archive.

docker/Dockerfile
```diff
@@ -115,5 +115,11 @@ RUN curl -LsSf https://astral.sh/uv/install.sh | sh
...
+# fetching it on every run.
+COPY .bazelversion /workspace/.bazelversion
```
philipcraig: `COPY` ignores the preceding `USER vscode`, so this leaves a root-owned file in the vscode-owned `/workspace`. `COPY --chown=vscode:vscode` avoids it.

philipcraig (commented):
Two points outside the diff:

The pre-warm goes stale silently on the first `.bazelversion` bump. `scripts/agentic-sandbox.py` only builds the image when `docker image inspect cc-protocol-sandbox` fails or `--rebuild-docker` is passed, the image carries no record of which Bazel it baked, and the runtime bind mount hides the baked copy. After a bump every existing image passes the check and each `--rm` run downloads Bazel again, with nothing pointing at `--rebuild-docker`. Cheapest fix is to run `docker build` unconditionally before `docker run`: the COPY and RUN are the last two layers, so it is a layer-cache no-op when nothing changed.

CONTRIBUTING.md still says Bazelisk "fetches a current Bazel release on first use". It now fetches the version pinned in `.bazelversion`, and nothing says how to bump it or that the sandbox image needs rebuilding afterwards.

jbcoe:
Rebuilding docker on every run will fail if there is no internet connection. We can consider running a build by default and disabling it with `--nobuild`. I'll tackle that in a later PR.

CONTRIBUTING.md is updated.

docker/Dockerfile
```diff
@@ -115,9 +115,19 @@ RUN curl -LsSf https://astral.sh/uv/install.sh | sh
...
 # Avoid dubious ownership issues flagged by recent versions of git by trusting `/workspace`.
 # https://git-scm.com/docs/git-config#Documentation/git-config.txt-safedirectory
 RUN git config --global --add safe.directory /workspace
+
+# Pre-warm bazelisk's download and install of the pinned Bazel binary to avoid
+# fetching and installing it on every run.
+# https://github.com/bazelbuild/bazelisk#how-does-bazelisk-know-which-bazel-version-to-run
+# We use .bazelversion to avoid hard-coding the bazel version number in multiple places.
+COPY .bazelversion .bazelversion
+RUN read -r USE_BAZEL_VERSION < .bazelversion \
```
philipcraig: `read -r` sets a shell variable, not an environment variable, so bazelisk never sees `USE_BAZEL_VERSION` and, with no `MODULE.bazel` beside it, ignores `.bazelversion` too. Checked with bazelisk 1.25.0 in a directory holding only `.bazelversion` set to `0.0.0-nonexistent`: this exact `read -r ... && bazel version` prints `Build label: 9.2.0` (it resolved "latest", with the "not invoked from within a workspace" warning), while the exported form tries to download `bazel-0.0.0-nonexistent-linux-x86_64`. Today the two agree by coincidence; the first 9.3.0 release makes an uncached build bake the wrong binary and every run download again.

```suggestion
RUN USE_BAZEL_VERSION="$(cat .bazelversion)" bazel version \
    && rm .bazelversion
```

philipcraig (commented):
The pin is still not reaching bazelisk (inline). Two points on the offline goal:

The bake only removes bazelisk's ~63 MB download. Every `--rm` sandbox run still starts with an empty Bazel output base, and the first build fetches `rules_cc`, `googletest` and their transitive modules (`bazel_features`, `platforms`, `rules_java`, abseil, re2, ...) from the BCR. On a fresh output base, `bazel build --nofetch //...` fails with `External repository @@bazel_features++version_extension+bazel_features_globals not found and fetching repositories is disabled`. So Bazel in the sandbox stays online-only until `~/.cache/bazel` is on a volume, as #304 does. Fine to land this first, but the description's "does not rely on internet access" only becomes true with that follow-up; worth one sentence in CONTRIBUTING next to the `.bazelversion` bump note.

This and #301 both append to the end of `docker/Dockerfile`, so whichever lands second needs a rebase. If #301 lands first, this Dockerfile change also needs a `pyproject.toml` version bump or every existing sandbox image stays stale, which is the case the version tag is meant to catch.

The re-indentation of the two `case` blocks and the reflowed comment are unrelated churn; dropping them keeps the diff to the pre-warm.

docker/Dockerfile
jbcoe: Fixed.

philipcraig (approved):

PR #304, opened 2026-09-06
Add bazel repository cache for agentic sandbox

jbcoe:
With a populated cache, bazel builds can complete in the agentic sandbox with no internet connection.

See: https://bazel.build/run/build#repository-cache

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -2,11 +2,6 @@
 build --cxxopt=-std=c++26
 build --cxxopt=-freflection
 
-# Cache downloaded external dependency archives in their own directory,
-# separate from the default output root (~/.cache/bazel), which also holds
-# build artifacts we don't want to persist across sandbox runs.
-common --repository_cache=~/.cache/bazel-repo
-
 # Treat external repositories' headers (e.g. googletest) as system headers,
 # matching CMake's SYSTEM FetchContent_Declare, so -Werror below only holds
 # our own code to account and not warnings from vendored dependencies.
--- a/.bazelrc
+++ b/.bazelrc
@@ -2,6 +2,11 @@
 build --cxxopt=-std=c++26
 build --cxxopt=-freflection
 
+# Cache downloaded external dependency archives in their own directory,
+# separate from the default output root (~/.cache/bazel), which also holds
+# build artifacts we don't want to persist across sandbox runs.
+common --repository_cache=~/.cache/bazel-repo
+
 # Treat external repositories' headers (e.g. googletest) as system headers,
 # matching CMake's SYSTEM FetchContent_Declare, so -Werror below only holds
 # our own code to account and not warnings from vendored dependencies.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -230,12 +230,14 @@ plain shell in the container.
 |------|-------------|
 | `--rebuild-docker` | Rebuild the Docker image before starting. |
 | `--update` | Update the agent CLI to the latest version before running. |
-| `--cache-volumes`, `--no-cache-volumes` | Mount the persistent `uv` cache volume (default: on). |
+| `--cache-volumes`, `--no-cache-volumes` | Mount the persistent cache volumes (default: on). |
 | `-v`, `--verbose` | Enable verbose logging. |
 
-The `uv` cache persists across sandbox runs in a Docker named volume,
-`cc-protocol-uv-cache`. The cache grows without bound since `uv` never prunes
-it; to reset it, run `docker volume rm cc-protocol-uv-cache`.
+The `uv` package cache, Bazel's downloaded dependency archives, and pre-commit's
+environment cache persist across sandbox runs in Docker named volumes,
+`cc-protocol-uv-cache`, `cc-protocol-bazel-cache`, and
+`cc-protocol-pre-commit-cache`. These grow without bound since none of the
+tools prune them; to reset one, run `docker volume rm <name>`.
 
 ### Using pre-commit Locally to run Github Workflow checks
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -235,7 +235,7 @@ plain shell in the container.
 
 The `uv` package cache, Bazel's downloaded dependency archives, and pre-commit's
 environment cache persist across sandbox runs in Docker named volumes,
-`cc-protocol-uv-cache`, `cc-protocol-bazel-cache`, and
+`cc-protocol-uv-cache`, `cc-protocol-bazel-repository-cache`, and
 `cc-protocol-pre-commit-cache`. These grow without bound since none of the
 tools prune them; to reset one, run `docker volume rm <name>`.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -233,11 +233,10 @@ plain shell in the container.
 | `--cache-volumes`, `--no-cache-volumes` | Mount the persistent cache volumes (default: on). |
 | `-v`, `--verbose` | Enable verbose logging. |
 
-The `uv` package cache, Bazel's downloaded dependency archives, and pre-commit's
-environment cache persist across sandbox runs in Docker named volumes,
-`cc-protocol-uv-cache`, `cc-protocol-bazel-repository-cache`, and
-`cc-protocol-pre-commit-cache`. These grow without bound since none of the
-tools prune them; to reset one, run `docker volume rm <name>`.
+The `uv` package cache and Bazel's downloaded dependency archives persist
+across sandbox runs in Docker named volumes, `cc-protocol-uv-cache` and
+`cc-protocol-bazel-repository-cache`. These grow without bound since neither
+tool prunes them; to reset one, run `docker volume rm <name>`.
 
 ### Using pre-commit Locally to run Github Workflow checks
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -233,10 +233,10 @@ plain shell in the container.
 | `--cache-volumes`, `--no-cache-volumes` | Mount the persistent cache volumes (default: on). |
 | `-v`, `--verbose` | Enable verbose logging. |
 
-The `uv` package cache and Bazel's downloaded dependency archives persist
-across sandbox runs in Docker named volumes, `cc-protocol-uv-cache` and
-`cc-protocol-bazel-repository-cache`. These grow without bound since neither
-tool prunes them; to reset one, run `docker volume rm <name>`.
+The `uv` package cache and Bazel's repository cache persist across sandbox runs
+in Docker named volumes, `cc-protocol-uv-cache` and
+`cc-protocol-bazel-repository-cache`. To reset either volume, run `docker volume
+rm <name>`.
 
 ### Using pre-commit Locally to run Github Workflow checks
 
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -144,4 +144,6 @@ ENV XYZ_BAZEL_REPOSITORY_CACHE_DIR="/home/vscode/.cache/bazel-repo"
 # into a root-owned directory.
 RUN mkdir -p "$UV_CACHE_DIR" "$XYZ_BAZEL_REPOSITORY_CACHE_DIR"
 
-RUN echo "common --repository_cache=$XYZ_BAZEL_REPOSITORY_CACHE_DIR" > /home/vscode/.bazelrc
+# repository_cache is defined in the sandbox user-directory config, not workspace config so only
+# the sandbox uses the respository cache.
+RUN echo "common --repository_cache=$XYZ_BAZEL_REPOSITORY_CACHE_DIR" >> /home/vscode/.bazelrc
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -137,7 +137,10 @@ FROM base AS sandbox
 # uv must copy files as the cache volume is on a different filesystem.
 ENV UV_LINK_MODE="copy"
 
+ENV BAZEL_CACHE_DIR="/home/vscode/.cache/bazel-repo"
+ENV PRE_COMMIT_HOME="/home/vscode/.cache/pre-commit"
+
 # Pre-create the cache dirs owned by vscode: Docker otherwise creates a
 # volume's mount point as root on first use, and these tools cannot write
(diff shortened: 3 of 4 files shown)


philipcraig:
testing finding, on Ubuntu with docker on WSL2 "The Linux cache-volume test worked (cold run 1m20s, warm run 2.9s, no recompilation), disproving the PR's macOS-only assumption."

jbcoe:
Thanks @philipcraig that's encouraging. Do we need dedicated bare-lInux tests or do we just trust that WSL is representative? What about plain Windows (git-bash)?

Let's land https://github.com/jbcoe/cc-protocol/pull/301 as-is and relax the default in a follow up PR. That leaves proper breadcrumbs for later.

philipcraig:
Yes merge as is. I never use bare windows and have no working setup there so I can't comment but wsl Ubuntu is extremely Linux so I'd have confidence for that

scripts/agentic-sandbox.py
```diff
@@ -15,7 +15,10 @@
 
 # Docker named volumes persisting each tool's cache across container instances.
 # The devcontainer is long-lived and does not use these, it keeps its cache locally.
-CACHE_VOLUMES: dict[str, str] = {"cc-protocol-uv-cache": "/home/vscode/.cache/uv"}
+CACHE_VOLUMES: dict[str, str] = {
+    "cc-protocol-uv-cache": "/home/vscode/.cache/uv",
+    "cc-protocol-bazel-repository-cache": "/home/vscode/.cache/bazel-repo",
```
philipcraig, line 21: An image built after #301 but before this PR is never rebuilt: `image_exists` is true, so the script mounts the new volume on a path the old image lacks. Docker then creates the mount point and the volume root as root:root, and every Bazel command in the sandbox aborts with `could not acquire lock on repo contents cache` before fetching anything, because Bazel 9 takes `<cache>/contents/gc_lock` at startup.

Either rebuild automatically when `docker/Dockerfile` is newer than the image (or its hash differs from a label on the image), or say in CONTRIBUTING.md that this change needs `--rebuild-docker`. Recovery is just `--rebuild-docker`, since the volume is still empty and the rebuilt image's ownership is copied in on the next mount.

.bazelrc
```diff
@@ -2,6 +2,11 @@
 build --cxxopt=-std=c++26
 build --cxxopt=-freflection
 
+# Cache downloaded external dependency archives in their own directory,
+# separate from the default output root (~/.cache/bazel), which also holds
+# build artifacts we don't want to persist across sandbox runs.
+common --repository_cache=~/.cache/bazel-repo
```
philipcraig: On Bazel 9 `--repository_cache` also relocates the repo contents cache, whose default is `<repository_cache>/contents`. So the volume holds extracted repo trees rather than only archives (about 104M of 138M in a test population), `external/<repo>` in the output base becomes a symlink into the volume, each miss falls back to a full copy because the rename crosses the overlay/volume boundary, and Bazel garbage-collects contents entries after 14 idle days. The comment, the volume name, and CONTRIBUTING.md all describe archives only.

If persisting extracted repos is unintended, add `common --repo_contents_cache=` here to disable it, or point it under the output root. If it is intended (it does help offline builds), say so in this comment and in CONTRIBUTING.md, including the 14-day expiry.

philipcraig: The comment's justification ("persist across sandbox runs") applies only to the sandbox, but this workspace-wide rc relocates the cache for the devcontainer, contributors' hosts, CI runners, and the Copilot agent. Each of those abandons its warm per-user cache under `~/.cache/bazel/_bazel_<user>/cache/repos/v1`, which is shared with every other Bazel workspace on the machine, and re-downloads everything once. CI gains nothing because setup-bazel's `repository-cache` input is off in all three workflows, and if it were on, setup-bazel writes this same line to `~/.bazelrc` itself.

Suggest leaving `.bazelrc` untouched and having the `sandbox` Docker stage write `common --repository_cache=/home/vscode/.cache/bazel-repo` into `/home/vscode/.bazelrc`. The user rc is read after the workspace rc, and neither `scripts/bazel.py` nor `agentic-sandbox.py` passes `--nohome_rc` or overrides `HOME`, so it takes effect only inside the sandbox.

docker/Dockerfile
```diff
@@ -137,7 +137,9 @@ FROM base AS sandbox
 # uv must copy files as the cache volume is on a different filesystem.
 ENV UV_LINK_MODE="copy"
 
+ENV BAZEL_REPOSITORY_CACHE_DIR="/home/vscode/.cache/bazel-repo"
```
philipcraig: Neither Bazel nor bazelisk reads `BAZEL_REPOSITORY_CACHE_DIR`. Unlike `UV_CACHE_DIR` next to it, which uv honours, this only feeds the `mkdir`; the real location is a literal in `.bazelrc` and a third copy in `CACHE_VOLUMES` in `scripts/agentic-sandbox.py`, with nothing tying the three together. Editing the ENV, as works for uv, would silently leave Bazel writing to one path while the volume is mounted on another, and every sandbox run would re-download.

Two ways to fix it. Make the ENV real: have this stage write `common --repository_cache=$BAZEL_REPOSITORY_CACHE_DIR` into `/home/vscode/.bazelrc` (see the `.bazelrc` comment), leaving `CACHE_VOLUMES` as the one remaining copy with a cross-reference comment. Or drop the ENV, inline the path in the `mkdir`, and add a comment naming the other two copies.

.bazelrc
```diff
@@ -2,6 +2,11 @@
 build --cxxopt=-std=c++26
 build --cxxopt=-freflection
 
+# Cache downloaded external dependency archives in their own directory,
+# separate from the default output root (~/.cache/bazel), which also holds
+# build artifacts we don't want to persist across sandbox runs.
```
philipcraig: This declines to persist build artefacts without saying why. Every `docker run --rm` sandbox session therefore recompiles googletest and all the reflection-heavy test and tutorial TUs from an empty output base, and each `--asan`/`--ubsan`/`--tsan` config recompiles googletest again. That is the dominant cost of a sandbox build, and the download this PR saves is seconds by comparison.

A `--disk_cache` volume would skip it: add `cc-protocol-bazel-disk-cache` -> `/home/vscode/.cache/bazel-disk` to `CACHE_VOLUMES` and `common --disk_cache=~/.cache/bazel-disk` (setup-bazel's path) alongside this line, with `--experimental_disk_cache_gc_max_size` if growth is a concern. It is safe here because `scripts/bazel.py` keys every action on `XYZ_PROTOCOL_COMPILER_VERSION` and the absolute compiler paths, and the agent already has write access to the source tree and the repository-cache volume, so a writable action cache adds no new trust boundary. If artefacts are deliberately excluded, the comment should give the reason.

CONTRIBUTING.md
```diff
@@ -230,12 +230,13 @@ plain shell in the container.
...
-| `--cache-volumes`, `--no-cache-volumes` | Mount the persistent `uv` cache volume (default: on). |
+| `--cache-volumes`, `--no-cache-volumes` | Mount the persistent cache volumes (default: on). |
 | `-v`, `--verbose` | Enable verbose logging. |
 
-The `uv` cache persists across sandbox runs in a Docker named volume,
-`cc-protocol-uv-cache`. The cache grows without bound since `uv` never prunes
-it; to reset it, run `docker volume rm cc-protocol-uv-cache`.
+The `uv` package cache and Bazel's downloaded dependency archives persist
+across sandbox runs in Docker named volumes, `cc-protocol-uv-cache` and
+`cc-protocol-bazel-repository-cache`. These grow without bound since neither
```
philipcraig: "These grow without bound since neither tool prunes them" is wrong for Bazel: the `contents/` half of the volume is garbage-collected after 14 idle days, and only `content_addressable/` archives are kept forever. "Downloaded dependency archives" also omits the extracted repos.

The feature itself is undocumented too. Please add that the first sandbox run, and any `MODULE.bazel` or `MODULE.bazel.lock` bump, must be online before offline builds work (an empty cache fails with `Failed to fetch registry file https://bcr.bazel.build/...`, with no hint that a prior online run is needed). If the rc line stays workspace-wide, also tell non-sandbox users that their cache moved to `~/.cache/bazel-repo` and that `rm -rf ~/.cache/bazel-repo` resets it, since `--clean` only expunges the output base.

philipcraig (changes requested):
The mechanism works: with a populated cache and a fresh output base, an offline `bazel fetch //...` succeeds on Bazel 9.2.0, and with an empty cache it fails at the first BCR registry fetch. The comments below are about a stale-image trap, an undocumented side effect, and where the setting lives.

scripts/agentic-sandbox.py
jbcoe: `CONTRIBUTING.md` is not the right place to put PR-specific notes.

I've added a note to the PR description: "Syncing past this change needs the agentic sandbox to be run with `--rebuild-docker`."

jbcoe:
@philipcraig this is looking pretty minimal now. Thanks for review input. Isolating changes to the sandbox is a solid improvement.

scripts/agentic-sandbox.py
```diff
@@ -15,7 +15,11 @@
 
 # Docker named volumes persisting each tool's cache across container instances.
 # The devcontainer is long-lived and does not use these, it keeps its cache locally.
-CACHE_VOLUMES: dict[str, str] = {"cc-protocol-uv-cache": "/home/vscode/.cache/uv"}
+# Note: Cache paths must match those set in docker/Dockerfile.
+CACHE_VOLUMES: dict[str, str] = {
+    "cc-protocol-uv-cache": "/home/vscode/.cache/uv",
+    "cc-protocol-bazel-repository-cache": "/home/vscode/.cache/bazel-repo",
```
philipcraig, line 21: With the rc line now inside the image, a stale image no longer fails loudly. `image_exists` (line 152) is true, the old image has no `~/.bazelrc`, Bazel uses its default cache in the discarded output base, and every session re-downloads with nothing pointing at the image. The PR-body note covers this one migration, but every later Dockerfile change needs the same note again.

Labelling the image with a hash of `docker/Dockerfile` and `.bazelversion` at build time and rebuilding when it differs, or simply always running the layer-cached `docker build`, fixes the class and retires the manual rebuild note at CONTRIBUTING.md line 20.

docker/Dockerfile
```diff
@@ -137,7 +137,11 @@ FROM base AS sandbox
...
 
+ENV XYZ_BAZEL_REPOSITORY_CACHE_DIR="/home/vscode/.cache/bazel-repo"
+
 # Pre-create the cache dirs owned by vscode: Docker otherwise creates a
 # volume's mount point as root on first use, and these tools cannot write
 # into a root-owned directory.
-RUN mkdir -p "$UV_CACHE_DIR"
+RUN mkdir -p "$UV_CACHE_DIR" "$XYZ_BAZEL_REPOSITORY_CACHE_DIR"
+
+RUN echo "common --repository_cache=$XYZ_BAZEL_REPOSITORY_CACHE_DIR" > /home/vscode/.bazelrc
```
philipcraig: Two small things on this line. It deserves a comment saying it lives in the user rc, not the workspace `.bazelrc`, so that only the sandbox relocates its cache; the first revision of this PR shows how easily it migrates. And `>` replaces the file, so any user-rc line the `base` stage later gains is silently dropped from the sandbox image. `>>` costs nothing.

CONTRIBUTING.md
```diff
@@ -230,12 +230,13 @@ plain shell in the container.
...
 | `--rebuild-docker` | Rebuild the Docker image before starting. |
 | `--update` | Update the agent CLI to the latest version before running. |
-| `--cache-volumes`, `--no-cache-volumes` | Mount the persistent `uv` cache volume (default: on). |
+| `--cache-volumes`, `--no-cache-volumes` | Mount the persistent cache volumes (default: on). |
 | `-v`, `--verbose` | Enable verbose logging. |
 
-The `uv` cache persists across sandbox runs in a Docker named volume,
-`cc-protocol-uv-cache`. The cache grows without bound since `uv` never prunes
-it; to reset it, run `docker volume rm cc-protocol-uv-cache`.
+The `uv` package cache and Bazel's repository cache persist across sandbox runs
```
philipcraig, line 236: This paragraph now says the volumes exist but not what the Bazel one is for or when it works. Please add that the first sandbox run, and any run after a volume reset or a `MODULE.bazel` / `MODULE.bazel.lock` change, must be online; an empty cache fails with `Failed to fetch registry file https://bcr.bazel.build/...` and nothing says a prior online run is needed.

The reset text also lost ground. The old sentence was wrong about Bazel but its point stands: uv wheels and Bazel's archives are never pruned, and only Bazel's extracted repos expire after 14 idle days. Keep one corrected sentence, give both `docker volume rm` commands verbatim rather than `<name>`, and say to exit any open sandbox first, since the command fails with `volume is in use` otherwise.

philipcraig (commented):
Second pass at 9640356. The rc line moving into the sandbox image and the ENV now feeding it resolve two of my earlier points. Two others I have reopened as they still apply unchanged; the rest are below.

The absence of any build or lint of the sandbox stage is filed as #325 rather than raised here.

.bazelrc
philipcraig: Still stands at 9640356; the line is now docker/Dockerfile:147. The volume holds extracted repo trees as well as archives, shares them live between concurrent sandboxes, and `--clean` does not touch them.

.bazelrc
philipcraig: Still stands at 9640356. Not a blocker, but a one-line reason next to the rc line, or a follow-up issue, would answer it.

scripts/agentic-sandbox.py
jbcoe: Labelling/versioning the docker image is not in scope for this change. It's not a new problem and not a problem that needs addressing with any urgency.

I'm perfectly happy with a PR body comment.

CONTRIBUTING.md
jbcoe: Saying that a prior online run is needed to get remote files is unnecessary.

The reset text contains sufficient information. 

I'm trying to minimise the wall of text and sideshows.

.bazelrc
jbcoe: We can handle this as follow up if and when it is needed.

philipcraig (approved):
awesome

PR #305, opened 2026-09-06
Death tests for valueless access through a protocol_view

### LINKED ISSUE #284: Constructing a protocol_view from a valueless protocol is undiagnosed, and calls through it hit a null function pointer
A moved-from `protocol` has `object_ == nullptr` and `vtable_ == &null_vtable`,
whose member entries are null function pointers. Calling a member function on
it directly is caught in debug builds: `method_thunk::operator()` asserts
`!valueless_after_move()` before dispatching.

Since #244 and #272 a `protocol_view<I>` or `protocol_view<const I>` constructs
implicitly from a `protocol<I, A>&`, copying `object_` and `vtable_`. Nothing
checks the source, so a valueless protocol converts silently, and the thunk
assert is `if constexpr (is_protocol_v<ProtocolType>)`, so it is compiled out
for views. A member call through such a view calls a null function pointer with
no diagnostic, even in a debug build.

```c++
protocol<Shape> a{Circle{}};
protocol<Shape> b = std::move(a);   // a is valueless
protocol_view<const Shape> v = a;   // accepted; comment says "Precondition: p is not valueless"
v.area();                           // release: null call; debug: also null call
```

The from-protocol constructors document the precondition in a comment
(`protocol.hh`, both `protocol_view` specializations) but do not enforce it.

Suggested fix, small enough for one PR:

- `assert(!p.valueless_after_move())` in the two from-protocol constructors.
- Drop the `is_protocol_v` gate in `method_thunk` and assert `object_ != nullptr`
  instead, which covers protocol and view alike. A view's `object_` is otherwise
  never null.
- A death test for each path.

#208 settled the direct-call case as a precondition violation diagnosed by an
assert in debug builds; this extends the same contract to views. Raised in the
review of #272, where it was agreed to be out of scope there but worth fixing.

jbcoe:
Fully populate the null vtable to avoid null pointer dereferences on a moved-from object.

Closes #284


protocol.hh
```diff
@@ -633,6 +624,53 @@ struct const_view_trampoline<R (*)(const void*, Args...) noexcept(Noexcept), U,
...
+// gcov never records this function's body as executed: EXPECT_DEATH runs it
+// in a forked child that terminates via SIGABRT, and gcov's counters are
+// only flushed by an atexit hook, which a signal-terminated process never
+// runs.
+// GCOVR_EXCL_START
+template <typename R, typename... Args, bool Noexcept>
+struct invalid_call_trampoline<R (*)(void*, Args...) noexcept(Noexcept)> {
+  static R call(void*, Args...) noexcept(Noexcept) {
+    assert(false && "cannot call member function of valueless protocol");
+    std::abort();
```
RyanJK5, line 643: Nitpick, but I think `std::terminate` is slightly more idiomatic here?

RyanJK5 (approved):

protocol.hh
```diff
@@ -813,16 +851,18 @@ class protocol
...
 
-  // A no-op vtable that is the stand-in for a nullptr vtable. Prevents
-  // redundant null checks in the special member functions. The member
-  // function entries are left null: calling a member function on a
-  // valueless protocol is a precondition violation.
+  // A vtable for a valueless protocol. Using the null vtable, the special
+  // member functions don't need a null check. Its ownership entries are
+  // no-ops; its member function entries are `invalid_call_trampoline`s, so
+  // calling a member function on a valueless protocol, or through a view of
+  // one, aborts with a diagnostic.
```
philipcraig, line 858: Two over-promises here. "aborts with a diagnostic" is true only without NDEBUG; a release build gets a bare SIGABRT with no message (the trampoline comment at 627-630 states this correctly). And "through a view of one" holds only for a view constructed from an already-valueless protocol: a view taken before its owner is moved through the allocator-unequal path keeps the real vtable and a dangling `object_`, so that call is still UB. Suggest scoping this to "a view constructed from a valueless protocol" and dropping "with a diagnostic", or routing both trampolines through one `[[noreturn]]` helper that always prints before aborting so the comment becomes true.

protocol_test.cc
```diff
@@ -2409,6 +2409,52 @@ TEST(ReflectionProtocolViewTest, ViewOfProtocolRemainsValidAfterMove) {
   EXPECT_EQ(view.get(), 5);
 }
 
+// Tests that dispatching through a view of a moved-from protocol fails the
+// same way as dispatching on the protocol directly.
+#if (defined(_MSC_VER) && defined(_DEBUG)) || (!defined(NDEBUG))
```
philipcraig, line 2414: This guard means the Release CMake job compiles both tests out, and every debug build trips `assert(false)` before reaching `std::abort()`. The release-only path this PR adds is therefore never executed by CI, so a later "the assert is enough" edit would reopen #284 for release users without a signal. `EXPECT_DEATH(view.foo(), "")` with no guard covers both configurations (an empty regex matches the silent abort), or pick the regex under `#ifdef NDEBUG`. Same applies to `MutableValuelessCall`/`ConstValuelessCall` at 2227, and these two tests are those two with one line inserted, so they could just be extra lines there.

philipcraig (commented):
The mechanism looks right: `object_ == nullptr` iff `vtable_ == &null_vtable` holds in every `protocol` state, and every vtable entry type matches a trampoline specialization. Three comments elsewhere in the header now disagree with it:

- Both `protocol_view(protocol<T, Alloc>&)` constructors (`protocol.hh` 1113 and 1180) still say "Precondition: `p` is not valueless", which the new tests contradict by construction. Either reword (constructing a view of a valueless `p` is allowed; calling a member through it aborts) or add the constructor assert #284 asked for and move the death expectation to the construction.
- The class comment at 762 still calls a valueless call a precondition violation. With this PR it is an unconditional abort in every build, including release. That is an observable design choice (UB vs hardened abort vs `std::terminate`, which is what RyanJK5's question is really about), so it belongs in that comment and in DRAFT.md, which currently says nothing about calls on a valueless protocol or view.

Inline: the release path is untested, one comment over-promises, and two simplifications.

protocol.hh
```diff
@@ -633,6 +624,53 @@ struct const_view_trampoline<R (*)(const void*, Args...) noexcept(Noexcept), U,
...
+// GCOVR_EXCL_START
+template <typename R, typename... Args, bool Noexcept>
+struct invalid_call_trampoline<R (*)(void*, Args...) noexcept(Noexcept)> {
+  static R call(void*, Args...) noexcept(Noexcept) {
+    assert(false && "cannot call member function of valueless protocol");
+    std::abort();
+  }
+};
+
+template <typename R, typename... Args, bool Noexcept>
```
philipcraig, line 647: The two specializations differ only in an unused `void*`/`const void*` first parameter. A single `invalid_call_trampoline<R (*)(Args...) noexcept(N)>` matches every entry (the object pointer is just the first `Arg`), and a non-template `[[noreturn]]` helper holding the assert and abort keeps one copy of the message, one GCOVR_EXCL region, and one `__assert_fail` site instead of one per signature.

protocol.hh
```diff
@@ -633,6 +624,53 @@ struct const_view_trampoline<R (*)(const void*, Args...) noexcept(Noexcept), U,
...
+// GCOVR_EXCL_STOP
+
+// Builds a vtable for `T` whose entries are all `invalid_call_trampoline`s:
+// the null vtable shared by every valueless `protocol<T>`.
+template <typename T>
+consteval typename vtable_generator<T>::vtable make_invalid_vtable() {
+  using Vtable = typename vtable_generator<T>::vtable;
+  Vtable result{};
+
+  template for (constexpr std::meta::info member :
```
philipcraig, line 664: This only needs to fill every entry, so iterating `nonstatic_data_members_of(^^Vtable, access_context::unprivileged())` directly is O(M) rather than a `find_vtable_entry` lookup per member, and it cannot leave an entry null if the vtable ever gains entries that are not 1:1 with `protocol_interface_functions_of`.

jbcoe:
A rethink needed here.

RyanJK5:
> A rethink needed here.

I think it should just be UB.

Issue #277, opened 2026-09-01
Track clang-p2996 issues that affect this repository


philipcraig:
Upstream bugs in [bloomberg/clang-p2996](https://github.com/bloomberg/clang-p2996) that we work around or that constrain how we use the fork's clang-tidy. When one is fixed: bump `.github/clang-p2996-commit`, confirm the Clang Tidy job is green, and either remove the workaround or note here why it stays.

| Upstream | Symptom here | Workaround | When fixed |
|---|---|---|---|
| [#342](https://github.com/bloomberg/clang-p2996/issues/342), fix [PR #353](https://github.com/bloomberg/clang-p2996/pull/353) | `^^derived::operator()` rejected when the call operator comes from a using-declaration ("cannot take the reflection of a using-declarator"); rejects-valid, GCC accepts | #276 excludes types with base classes from `is_maybe_lambda` | Keep the exclusion; closure types have no bases, so it remains a correct refinement. Once the pin is bumped, `is_protocol_conformant_v` no longer needs it to pass clang-tidy. |
| [#349](https://github.com/bloomberg/clang-p2996/issues/349), fixed by [PR #351](https://github.com/bloomberg/clang-p2996/pull/351) (merged 2026-09-08 as bde8bcf3) | `define_aggregate` with a member type spelled as a qualified name under a pointer or cv-qualifier (`^^const std::type_info*`) yields a field with a broken qualifier location; clang-tidy's renamer checks segfault (#306 CI) | Reflect an unqualified alias (`using type_info_pointer = const std::type_info*;`) instead of the spelled type | Alias can stay; drop the explanatory comment. The fix is in the type-location filler, not `define_aggregate`, so any spelled qualified type under a pointer or cv-qualifier is covered. |
| [#350](https://github.com/bloomberg/clang-p2996/issues/350), fix [PR #352](https://github.com/bloomberg/clang-p2996/pull/352) | `vtable->[:entry:]` on an lvalue pointer trips "-> base must be a pointer prvalue" in assertion-enabled builds only; release builds (CI) are unaffected | None needed; do not build the pinned toolchain with `LLVM_ENABLE_ASSERTIONS=ON` for use on this code | Asserting local builds become usable; `array->[:m:]` also starts working. |

Not bugs, noted so they are not re-reported: [#230](https://github.com/bloomberg/clang-p2996/issues/230) (`^^std::int32_t` rejected; [expr.reflect]/5.1 applies to the identifier form).

PR #306, opened 2026-09-07
Add protocol_cast

### LINKED ISSUE #279: Consider function to extract underlying type from protocol
It may be useful to provide some utility for getting the type underlying a `protocol`, similar to `any_cast`, e.g. :

```c++
protocol<Interface> p{MyType{}};
// pass it around a bunch of places...
auto reclaimed = xyz::get<MyType>(std::move(p));
```

RyanJK5:
Adds protocol_cast for protocol and protocol_view using RTTI.

Closes #279

### DIFF
--- a/protocol.hh
+++ b/protocol.hh
@@ -852,6 +852,18 @@ class protocol
   template <is_valid_view_interface>
   friend class protocol_view;
 
+  // Grants protocol_cast access to protocol's underlying data.
+  template <typename T, typename Protocol>
+    requires(is_protocol_v<std::remove_cvref_t<Protocol>>)
+  friend constexpr T protocol_cast(Protocol&& operand);
+
+  template <typename T, typename Interface>
+  friend constexpr T* protocol_cast(protocol<Interface>* operand) noexcept;
+
+  template <typename T, typename Interface>
+  friend constexpr const T* protocol_cast(
+      const protocol<Interface>* operand) noexcept;
+
   [[no_unique_address]] Alloc alloc_;
 
   void* object_ = nullptr;
@@ -1052,6 +1064,40 @@ class protocol
     return object_ == nullptr;
   }
 };
+
+template <typename T, typename Protocol>
+  requires(is_protocol_v<std::remove_cvref_t<Protocol>>)
+constexpr T protocol_cast(Protocol&& operand) {
+  if (operand.vtable_ != &Protocol::template vtable_for<T>) {
+    throw bad_protocol_cast{};
+  }
+
+  using pointer_type =
+      std::conditional_t<std::is_const_v<std::remove_reference_t<Protocol>>,
+                         const T*, T*>;
+  return std::forward_like<Protocol>(
+      *static_cast<pointer_type>(operand.object_));
+}
+}
+
+template <typename T, typename Interface>
+constexpr T* protocol_cast(
+    xyz::reflection::protocol<Interface>* operand) noexcept {
+  if (operand.vtable_ != &protocol<Interface>::template vtable_for<T>) {
+    return nullptr;
+  }
+
+  return static_cast<T*>(operand_ > object_);
+}
+
+template <typename T, typename Interface>
+constexpr const T* protocol_cast(const protocol<Interface>* operand) noexcept {
+  if (operand.vtable_ != &protocol<Interface>::template vtable_for<T>) {
+    return nullptr;
+  }
+
+  return static_cast<const T*>(operand->object_);
+}
 
 // ---------------------------------------------------------------------------
 // protocol_view<T>
--- a/protocol.hh
+++ b/protocol.hh
@@ -1185,11 +1185,13 @@ class protocol_view
   friend struct detail::method_thunk;
 
   template <typename U, typename Interface>
-    requires(is_protocol_conformant_v<Interface, U>)
+    requires(!std::is_const_v<Interface> &&
+             is_protocol_conformant_v<Interface, U>)
   friend constexpr U& protocol_cast(protocol_view<Interface> operand);
 
   template <typename U, typename Interface>
-    requires(is_protocol_conformant_v<Interface, U>)
+    requires(!std::is_const_v<Interface> &&
+             is_protocol_conformant_v<Interface, U>)
   friend constexpr U* protocol_cast(protocol_view<Interface>* operand) noexcept;
 
   // Non-owning pointer to the viewed object.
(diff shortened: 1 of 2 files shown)


RyanJK5:
This appears to make clang-tidy segfault 😆

philipcraig (commented):
The clang-tidy crash is a clang-p2996 bug in `define_aggregate`, reported as bloomberg/clang-p2996#349 and tracked here in #277. A `data_member_spec` whose type is spelled with a nested-name-specifier under a pointer or cv-qualifier (`^^const std::type_info*`) gets a field whose `std::` qualifier has no source-location data; the compiler doesn't care, but clang-tidy's renamer checks walk that location and segfault. Any of the existing function-pointer entries are fine because function types take a different initialisation path.

The minimal trigger is a `const ns::T*` member; a bare `^^ns::T` doesn't trip it.

Workaround that keeps the design: reflect an unqualified alias instead of the spelled type.

```cpp
// in namespace detail, before generate_vtable_specs
using type_info_pointer = const std::type_info*;
...
  function_pointer_specs.push_back(data_member_spec(
      ^^type_info_pointer, {.name = "xyz_protocol_typeid"}));
```

`<typeinfo>` should also be included directly for `typeid`.

With that change clang-tidy on the pinned commit is clean and the GCC Bazel tests pass. `substitute(^^std::add_pointer_t, {^^const std::type_info})` also works if you'd rather not name an alias.

Once the crash is gone, clang-tidy reports five findings in the new tests, all `-warnings-as-errors`:

- `ConstProtocolCast`: `auto&`/`auto*` on the two `protocol_cast<Conforming>(cv)` results should be `const auto&`/`const auto*` (readability-qualified-auto), and `underlying_ptr` is written but never read (clang-analyzer DeadStores); an `EXPECT_EQ` on `.value` fixes the latter.
- `ProtocolCastValueCategory`: `const protocol p2(p);` after `std::move(p)` is flagged as use-after-move; taking `b` last avoids it. `std::move(p2)` on a const object is flagged as a no-op move (performance-move-const-arg); `static_cast<const protocol<Interface>&&>(p2)` gets the same `const Conforming&&` without it.

On the RTTI question: a `const void*` pointing at a per-type tag object (`template <class T> inline constexpr char type_tag = 0;`) has an unqualified type, so it also sidesteps the bug, and it drops the RTTI dependency. Comparing addresses of an inline variable is reliable within one program on ELF; `typeid` comparison is the safer choice across shared-library boundaries on other platforms.

philipcraig:
The changes above are on `protocol-cast-clang-tidy`, one commit on top of 2e3b480: https://github.com/jbcoe/cc-protocol/compare/protocol-cast...protocol-cast-clang-tidy

`git cherry-pick 4d778f7` onto this branch picks it up, or take whichever parts you want.

RyanJK5:
> On the RTTI question: a const void* pointing at a per-type tag object (template <class T> inline constexpr char type_tag = 0;) has an unqualified type, so it also sidesteps the bug, and it drops the RTTI dependency. Comparing addresses of an inline variable is reliable within one program on ELF; typeid comparison is the safer choice across shared-library boundaries on other platforms.

I was thinking about some kind of tag as well, but like you said it's less bulletproof if there's something dynamically linked. I suppose RTTI is the simplest/most-likely-to-be-correct solution for now.

jbcoe:
I'm happy with RTTI as it is part of standard C++ and I think it's the approach used by `std::any`.

RyanJK5:
> I'm happy with RTTI as it is part of standard C++ and I think it's the approach used by `std::any`.

Do we want some way to let users access the RTTI `protocol` is storing then, like `std::any::type`?

jbcoe:
> > I'm happy with RTTI as it is part of standard C++ and I think it's the approach used by `std::any`.
> 
> Do we want some way to let users access the RTTI `protocol` is storing then, like `std::any::type`?

I'd follow `std::any` until we can't or have a _very_ strong reason not to.

RyanJK5:
> > > I'm happy with RTTI as it is part of standard C++ and I think it's the approach used by `std::any`.
> > 
> > 
> > Do we want some way to let users access the RTTI `protocol` is storing then, like `std::any::type`?
> 
> I'd follow `std::any` until we can't or have a _very_ strong reason not to.

I think it would be useful to have. My one concern is that adding functions to `protocol` and `protocol_view` can result in clashes with functions in user-defined interfaces. So if we say `myProtocol.type()` is always defined, then users can't have an interface member called `type()`. It's a small tradeoff though.

I think this makes sense as a follow-up, I will open an issue.

jbcoe:
We can add free functions `protocol::type(protocol_like T& t)` etc, like `std::variant` does.

jbcoe:
I'm leaning towards having `valueless_after_move` as a free function too.

protocol.hh
```diff
@@ -1187,11 +1292,42 @@ class protocol_view<const T> : public detail::protocol_wrappers_t<
...
+
   // Non-owning pointer to the viewed object.
   const void* object_ = nullptr;
 
   const detail::vtable_generator<T>::vtable* vtable_;
 };
 
+template <typename U, typename Interface>
+  requires(is_protocol_conformant_v<Interface, std::decay_t<U>>)
+constexpr const U& protocol_cast(protocol_view<const Interface> operand) {
```
jbcoe: Might these be better as hidden friends? 

https://isocpp.org/blog/2023/01/argument-dependent-lookup-and-the-hidden-friend-idiom-rainer-grimm

RyanJK5: I considered it, what do you think? I was trying to follow the precedent of std::any, which I believe just plops it in the std namespace. That could very well be the wrong decision though.

jbcoe: I suspect that the design of `std::any` predates hidden friends.

I'd prefer hidden friends for faster overload set resolution but it's a very weak preference.

RyanJK5: It would simplify the code as well. I'm in favor.

jbcoe (approved):
Thanks for this.

jbcoe (commented):
The PR comment will need amending to give us a nice commit message. Otherwise everything here looks good.

jbcoe (changes requested):
Blocking submission on a PR description update.

philipcraig:
#316 bumped the pinned clang-p2996 to a commit that includes bloomberg/clang-p2996#351, so now the Clang Tidy job accepts `^^const std::type_info*` directly in `data_member_spec`. The `type_info_pointer` alias, its `[[maybe_unused]]` and the comment citing bloomberg/clang-p2996#349 can then be dropped and the type spelled inline.

jbcoe (approved):
Looks great - thanks for this!

protocol.hh
```diff
@@ -558,6 +559,16 @@ template <std::meta::info interface_type>
 consteval std::vector<std::meta::info> generate_vtable_specs() {
   std::vector<std::meta::info> function_pointer_specs;
 
+  // The type of the vtable's `xyz_protocol_typeid` entry. Reflected through an
```
philipcraig: It would be nice to remove all this rigmarole now that this bug in clang-p2996 is fixed

RyanJK5: Should be fixed

PR #315, opened 2026-09-07
Add CMake FetchPackage cache for agentic sandbox

jbcoe:
With a populated cache, CMake builds can complete in the agentic sandbox with no internet connection.

See: https://cmake.org/cmake/help/latest/module/FetchContent.html


CMakeLists.txt
```diff
@@ -131,12 +131,23 @@ target_sources(
 if(XYZ_PROTOCOL_IS_NOT_SUBPROJECT)
 
   if(${BUILD_TESTING})
+    # Caches downloaded archives outside the build directory, so a fresh build
+    # directory doesn't re-fetch. Left empty when $HOME is unset (e.g. on
+    # Windows), which FetchContent treats as no persistent cache.
+    set(XYZ_PROTOCOL_FETCHCONTENT_DOWNLOAD_CACHE_DIRECTORY "")
```
philipcraig: Before this PR configure never touched anything outside the build directory. Now every CMake build (devcontainer, host, CI) writes to `$HOME/.cache/cmake-fetchcontent`, and because this is a plain `set()` rather than a cache variable, `-DXYZ_PROTOCOL_FETCHCONTENT_DOWNLOAD_CACHE_DIRECTORY=...` is silently overwritten and there is no opt-out. Three reproduced failures follow:

- `HOME=` (empty, as cron/systemd/container shells sometimes set it) passes `if(DEFINED ENV{HOME})`, the path becomes `/.cache/cmake-fetchcontent`, and the sub-build fails with `file failed to create directory`.
- `HOME=C:\Users\...` (Git Bash, GitHub Windows runners) is written verbatim into the generated sub-build `CMakeLists.txt` and fails with `Invalid character escape '\U'`. The comment's premise that HOME is unset on Windows is the benign case, not the common one.
- A root-owned or read-only `~/.cache` (sudo-created, NFS home, `docker run --user`) fails at `file DOWNLOAD cannot open file for write`.

Suggest the pattern `XYZ_PROTOCOL_CLANG_P2996_DIRECTORY` already uses: `set(XYZ_PROTOCOL_FETCHCONTENT_DOWNLOAD_CACHE_DIRECTORY "$ENV{XYZ_PROTOCOL_FETCHCONTENT_DOWNLOAD_CACHE_DIRECTORY}" CACHE PATH "...")`, defaulting to empty, and drop the HOME branch. Then the sandbox stage in `docker/Dockerfile` sets that one ENV instead of `CMAKE_FETCHCONTENT_DOWNLOAD_CACHE_DIR`, which nothing currently reads (it only feeds the `mkdir`, and its `_DIR` suffix breaks the AGENTS.md naming rule), and `CACHE_VOLUMES` becomes the only other copy of the path. Note also that an empty `DOWNLOAD_DIR` is not a FetchContent mode, it just falls back to the default location inside the build directory, so the comment should say that.

CMakeLists.txt
```diff
@@ -131,12 +131,23 @@ target_sources(
...
     # SYSTEM marks the dependencies' include directories as system headers so
     # that clang-tidy and compiler warnings do not report on third-party code.
     FetchContent_Declare(
       googletest
-      GIT_REPOSITORY https://github.com/google/googletest.git
-      GIT_TAG v1.18.0
+      URL https://github.com/google/googletest/archive/refs/tags/v1.18.0.tar.gz
+      URL_HASH
+        SHA256=6e3191c1455468b3fc35a417fb565c1c5071aee1b7e7f85e30cf48a98d37d8b5
+      DOWNLOAD_DIR "${XYZ_PROTOCOL_FETCHCONTENT_DOWNLOAD_CACHE_DIRECTORY}"
```
philipcraig: All configures under one HOME (or one Docker volume) now share a single download target with no temp-file rename and no lock. With a cold cache, `./scripts/cmake.sh --debug` and `--release` in two terminals delete each other's partial archive (the download step `file(REMOVE)`s any existing file whose hash mismatches), ping-pong through all six retries, and both fail with `Each download failed!`, leaving the cache empty. Reproduced with a throttled server and a short stagger. Before the PR each build directory had a private download.

Either wrap `FetchContent_MakeAvailable` in `file(LOCK "${XYZ_PROTOCOL_FETCHCONTENT_DOWNLOAD_CACHE_DIRECTORY}/.lock" GUARD PROCESS)` when the directory is set, or document that the cache must not be shared by concurrent configures.

philipcraig: `DOWNLOAD_DIR` is honoured only because CMP0168 is OLD under `cmake_minimum_required(3.25)`. Under CMP0168 NEW (a minimum-version bump to 3.30, a `cmake_policy(VERSION)` call, or `-DCMAKE_POLICY_DEFAULT_CMP0168=NEW`, which the policy docs recommend for speed) FetchContent unconditionally redirects the download to `build/_deps/googletest-tmp`, the cache directory stays empty, and offline builds silently regress. Reproduced.

Pin it visibly with `cmake_policy(SET CMP0168 OLD)` and a comment next to the existing policy lines, or add a `cmake_policy(GET CMP0168 ...)` guard that warns when the cache directory is set but will be ignored.

CMakeLists.txt
```diff
@@ -131,12 +131,23 @@ target_sources(
...
+          "$ENV{HOME}/.cache/cmake-fetchcontent")
+    endif()
+
     # SYSTEM marks the dependencies' include directories as system headers so
     # that clang-tidy and compiler warnings do not report on third-party code.
     FetchContent_Declare(
       googletest
-      GIT_REPOSITORY https://github.com/google/googletest.git
-      GIT_TAG v1.18.0
+      URL https://github.com/google/googletest/archive/refs/tags/v1.18.0.tar.gz
```
philipcraig: The tag-archive URL caches under the generic name `v1.18.0.tar.gz` in a directory the PR intends to be shared, so any other project (or a second dependency here) fetching a `v1.18.0` tag into the same cache hash-mismatches, deletes, and re-downloads on every alternation. It also pins the hash of an on-the-fly archive that GitHub only guarantees byte-stable for a year, with a recommendation to use release assets instead.

The release asset `https://github.com/google/googletest/releases/download/v1.18.0/googletest-1.18.0.tar.gz` has the identical SHA256, so swapping the URL line fixes both with no other change. Alternatively keep the URL and add `DOWNLOAD_NAME googletest-1.18.0.tar.gz`.

CONTRIBUTING.md
```diff
@@ -233,10 +233,15 @@ plain shell in the container.
 | `--cache-volumes`, `--no-cache-volumes` | Mount the persistent cache volumes (default: on). |
 | `-v`, `--verbose` | Enable verbose logging. |
 
-The `uv` package cache and Bazel's downloaded dependency archives persist
-across sandbox runs in Docker named volumes, `cc-protocol-uv-cache` and
-`cc-protocol-bazel-repository-cache`. These grow without bound since neither
-tool prunes them; to reset one, run `docker volume rm <name>`.
+These persist across sandbox runs in Docker named volumes:
```
philipcraig: "These persist" now has no antecedent, since it follows the options table. Suggest "The following caches persist across sandbox runs in Docker named volumes:".

As written, the cache is documented only as a sandbox volume, but every CMake build now writes to `~/.cache/cmake-fetchcontent`, and `./scripts/cmake.sh --clean` does not touch it. If the directory becomes opt-in per the `CMakeLists.txt` comment this paragraph is enough; otherwise the CMake build section needs a sentence naming the directory and how to relocate or reset it.

philipcraig (changes requested):
The cache works in the sandbox, but deriving the directory from `$HOME` unconditionally turns three environments that used to configure fine into hard failures. Making the directory opt-in fixes all of them at once; the other comments are independent.

CMakeLists.txt
jbcoe: Reverted.

CMakeLists.txt
jbcoe: Reverted.

CMakeLists.txt
```diff
@@ -131,6 +131,19 @@ target_sources(
 if(XYZ_PROTOCOL_IS_NOT_SUBPROJECT)
 
   if(${BUILD_TESTING})
+    set(XYZ_CMAKE_FETCHCONTENT_DOWNLOAD_CACHE_DIR
+        "$ENV{XYZ_CMAKE_FETCHCONTENT_DOWNLOAD_CACHE_DIR}"
+        CACHE PATH "Persistent FetchContent cache directory")
+
+    if(XYZ_CMAKE_FETCHCONTENT_DOWNLOAD_CACHE_DIR)
+      set(FETCHCONTENT_BASE_DIR
```
philipcraig: `FETCHCONTENT_BASE_DIR` is not a download location. FetchContent puts `googletest-src`, `googletest-subbuild` and `googletest-build` under it, and `FetchContent_MakeAvailable` does `add_subdirectory(googletest-src googletest-build)`, so every build directory in the sandbox now compiles gtest into the same objects and the same `libgtest.a`. Two trees with different flags (`cmake.sh` and `cmake.sh --asan -B asan`, or `--clang-tidy` with libc++) rebuild gtest against each other on alternate builds, and two concurrent builds link against a half-written archive. The shared subbuild also records the generator in its own `CMakeCache.txt`, so a configure with a different generator fails until the volume is removed.

Caching only the read-only part avoids all of this. Bake the checkout into the image (`git clone --depth 1 --branch v1.18.0`, the pattern already used for bazelisk) and point `FETCHCONTENT_SOURCE_DIR_GOOGLETEST` at it: no volume, no lock, and each tree keeps its own gtest build. The first draft's `URL` + `DOWNLOAD_DIR` also kept the build tree per build directory.

philipcraig (commented):
Second pass on the reworked head. Moving from a download cache to `FETCHCONTENT_BASE_DIR` changes what the volume holds: it is now the git clone, the populate subbuild and googletest's build tree, shared by every build directory in the sandbox. Details inline.

One thing outside the diff: `scripts/consteval_coverage.py` still globs `build/*/_deps/googletest-src` for the gtest headers, so in the sandbox `./scripts/cmake.sh build` followed by `./scripts/consteval_coverage.sh` now fails with "googletest headers not found under build/*/_deps". It should read `googletest_SOURCE_DIR` from the build's `CMakeCache.txt` instead.

CMakeLists.txt
```diff
@@ -131,6 +131,19 @@ target_sources(
...
+          "${XYZ_CMAKE_FETCHCONTENT_DOWNLOAD_CACHE_DIR}")
+
+      file(MAKE_DIRECTORY "${FETCHCONTENT_BASE_DIR}")
+      # LOCK stops concurrent configures deleting each other's in-progress fetch.
+      file(LOCK "${FETCHCONTENT_BASE_DIR}/.lock" GUARD PROCESS)
```
philipcraig: `GUARD PROCESS` is released when configure exits, so the build step, which is what writes `googletest-build`, is unguarded. `cmake.sh --clean` runs `--clean-first`, which deletes the other trees' gtest outputs. Without `TIMEOUT` a second configure behind a stalled one waits forever with no output; `file(LOCK ... TIMEOUT 60 RESULT_VARIABLE r)` fails fast.

CMakeLists.txt
```diff
@@ -131,6 +131,19 @@ target_sources(
 if(XYZ_PROTOCOL_IS_NOT_SUBPROJECT)
 
   if(${BUILD_TESTING})
+    set(XYZ_CMAKE_FETCHCONTENT_DOWNLOAD_CACHE_DIR
```
philipcraig: `CACHE PATH` seeded from the environment is captured on the first configure of a build directory only. `build/` is bind-mounted into both the devcontainer (env unset) and the sandbox (env set), so a tree first configured in one keeps that value in the other: either the sandbox re-clones googletest into `build/asan/_deps`, or the devcontainer writes gtest into `~/.cache/cmake-fetchcontent` on the host.

The plain `set(FETCHCONTENT_BASE_DIR)` on line 139 also shadows any `-DFETCHCONTENT_BASE_DIR` a developer passes, and a relative value creates the cache inside the checkout, which neither `.gitignore` nor `.bazelignore` covers. `scripts/cmake.py` already turns `XYZ_PROTOCOL_CLANG_P2996_DIRECTORY` into `-D` flags; passing `-DFETCHCONTENT_SOURCE_DIR_GOOGLETEST=` from there needs no `CMakeLists.txt` change and matches how the Bazel cache lives in `~/.bazelrc` rather than workspace config.

CONTRIBUTING.md
```diff
@@ -233,10 +233,15 @@ plain shell in the container.
...
-`cc-protocol-bazel-repository-cache`. To reset either volume, run `docker volume
-rm <name>`.
+Caches persist across sandbox runs in Docker named volumes:
+
+- `cc-protocol-uv-cache`: the `uv` package cache.
+- `cc-protocol-bazel-repository-cache`: Bazel's downloaded dependency archives.
+- `cc-protocol-cmake-fetchcontent-cache`: CMake `FetchContent`'s downloaded
+  dependency archives.
+
+None of the tools prune their cache, so each grows without bound; to reset
```
philipcraig: googletest is declared with `GIT_REPOSITORY`, so the volume holds a git clone, the populate subbuild and the compiled gtest build tree of whichever tree last built, not "downloaded dependency archives". `uv cache prune` and `uv cache clean` exist. Worth saying that `docker volume rm` on this volume invalidates every existing build tree's gtest outputs until each is reconfigured with `--clean`.

cmake/xyz_fetchcontent_declare_cached.cmake
```diff
@@ -0,0 +1,102 @@
+include_guard(GLOBAL)
```
jbcoe, line 1: I'm adding this function now so that adding later packages (like benchmarking) does not become more complicated.

CONTRIBUTING.md
jbcoe: I'm happy to omit this. The sandbox containers are ephemeral and the cache volume cannot be removed while any of them are live.

CMakeLists.txt
jbcoe: Outdated so marking as resolved.

CMakeLists.txt
jbcoe: `FETCHCONTENT_BASE_DIR` was the wrong approach. I've introduced `xyz_fetchcontent_declare_cached` that sets the right variable `FETCHCONTENT_SOURCE_DIR_GOOGLETEST`.

scripts/cmake.py
```diff
@@ -161,6 +161,8 @@ def log(msg: Any) -> None:
         f"-DENABLE_TSAN={'ON' if args.tsan else 'OFF'}",
         f"-DENABLE_COVERAGE={'ON' if args.coverage else 'OFF'}",
         f"-DCLANG_TIDY_ENABLE={'ON' if args.clang_tidy else 'OFF'}",
+        "-DXYZ_CMAKE_FETCHCONTENT_CACHE_DIR="
```
jbcoe, line 171: This is set unconditionally to overwrite CMakeCache.txt content if `XYZ_CMAKE_FETCHCONTENT_CACHE_DIR` becomes unset on a later run (this would currently require manual running of CMake).

scripts/consteval_coverage.py
```diff
@@ -329,17 +330,33 @@ def _on_close_brace(self) -> None:
 
 
 def find_googletest_include_directories() -> list[str]:
-    """Locate googletest headers under an existing CMake build directory."""
-    pattern = os.path.join(SOURCE_ROOT, "build", "*", "_deps", "googletest-src")
-    for googletest_source in sorted(glob.glob(pattern)):
+    """
```
jbcoe, line 333: I suspect that this can be cleaned up but is good enough for now.

cmake/xyz_fetchcontent_declare_cached.cmake
```diff
@@ -0,0 +1,102 @@
...
+      TIMEOUT 60
+      RESULT_VARIABLE XYZ_LOCK_RESULT)
+    if(NOT XYZ_LOCK_RESULT EQUAL 0)
+      message(
+        FATAL_ERROR
+          "Could not lock ${XYZ_CMAKE_FETCHCONTENT_CACHE_DIR}: ${XYZ_LOCK_RESULT}"
+      )
+    endif()
+
+    if(NOT EXISTS "${XYZ_SOURCE_DIR}/CMakeLists.txt")
```
philipcraig, line 74: The only reuse gate is `EXISTS CMakeLists.txt`, and the clone writes straight into the final path, so an interrupted clone poisons the volume for every later sandbox. Killed before `CMakeLists.txt` is checked out: the directory is non-empty, the next configure re-runs `git clone` into it, git exits 128 with `destination path already exists`, and every configure fails with `Failed to clone` until someone removes the directory. Killed after it (git checks out in index order, so `CMakeLists.txt` precedes `googlemock/` and `googletest/`): the gate passes forever and every build fails on `gtest/gtest.h` with no hint that the cache is the cause.

Clone into `${XYZ_SOURCE_DIR}.tmp` after `file(REMOVE_RECURSE)` of any leftover, then `file(RENAME)` into place, and gate on a stamp written after the rename rather than on a file git happens to write first.

cmake/xyz_fetchcontent_declare_cached.cmake
```diff
@@ -0,0 +1,102 @@
...
+  if(NOT XYZ_GIT_REPOSITORY)
+    message(FATAL_ERROR "GIT_REPOSITORY parameter must be supplied")
+  endif()
+  if(NOT XYZ_GIT_TAG)
+    message(FATAL_ERROR "GIT_TAG parameter must be supplied")
+  endif()
+
+  if(XYZ_CMAKE_FETCHCONTENT_CACHE_DIR)
+    string(TOUPPER "${XYZ_NAME}" XYZ_NAME_UPPER)
+    set(XYZ_SOURCE_DIR "${XYZ_CMAKE_FETCHCONTENT_CACHE_DIR}/${XYZ_NAME}-src")
```
philipcraig, line 57: The checkout is keyed on `NAME` only. With `FETCHCONTENT_SOURCE_DIR_<NAME>` set FetchContent skips the download and update steps entirely, so once a volume holds `googletest-src` at v1.18.0 a later `GIT_TAG` bump is ignored by every populated sandbox: no clone runs, no message is printed, and the sandbox keeps building the old version while CI and the devcontainer build the new one. `--rebuild-docker` does not help because the clone lives in the volume, not the image.

Either put the tag in the path (`${XYZ_NAME}-${XYZ_GIT_TAG}-src`) or record `GIT_REPOSITORY` and `GIT_TAG` in the stamp file and re-clone when they differ.

cmake/xyz_fetchcontent_declare_cached.cmake
```diff
@@ -0,0 +1,102 @@
...
+
+    file(MAKE_DIRECTORY "${XYZ_CMAKE_FETCHCONTENT_CACHE_DIR}")
+    # Guards the clone below; each build tree's compiled output stays
+    # private, so nothing else here touches shared state.
+    file(
+      LOCK "${XYZ_CMAKE_FETCHCONTENT_CACHE_DIR}/${XYZ_NAME}.lock"
+      GUARD PROCESS
```
philipcraig, line 64: Now that `TIMEOUT 60` is in, the lock needs to cover only the clone. `GUARD PROCESS` releases at cmake exit, and the lock is taken before the `EXISTS` check, so it is held through compiler probes, googletest's configure and generation. `XYZ_CMAKE_FETCHCONTENT_CACHE_DIR` is set container-wide, so `./scripts/cmake.sh build` alongside the clang-tidy pre-commit hook (which configures a second tree) or `cmake.sh --asan -B asan` blocks the second configure on `googletest.lock` and, past 60 s, fails it with `Could not lock` even though the checkout it needs is already on disk.

`GUARD FUNCTION` (or an explicit `file(LOCK ... RELEASE)` after the clone) plus an `EXISTS` check before taking the lock fixes both.

cmake/xyz_fetchcontent_declare_cached.cmake
```diff
@@ -0,0 +1,102 @@
...
+    if(NOT XYZ_LOCK_RESULT EQUAL 0)
+      message(
+        FATAL_ERROR
+          "Could not lock ${XYZ_CMAKE_FETCHCONTENT_CACHE_DIR}: ${XYZ_LOCK_RESULT}"
+      )
+    endif()
+
+    if(NOT EXISTS "${XYZ_SOURCE_DIR}/CMakeLists.txt")
+      execute_process(
+        COMMAND git clone --depth 1 --branch "${XYZ_GIT_TAG}"
```
philipcraig, line 76: `git clone --depth 1 --branch` accepts only branch and tag names; with a commit hash it fails with `Remote branch <sha> not found in upstream origin`. The docstring says `GIT_TAG` is "the tag, branch, or commit to check out", and FetchContent itself recommends hashes, so a SHA-pinned dependency configures in CI and the devcontainer and FATAL_ERRORs only in the sandbox. A branch name also freezes at first clone.

`git init` + `git fetch --depth 1 <repo> <GIT_TAG>` + `git checkout FETCH_HEAD` handles all three forms; otherwise narrow the docstring to tags.

cmake/xyz_fetchcontent_declare_cached.cmake
```diff
@@ -0,0 +1,102 @@
...
+
+  ``SYSTEM``
+    Forwarded to FetchContent_Declare; marks the dependency's include
+    directories as system headers.
+
+#]=======================================================================]
+function(xyz_fetchcontent_declare_cached)
+  set(options SYSTEM)
+  set(oneValueArgs NAME GIT_REPOSITORY GIT_TAG)
+  cmake_parse_arguments(XYZ "${options}" "${oneValueArgs}" "" ${ARGN})
```
philipcraig, line 44: `XYZ_UNPARSED_ARGUMENTS` is never checked or forwarded, so any `FetchContent_Declare` option the wrapper does not enumerate (`GIT_SHALLOW TRUE`, `EXCLUDE_FROM_ALL`, `PATCH_COMMAND`) is silently dropped. Dropping `SYSTEM` from `options` and writing the single call `FetchContent_Declare(${XYZ_NAME} GIT_REPOSITORY ${XYZ_GIT_REPOSITORY} GIT_TAG ${XYZ_GIT_TAG} ${XYZ_UNPARSED_ARGUMENTS})` forwards `SYSTEM` and everything else, and removes the duplicated `if/else` below.

cmake/xyz_fetchcontent_declare_cached.cmake
```diff
@@ -0,0 +1,102 @@
...
+  endif()
+  if(NOT XYZ_GIT_TAG)
+    message(FATAL_ERROR "GIT_TAG parameter must be supplied")
+  endif()
+
+  if(XYZ_CMAKE_FETCHCONTENT_CACHE_DIR)
+    string(TOUPPER "${XYZ_NAME}" XYZ_NAME_UPPER)
+    set(XYZ_SOURCE_DIR "${XYZ_CMAKE_FETCHCONTENT_CACHE_DIR}/${XYZ_NAME}-src")
+
+    file(MAKE_DIRECTORY "${XYZ_CMAKE_FETCHCONTENT_CACHE_DIR}")
```
philipcraig, line 59: `file(MAKE_DIRECTORY)` is redundant: `file(LOCK)` creates intermediate directories and the lock file itself, and `git clone` creates missing parents. The error message also names the cache directory rather than `${XYZ_NAME}.lock`, so with a second dependency a timeout does not say which lock was contended.

philipcraig (commented):
Third pass on the `xyz_fetchcontent_declare_cached` head. The shared checkout is the right shape; the remaining points are about what happens to the volume when a clone is interrupted, a tag changes, or two configures overlap. Two replies on the earlier `CMakeCache.txt` and `consteval_coverage.py` threads.

CMakeLists.txt
philipcraig: The unconditional `-D` in `cmake.py` does not cover this. `set(FETCHCONTENT_SOURCE_DIR_GOOGLETEST ... PARENT_SCOPE)` is a normal variable, and FetchContent's own `set(... CACHE PATH ...)` for the same name then seeds the cache from it, so `FETCHCONTENT_SOURCE_DIR_GOOGLETEST` lands in `CMakeCache.txt` as a persistent entry. Reconfiguring with the env var unset clears `XYZ_CMAKE_FETCHCONTENT_CACHE_DIR` but leaves that entry behind, and if the cache directory is gone configure fails with `Manually specified source directory is missing` until `--clean`. A tree first configured in the sandbox then reconfigured in the devcontainer hits this. `set(FETCHCONTENT_SOURCE_DIR_${XYZ_NAME_UPPER} "${XYZ_SOURCE_DIR}" CACHE PATH "" FORCE)` in the cached branch, and the same with an empty value in an `else()`, keeps it in step with the variable.

scripts/cmake.py
philipcraig: One gap with the `-D` route: `cmake --preset Debug`, `cmake --fresh`, and any direct cmake call in the sandbox never see it and re-clone into that tree's `_deps`, and `-DBUILD_TESTING=OFF` now warns that the variable was not used because its only consumer is inside `if(${BUILD_TESTING})`. If either matters, `set(XYZ_CMAKE_FETCHCONTENT_CACHE_DIR "$ENV{XYZ_CMAKE_FETCHCONTENT_CACHE_DIR}" CACHE PATH "...")` in the module, as `FindClangTidy.cmake` does for `XYZ_PROTOCOL_CLANG_P2996_DIRECTORY`, covers every entry point and drops this hunk. Combined with the `CACHE PATH FORCE` change on the other thread it also stays correct when the env var goes away.

scripts/consteval_coverage.py
philipcraig: `googletest-distribution_SOURCE_DIR` is a side effect of googletest's `project()` name rather than a documented entry. This PR's module already sets the documented one: `FETCHCONTENT_SOURCE_DIR_GOOGLETEST:PATH=` is non-empty exactly when the shared checkout is in use, and otherwise the source is `FETCHCONTENT_BASE_DIR/googletest-src`. Reading those two keeps the script working if googletest renames its project, and a line loop with `startswith` removes the whole-file read and the `re` import.

jbcoe:
I'm happy to handle interrupted clones, tag changes and overlapping configures as follow up PRs. The sandbox setup is a functioning hack and does not need to be robust to pathological behaviour. I need an offline setup for use on a long haul flight, not a production ready workflow.

We should make the cache robust to Google test being upgraded. The rest we can probably let slide.

jbcoe:
FetchContent does not appear to provide useable support for caching that will avoid me manually writing lock and stamp files. Anything simple I can put together will have edge cases that I do not have bandwidth to exhaustively test.

I'd rather focus on C++ than CMake build infrastructure. Closing this PR.

philipcraig:
Agreed on the direction, and sorry for the grind on this one: every point in the last pass came from the volume being shared and writable, and baking the checkout into the image removes all of them at once. With the source in the image there is nothing to lock, stamp or invalidate, and a tag bump is a Dockerfile edit plus `--rebuild-docker`, which is already the documented sync step for `.bazelversion`.

I don't think it needs an offline flag or a second image. With `FETCHCONTENT_SOURCE_DIR_GOOGLETEST` set FetchContent performs no download or update step, so the baked checkout costs nothing when online and the sandbox can use it unconditionally:

- `docker/Dockerfile` sandbox stage: `git clone --depth 1 --branch v1.18.0 https://github.com/google/googletest.git /home/vscode/.cache/googletest-src` and `ENV FETCHCONTENT_SOURCE_DIR_GOOGLETEST=/home/vscode/.cache/googletest-src`.
- `scripts/cmake.py`: forward it as `-DFETCHCONTENT_SOURCE_DIR_GOOGLETEST=` + `os.environ.get(..., "")`, unconditionally, so a build directory shared with the devcontainer picks up the right value on every configure.
- `scripts/consteval_coverage.py`: the CMakeCache lookup from this PR.

No `CMakeLists.txt` change and no volume for CMake; Bazel keeps the repository cache from #304. The one rule is that bumping the googletest tag needs `--rebuild-docker`. `docker run --network none` on the image is a quick way to prove the build is offline-capable end to end.

PR to follow.

jbcoe:
> Agreed on the direction, and sorry for the grind on this one: every point in the last pass came from the volume being shared and writable, and baking the checkout into the image removes all of them at once. With the source in the image there is nothing to lock, stamp or invalidate, and a tag bump is a Dockerfile edit plus `--rebuild-docker`, which is already the documented sync step for `.bazelversion`.
> 
> I don't think it needs an offline flag or a second image. With `FETCHCONTENT_SOURCE_DIR_GOOGLETEST` set FetchContent performs no download or update step, so the baked checkout costs nothing when online and the sandbox can use it unconditionally:
> 
> * `docker/Dockerfile` sandbox stage: `git clone --depth 1 --branch v1.18.0 https://github.com/google/googletest.git /home/vscode/.cache/googletest-src` and `ENV FETCHCONTENT_SOURCE_DIR_GOOGLETEST=/home/vscode/.cache/googletest-src`.
> * `scripts/cmake.py`: forward it as `-DFETCHCONTENT_SOURCE_DIR_GOOGLETEST=` + `os.environ.get(..., "")`, unconditionally, so a build directory shared with the devcontainer picks up the right value on every configure.
> * `scripts/consteval_coverage.py`: the CMakeCache lookup from this PR.
> 
> No `CMakeLists.txt` change and no volume for CMake; Bazel keeps the repository cache from #304. The one rule is that bumping the googletest tag needs `--rebuild-docker`. `docker run --network none` on the image is a quick way to prove the build is offline-capable end to end.
> 
> PR to follow.

Should we do the same for UV and bazel cache? We can then dispense with Docker volumes altogether.

PR #316, opened 2026-09-08
Bump clang-p2996 to bde8bcf3 to pick up the define_aggregate qualifier fix

philipcraig:
Moves `.github/clang-p2996-commit` from 060be176 to bde8bcf3, the `p2996` commit that squash-merged bloomberg/clang-p2996#351 (upstream #349). Clang Tidy now accepts a `define_aggregate` member type spelled as a qualified name under a pointer, such as `^^const std::type_info*`, so #306 can drop its `type_info_pointer` alias.


jbcoe (approved):
Not much to review but intent LGTM. Thanks for this!

PR #319, opened 2026-09-08
Save the Bazel disk cache on every CI run

### LINKED ISSUE #298: Save the Bazel disk cache on every CI run
`bazel-contrib/setup-bazel`'s `disk-cache` saves only when its exact key misses, and that key hashes Bazel build files rather than sources. In PR #293's runs the sanitizer disk cache froze at its first contents: after `protocol.hh` changed on main, every later run reported the same 33 disk-cache hits and 9 recompiles. A fully warm rerun took 16 to 21 s against 61 to 77 s partially warm and 85 to 92 s cold.

Manage the cache with `actions/cache` directly: key on `github.sha` with a restore-keys prefix so each run restores the newest cache and saves its own, point `--disk_cache` at that directory, and bound its size with Bazel's disk-cache garbage-collection flags (`--experimental_disk_cache_gc_max_size`). Applies to the Bazel test and Sanitizers workflows.

philipcraig:
setup-bazel's `disk-cache` saves only when its key, a hash of the BUILD files, misses, so the cache froze at its first contents and every later run recompiled the same files.

Manage the directory with `actions/cache` instead: each Bazel and Sanitizers job restores the newest cache for its compiler or sanitizer and saves under its own commit. Bazel's disk-cache garbage collector, run as soon as the test command returns, keeps each cache under 512M.

### Measurements

`Run Bazel` step time and Bazel's own process summary, for the latest scheduled runs on main ([Bazel](https://github.com/jbcoe/cc-protocol/actions/runs/34190576328), [Sanitizers](https://github.com/jbcoe/cc-protocol/actions/runs/34191927160)), this PR's first run with no cache to restore, and a rerun of the same commit that restored what the first run saved ([Bazel](https://github.com/jbcoe/cc-protocol/actions/runs/34272039168), [Sanitizers](https://github.com/jbcoe/cc-protocol/actions/runs/34272039209)).

| Job | main, setup-bazel cache | This PR, cold | This PR, warm |
|---|---|---|---|
| GCC trunk | 78 s, 42 sandboxed, 0 hits | 68 s, 42 sandboxed, 0 hits | 17 s, 42 hits |
| GCC-16 | 50 s, 42 sandboxed, 0 hits | 59 s, 42 sandboxed, 0 hits | 18 s, 42 hits |
| asan | 122 s, 42 sandboxed, 0 hits | 107 s, 42 sandboxed, 0 hits | 16 s, 42 hits |
| tsan | 105 s, 42 sandboxed, 0 hits | 92 s, 42 sandboxed, 0 hits | 15 s, 42 hits |

The main runs did restore setup-bazel's cache, but it yielded no hits: its contents were compiled by an earlier GCC snapshot, and `scripts/bazel.py` stamps the compiler version into every action key. The saved caches are 2 MB (GCC trunk and GCC-16), 16 MB (asan) and 10 MB (tsan) compressed; restore and save each take about a second, and the garbage-collector wait under two.

Closes #298

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -141,10 +141,46 @@ jobs:
         uses: bazel-contrib/setup-bazel@0.19.0
         with:
           bazelisk-cache: true
-          disk-cache: ${{ github.workflow }}
+          # The disk cache is not setup-bazel's `disk-cache`: that saves only
+          # when its key, a hash of the BUILD files, misses, so it froze at its
+          # first contents and every later run recompiled the same files. The
+          # restore and save steps below manage the directory instead, and
+          # Bazel's garbage collector bounds what each run carries forward.
+          # The collector is an idle task, so a zero delay lets it run as soon
+          # as the test command returns.
+          bazelrc: |
+            build --disk_cache=~/.cache/bazel-disk
+            build --experimental_disk_cache_gc_max_size=512M
+            build --experimental_disk_cache_gc_idle_delay=0
+      - name: Restore the Bazel disk cache
+        if: needs.detect-changes.outputs.relevant != 'false'
+        id: disk-cache
+        uses: actions/cache/restore@v4
+        with:
+          path: ~/.cache/bazel-disk
+          # Every run saves under its own commit and restores the newest cache
+          # with a matching prefix, usually the previous run's. One cache per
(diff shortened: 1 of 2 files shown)


jbcoe (approved):
Thanks for this. Faster CI and a cooler planet. 🌍

PR #320, opened 2026-09-08
Test implicit protocol_view conversion at call sites

### LINKED ISSUE #270: Consider non-explicit constructors for protocol_view<T> and protocol_view<const T>
Currently this won't compile:

```
TEST(ReflectionProtocolViewTest, ViewOfProtocolPassedByValue) {
  struct Interface {
    int get() const;
  };

  struct Conforming {
    int value;

    int get() const { return value; }
  };

  auto read = [](protocol_view<const Interface> view) { return view.get(); };

  protocol<Interface> p(Conforming{5});
  EXPECT_EQ(read(p), 5);
}
```

I think that this should compile. Implicit construction of a view type is safe.

philipcraig:
#272 made `protocol_view<T>` and `protocol_view<const T>` implicitly constructible, which is what #270 asks for, and the issue's own test landed in #244. This adds the sibling call sites: a mutable view from a `protocol` lvalue, and both views from a conforming object lvalue, each passed by value to a function.

Closes #270

### DIFF
--- a/protocol_test.cc
+++ b/protocol_test.cc
@@ -2523,4 +2523,47 @@ TEST(ReflectionProtocolViewTest, ViewOfProtocolPassedByValue) {
   EXPECT_EQ(read(p), 5);
 }
 
+TEST(ReflectionProtocolViewTest, MutableViewOfProtocolPassedByValue) {
+  struct Interface {
+    int get() const;
+    void update(int value);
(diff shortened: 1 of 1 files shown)


jbcoe (approved):
These are simple to understand and instructive.

Thanks for this.

PR #321, opened 2026-09-08
Add a CMake install step and a package consumer check

### LINKED ISSUE #132: Add install step in CMakeLists
We should probably wait until the full C++ implementation is submitted to do this.

philipcraig:
Closes #132.

Installs `protocol.hh`, `name_mangling.h`, the licence and a CMake package exporting `xyz_protocol::protocol`, so downstream projects can use

```cmake
find_package(xyz_protocol REQUIRED)
target_link_libraries(my_target PRIVATE xyz_protocol::protocol)
```

The reflection flags move from directory-wide options onto the target as usage requirements, written as compiler-id generator expressions so the installed package hands the right flags to whichever compiler the consumer builds with. Only the targets that link the library now get `-freflection`; the plain tutorials no longer do. The stale `xyz_value_types-config.cmake.in`, copied from value_types, becomes `xyz_protocol-config.cmake.in`.

`./scripts/cmake.sh install` installs into `<build-dir>/install`, then configures, builds and runs the small downstream project in `install_test/` against that package. That project sets neither a C++ standard nor a reflection flag, so it only builds if the exported target carries both. The CMake CI job runs it after the test step in every matrix entry.


protocol.hh
```diff
@@ -473,12 +473,19 @@ struct member_base_generator {
     std::meta::info thunk_type = substitute(
         ^^member_thunk, {^^member_base, ^^ProtocolType, ^^Vtable, ^^Specs...});
 
+    // clang-p2996 deprecates data_member_options::no_unique_address in
+    // favour of a fork-specific attributes member that GCC does not have,
+    // so the warning is silenced here rather than moving off the standard
+    // field.
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
```
jbcoe: Can we avoid doing this in source? It's cognitive load in already complex code.

If not, can we make it file-wide?

CONTRIBUTING.md
```diff
@@ -47,6 +47,22 @@ build and test with it, execute:
 
 For more detailed CMake options, run `./scripts/cmake.sh --help`.
 
+Passing `install` installs the headers and a CMake package into
+`<build-dir>/install`, then configures, builds and runs the downstream
+project in `install_test/` against that package. This is how the CMake CI
+job checks the package. Arguments the script does not recognise, such as
```
jbcoe: Please cut "This is how the CMake CI job checks the package.".

jbcoe: Please cut "This is how the CMake CI job checks the package.".

protocol.hh
philipcraig: We of course could do that. Isn't it better to have this at as narrow a scope as we can though?

jbcoe (approved):
Thanks for this.

I'm not sure that I understand why a header-only library needs RPATH handling but the rest of this looks good.

I made some very minor comment-reduction comments.

protocol.hh
philipcraig: Moved to file scope in 08eb36b: push after the includes, pop before the header guard closes. It can't go away entirely because the installed package no longer passes `-Wno-deprecated-declarations`, so a Clang consumer would see the warning from `no_unique_address`.

CONTRIBUTING.md
philipcraig: Cut in 08eb36b.

philipcraig:
The rpath is for running `install_test`, not for the header. The library is header-only, but the consumer binary still links the compiler's libstdc++ or libc++, and both CI compilers keep theirs outside ld.so's search path: GCC trunk in `/opt/gcc-latest/lib64`, and the clang-p2996 toolchain's libc++ in `~/clang-p2996-toolchain/lib/x86_64-unknown-linux-gnu`.

Without it, a trunk-built binary that uses any GCC 16+ symbol resolves `libstdc++.so.6` to the distro copy and fails at load with `version 'GLIBCXX_3.4.36' not found` (Ubuntu 26.04's libstdc++ stops at 3.4.35). Under Clang the loader cannot find `libc++.so.1` at all. The main build embeds the same directory as `CMAKE_BUILD_RPATH` for the test executables; `install_test` is a plain consumer with no rpath of its own, and an imported interface target cannot carry one, so `cmake.py` passes the directory when it configures the consumer.

jbcoe:
I see. It's a test executables problem.

PR #323, opened 2026-09-09
Add target_type

### LINKED ISSUE #314: Add way to access the typeid of a protocol
`std::any` has `myAny.type()` to access the `std::type_info` of whatever object is being stored in `std::any` at the time. Should we add something similar for `protocol` and `protocol_view`? If so, should it be a member or free function? A member function could conflict with interface functions, but a free function breaks precedent with the STL.

RyanJK5:
Closes #314

### DIFF
--- a/protocol_test.cc
+++ b/protocol_test.cc
@@ -2718,10 +2718,10 @@ TEST(ReflectionProtocolTest, TargetType) {
   EXPECT_EQ(target_type(p), typeid(OtherType));
 
   auto _ = std::move(p);
-  // NOLINTBEGIN(bugprone-use-after-move): the test exercises the moved-from
-  // state on purpose.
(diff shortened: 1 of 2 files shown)


protocol_test.cc
```diff
@@ -33,7 +33,7 @@ concept has_update = requires(P& p) { p.update(0); };
 template <typename P>
 concept has_get_value = requires(P& p) { p.get_value(); };
 
-// Concepts for overloaded `get`/`get(int)` negative tests.
+// Concepts for overloaded `g et`/`get(int)` negative tests.
```
philipcraig: This comment has a typo "g et" but also doesn't actually seem to apply to any thing or any concepts?

philipcraig (approved):
very nice. minor nit question on the comment in the test

jbcoe (approved):
Thanks for this!

PR #330, opened 2026-09-10
Bake the googletest checkout into the sandbox image

philipcraig:
Follow-up to #315.

The sandbox image clones googletest at the tag pinned in `CMakeLists.txt` and sets `FETCHCONTENT_SOURCE_DIR_GOOGLETEST` to it. With that variable set FetchContent performs no download or update step, so CMake builds in the sandbox need no network, and each build tree still compiles its own copy. The checkout is image content: nothing to lock, stamp or invalidate. Bumping the tag is a Dockerfile edit plus `--rebuild-docker`, as for `.bazelversion`.

`cmake.py` forwards the variable on every configure, empty outside the sandbox, so a build directory shared between the sandbox and the devcontainer picks up the right value each time. `consteval_coverage.py` checks the same variable before the build tree's own `_deps`.

No `CMakeLists.txt` change and no new volume.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -253,6 +253,10 @@ in Docker named volumes, `cc-protocol-uv-cache` and
 `cc-protocol-bazel-repository-cache`. To reset either volume, run `docker volume
 rm <name>`.
 
+The sandbox image also carries a googletest checkout at the tag pinned in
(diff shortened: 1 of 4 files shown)


jbcoe (approved):
Tested locally. Thanks for this.

PR #332, opened 2026-09-10
Add docker-in-docker to the devcontainer

philipcraig:
Adds the docker-in-docker devcontainer feature. Codespaces has no host Docker socket to share, so the feature runs its own daemon inside the devcontainer. That makes `./scripts/agentic-sandbox.sh --rebuild-docker` runnable from the devcontainer, so a `docker/Dockerfile` change such as #330 can be built and checked before it lands.

Needs a devcontainer rebuild to take effect.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -5,6 +5,12 @@
 		"context": "..",
 		"target": "base"
 	},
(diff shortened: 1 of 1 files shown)


jbcoe (approved):
Looks good.

Do we need docker adding to the docker image for this to be useable?

PR #347, opened 2026-09-12
Remove EnclosingType from operator_thunk as it is not used

jbcoe:


philipcraig:
clang-tidy fails here because the three `friend struct detail::operator_thunk` declarations in `protocol`, `protocol_view` and the const view still list `typename EnclosingType`, so clang-p2996 rejects them as redeclarations with too many template parameters. GCC accepts the mismatch silently, which is why the other jobs are green. The same failure carries through #348 to #354.

#355 already removes `typename EnclosingType,` from those three declarations, and that is the whole fix. Could that change move down into this PR? Otherwise clang-tidy stays red on the stack, and on main from the moment this lands until #355 does, since clang-tidy is not a required check.

The `readability-implicit-bool-conversion` findings on `FAIL()` in forwarding_test.cc are a knock-on of the compile error and disappear with it.

philipcraig (approved):

PR #355, opened 2026-09-12
Split detail into logical parts

jbcoe:


conformance.hh
```diff
@@ -0,0 +1,205 @@
...
+// The named, non-special member functions and call operators of `Type`,
+// static or not, in declaration order.
+template <std::meta::info Type>
+consteval auto conformance_candidate_infos() {
+  auto named =
+      members_of(Type, std::meta::access_context::unprivileged()) |
+      std::views::filter(std::meta::is_function) |
+      std::views::filter([](std::meta::info member) consteval {
+        return has_identifier(member) ||
+               is_operator<std::meta::operators::op_star>(member) ||
```
philipcraig, line 146: This admits every member `operator*` and `operator->`, but `operator_thunk` only specialises the nullary `R(*)()` form for `op_star` and `op_arrow`. An interface with a binary `operator*` compiled on main (it was never a candidate) and now fails deep inside `operator_overload_set`:

```cpp
struct Vec { Vec operator*(double) const; int size() const; };
protocol<Vec> p(Impl{});
// operator_thunks.hh:270: error: 'operator*' has not been declared in
//   operator_thunk_t<overload_spec<^^Vec::operator*(double) const, true>, ...>
```

`fn_ptr_type()` is `Vec(*)(double)`, which matches no partial specialisation and lands on the primary template, so the `delete("Unspecialized operator thunk...")` message is never reached. Either generalise the `op_star`/`op_arrow` thunks to `R(*)(Args...)` as `op_parentheses` and `op_square_brackets` are, or throw from `protocol_interface_function_infos` next to the ref-qualified check so the error surfaces at the interface.

vtable.hh
```diff
@@ -0,0 +1,173 @@
...
+                protocol_interface_functions_of<^^T>) {
+    constexpr std::meta::info vtable_member =
+        find_vtable_entry<^^vtable_t<T>, member>();
+    using FnPtrType = typename[:type_of(vtable_member):];
+    if constexpr (is_const(member)) {
+      constexpr std::meta::info candidate =
+          find_conforming_member<member, ^^U>();
+      vtable.[:vtable_member:] = &const_view_trampoline<FnPtrType, U,
+                                                        candidate>::operator();
+    } else /*constexpr*/ {
```
philipcraig, line 156: `make_view_vtable` lost the `const_only` skip, so `protocol_view<const T>` now takes the address of a `mutable_view_trampoline` for every non-const interface member. That instantiates the bodies of `U`'s non-const members, which the conformance check never looks at:

```cpp
struct I { int get() const; void set(int); };
template <class X> struct W {
  int get() const { return 0; }
  void set(int v) { x_.set(v); }
  X x_;
};
struct NoSet {};
const W<NoSet> w{};
protocol_view<const I> v(w);  // main: ok
                              // branch: error: 'struct NoSet' has no member named 'set'
```

It also emits one address-taken trampoline per non-const member in every const-view TU. The comment above documents the simplification, but the old behaviour was observable, not just a code-size choice.

protocol_wrappers.hh
```diff
@@ -0,0 +1,191 @@
...
+  consteval {
+    // clang-format off
+    std::meta::info thunk_type = substitute(
+        ^^member_function_overload_set, {^^type, ^^ProtocolType, ^^Vtable, ^^OverloadSpecs...});
+
+    define_aggregate(
+      ^^type, {data_member_spec(thunk_type,
+                             std::meta::data_member_options{
+                              .name = identifier_of(Member),
+                              .no_unique_address = true
```
philipcraig, line 100: This is now the only use of `data_member_options::no_unique_address`, and this header has no `-Wdeprecated-declarations` guard. `protocol.hh` includes it at line 56, before its own pragma push at line 62, so with clang-p2996 any consumer of the installed headers gets:

```
protocol_wrappers.hh:94:32: warning: 'no_unique_address' is deprecated: Use .attributes = { ^^[[no_unique_address]] } [-Wdeprecated-declarations]
```

CI does not see it because `CMakeLists.txt` adds `-Wno-deprecated-declarations` in-tree only. Meanwhile the pragma pair was copied into `member_function_thunks.hh` and `operator_thunks.hh`, neither of which names `data_member_options`. One push/pop around this `define_aggregate` would do, and the three other copies can go.

operator_thunks.hh
```diff
@@ -0,0 +1,276 @@
...
+template <std::meta::operators Operator, typename ProtocolType, typename Vtable,
+          typename... OverloadSpecs>
+struct operator_overload_set {
+  operator_overload_set() =
+      delete ("Unspecialized operator overload set cannot be instantiated");
+};
+
+// operator()
+template <typename ProtocolType, typename Vtable, typename... OverloadSpecs>
+struct operator_overload_set<std::meta::operators::op_parentheses, ProtocolType,
```
philipcraig, line 237: Every `operator_thunk` specialisation and `operator_overload_set` has public defaulted special members, so the anti-slicing from #215 is lost for operators:

```cpp
struct I { int operator()(int) const; };
protocol_view<I> v(c);
template <std::meta::operators Op, class P, class V, class... S>
int take(xyz::detail::operator_overload_set<Op, P, V, S...> s) { return s(1); }
take(v);  // main: rejected (protected copy constructor)
          // branch: compiles; s(1) casts a non-view to ProtocolType* and reads garbage
```

`member_function_overload_set` keeps the private-defaulted + `friend` pattern (`member_function_thunks.hh:142`). The operator sets need the same, and `MemberThunksCannotBeDetached` in `protocol_test.cc` should cover them.

protocol_traits.hh
```diff
@@ -0,0 +1,67 @@
...
+#include <meta>
+#include <ranges>
+#include <type_traits>
+
+namespace xyz::reflection {
+
+template <typename I>
+concept is_valid_interface =
+    is_class_type(^^I) && std::same_as<I, std::remove_cvref_t<I>> &&
+    std::ranges::none_of(
```
philipcraig, line 34: `std::ranges::none_of` needs `<algorithm>`. This header, and `member_function_thunks.hh` and `operator_thunks.hh` which include it first, fail when included on their own with both GCC trunk and clang-p2996. All three are installed via the CMake `FILE_SET`, so a consumer including `protocol_traits.hh` for `is_protocol_v` hits it. `protocol.hh` only works because it includes `<algorithm>` first.

vtable.hh
```diff
@@ -0,0 +1,173 @@
...
+      ^^const std::type_info*, {
+                                   .name = "xyz_protocol_typeid"}));
+
+  template for (constexpr std::meta::info member :
+                protocol_interface_functions_of<interface_type>) {
+    // Build the function-pointer type R(*)(void*, Args...) noexcept(...)
+    // from the method's return type, parameter types and noexcept-ness; a
+    // const method takes `const void*` instead, matching the constness of
+    // the access path it's called through.
+    std::vector<std::meta::info> fn_args{dealias(return_type_of(member))};
```
philipcraig, line 74: `fn_ptr_t` lost its `noexcept` parameter, so vtable entries no longer carry `noexcept`: for `int f(int) const noexcept` the entry type is now `int(*)(const void*, int)` where main had `int(*)(const void*, int) noexcept`. The comment above still promises `noexcept(...)`, the trampolines' `bool Noexcept` parameter is now always `false`, and each `noexcept(IsNoexcept)` thunk calls through a potentially-throwing pointer, so GCC wraps the indirect call in a terminate-on-throw region instead of a tail jump. Either keep a `noexcept`-carrying alias for the vtable specs, or drop the dead `Noexcept` parameter from both trampolines and fix the comment.

conformance.hh
```diff
@@ -0,0 +1,205 @@
...
+template <std::meta::info Type>
+consteval std::vector<std::meta::info> protocol_interface_function_infos() {
+  std::vector<std::meta::info> result;
+  for (std::meta::info member : conformance_candidates_of<Type>) {
+    if (is_static_member(member)) continue;
+    if (is_lvalue_reference_qualified(member) ||
+        is_rvalue_reference_qualified(member)) {
+      std::string name = has_identifier(member)
+                             ? std::string(identifier_of(member))
+                             : "operator()";
```
philipcraig: The hard-coded `"operator()"` is now wrong for `operator[]`, `operator*` and `operator->`: a ref-qualified `operator[]` reports `ref-qualified member function 'operator()' is not supported`, naming a member the interface does not have. `display_string_of(member)` or a switch on `operator_of(member)` gives the right name.

operator_thunks.hh
```diff
@@ -0,0 +1,276 @@
...
+          bool IsConst, bool IsNoexcept>
+struct operator_thunk {
+  operator_thunk() =
+      delete ("Unspecialized operator thunk cannot be instantiated");
+};
+
+// operator()
+template <typename R, typename... Args, typename ProtocolType, typename Vtable,
+          std::meta::info Member, bool IsConst, bool IsNoexcept>
+struct operator_thunk<std::meta::operators::op_parentheses, R (*)(Args...),
```
philipcraig, line 50: The eight-line call body appears ten times in this file, `fn_ptr_type()` is duplicated byte-for-byte with `member_function_thunks.hh`, and the supported-operator set `{(), [], *, ->}` is spelled out independently here, in `conformance.hh:145`, `name_mangling.h:244` and `protocol_wrappers.hh:151` (whose `if` chain silently falls through for an unknown operator). Any change to the call protocol (the valueless handling in #284, ref-qualifier forwarding in #356) has to land in ten places, and a fifth operator means editing four files in lockstep, with a different unrelated error for each omission. One shared call helper that the four operator declarators forward to, one consteval `{op, mangling code, name}` table in `name_mangling.h` next to `builtin_type_code`, and `reflect_constant(operator_of(member))` in `protocol_wrappers.hh` (as line 218 here already does) would collapse most of it.

philipcraig (commented):
Review of the whole stack (#336 → #355) as one diff against main, at f78871d. Inline comments are on this PR because it holds the final form of every file. Two points fall outside this PR's diff:

- `scripts/consteval_coverage.py` only instruments `protocol.hh`, which keeps 3 of its 21 consteval sites after the split. The `consteval coverage` job now measures almost nothing, and Codecov reports a pass because it compares percentages. The script needs to iterate over the new headers.
- `name_mangling.h:255` encodes every `operator*` as `de` (unary dereference). A binary `operator*` is `ml`. See the comment on `conformance.hh:146`.

conformance.hh
jbcoe: We will fix this in a follow-up PR.

vtable.hh
jbcoe: I'm happy with the vtable changes as they simplify code and the old const view vtable had spurious entries. The member-function forwarding logic will ensure that functions cannot be called.

The bug above likely existed before but with a more obscure set of member functions being needed.

We can fix this in a follow up PR.

operator_thunks.hh
jbcoe: This file will soon-enough be generated from Python. 

We'll address duplication in a follow-up PR.

vtable.hh
jbcoe: I'll tackle this regression as follow-up work. The reflection-based compile errors I ran into from `fn_ptr_t`'s support for `noexcept` were too time consuming to debug.

jbcoe:
Thanks for the detailed review.

I've added some follow-up PRs to the stack and fixed other issues as they were introduced.

I'll tackle the noexcept `fn_ptr` challenge as follow-up (#365).

conformance.hh
jbcoe: We'll forbid binary `operator*` for now. Tackled in a follow-up PR #367

protocol_wrappers.hh
```diff
@@ -0,0 +1,198 @@
...
+  switch (ConstPolicy) {
+    case const_policy::propagate:
+      // `protocol<T>` propagates const through forwarded member function calls
+      // so supports const and non-const qualified member functions.
+      return true;
+    case const_policy::const_only:
+      // `protocol_view<const T>` forwards only const-qualified member
+      // functions.
+      return is_const(member);
+    case const_policy::all_const:
```
philipcraig, line 66: Under `all_const` the idiomatic const/non-const operator pair with differing return types is kept as two overloads and both are made const, so every use of the three new operators on `protocol_view<I>` is ambiguous:

```cpp
struct I {
  const X& operator[](int) const;
  X& operator[](int);
};
protocol_view<I> mv(c);
mv[0];   // error: ambiguous overload for 'operator[]' (operand types are 'protocol_view<I>' and 'int')
*mv;     // same, for operator* / operator-> pairs
```

`protocol<I>` and `protocol_view<const I>` are fine. The rule predates this stack for named members, but `operator[]`, unary `operator*` and `operator->` are almost always written in exactly this shape, so `protocol_view<I>` cannot be used with them. The new operator tests only use pairs that both return `int`. Either drop the const overload when a non-const one matches ignoring const and return type, or document the limitation and add a differing-return-type test either way.

operator_thunks.hh
```diff
@@ -0,0 +1,269 @@
...
+                      ProtocolType, Vtable, Member, IsConst, IsNoexcept> {
+  static constexpr std::meta::info vtable_entry =
+      find_vtable_entry<^^Vtable, Member>();
+
+  R operator()(Args... args) noexcept(IsNoexcept)
+    requires(!IsConst)
+  {
+    auto* protocol_object = static_cast<ProtocolType*>(this);
+    if constexpr (xyz::reflection::is_protocol_v<ProtocolType>) {
+      assert(!protocol_object->valueless_after_move() &&
```
philipcraig, line 60: This assert now lives in eight operator bodies, but the only death tests (`MutableValuelessCall`, `ConstValuelessCall`) call `p.foo()`. On main the call operator shared the named-member thunk, so those tests covered `p(...)` transitively; here they do not. Dropping or mis-conditioning any of the eight copies leaves CI green, which is the mistake #337 made for the named thunk and the death tests caught. `EXPECT_DEATH` for `p(...)`, `p[...]`, `*p` and `p.operator->()` on a moved-from protocol, const and non-const, closes that.

conformance.hh
philipcraig, line 146: Binary `operator*` gets a bespoke diagnostic in #367, but every other operator an interface declares is still silently dropped by this allowlist: `struct I { int operator+(int) const; }; protocol<I> p(C{});` compiles with no diagnostic and fails only at `p + 1`, and `explicit operator bool` likewise at the conversion site. The silent drop is pre-existing, but this stack turns the list into three hand-kept copies (here, `protocol_wrappers.hh:157`, `name_mangling.h:245`), and adding `op_plus` to this one alone yields "name mangling: unsupported operator" from `vtable.hh` and "can_substitute returned false" from `protocol_wrappers.hh`. One table of supported operators here that rejects everything else by name, plus `reflect_constant(operator_of(member))` in `protocol_wrappers.hh` as `operator_thunk_for` already does (drop-in, `protocol_test.cc` compiles unchanged), leaves a single place to extend.

protocol.hh
```diff
@@ -48,8 +48,13 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 #include <utility>
 #include <vector>
 
-#include "detail.hh"
+#include "conformance.hh"
+#include "member_function_thunks.hh"
 #include "name_mangling.h"
+#include "operator_thunks.hh"
+#include "protocol_traits.hh"
+#include "protocol_wrappers.hh"
```
philipcraig, line 56: The `-Wdeprecated-declarations` push/ignore/pop below (lines 62–63 and 701) is now dead: the only `data_member_options::no_unique_address` use is in `protocol_wrappers.hh:98`, which carries its own balanced pair, and clang-p2996 is clean with the three lines here deleted. Its only remaining effect is masking any future deprecation across the rest of this header on the clang-tidy job, which runs with `WarningsAsErrors: '*'`. `CMakeLists.txt:44` still says "see protocol.hh" for the same warning.

philipcraig (commented):
Second-round pass over the whole stack (#336 → #369) after the fixes pushed since 5187413403. The pragma move, the `<algorithm>` include and the ref-qualified diagnostic all work; the three deferred items (const-view laziness, #365, thunk duplication) are not repeated. Four new things inline.

philipcraig (approved):

protocol_wrappers.hh
jbcoe: Well-spotted. This needs fixing and is not specific to operators. #375

conformance.hh
jbcoe: We will handle this in follow up. Better error handling is about as complicated as supporting operators.

protocol.hh
jbcoe: Will fix in follow up #376

operator_thunks.hh
jbcoe: Will fix in #377

Issue #268, opened 2026-08-31
Should an unqualified interface member allow two dispatch paths?


RyanJK5:
For example:

```c++
struct Interface {
    int foo();
};

struct A {
    int foo() && { return 1; }
    int foo() & { return 2; }
};

protocol<Interface> p{A{}};
p.foo(); // 2.
std::move(p).foo(); // 1?
```

This would require us to generate two vtable entries for `int foo()`. Alternatively, we could say that `T` only conforms if it provides `int foo()` as one member function without overloading.

jbcoe:
I think I prefer `A` to be non-conforming (for now). We can relax this later at no cost.

PR #356, opened 2026-09-12
Add reference qualifier support

### LINKED ISSUE #97: Handle value category propagation

RyanJK5:
Closes #97 and #268

### DIFF
--- a/protocol.hh
+++ b/protocol.hh
@@ -141,16 +141,17 @@ consteval bool same_name_and_parameters(std::meta::info candidate,
 }
 
 // Returns `true` if the member functions `candidate` and `interface` have
-// the same name, reference qualifiers, de-aliased return type and de-aliased
-// parameter types; const and noexcept are not compared.
-consteval bool same_signature_ignoring_const(std::meta::info candidate,
-                                             std::meta::info interface) {
+// the same name, cvref qualifiers, de-aliased return type and de-aliased
+// parameter types; noexcept is not compared.
+consteval bool same_signature(std::meta::info candidate,
+                              std::meta::info interface) {
   if (is_lvalue_reference_qualified(interface) !=
       is_lvalue_reference_qualified(candidate))
     return false;
   if (is_rvalue_reference_qualified(interface) !=
       is_rvalue_reference_qualified(candidate))
     return false;
+  if (is_const(interface) != is_const(candidate)) return false;
   return same_name_and_parameters(candidate, interface);
 }
 
@@ -186,12 +187,8 @@ consteval bool member_function_conforms_to(std::meta::info candidate,
     // A static candidate has no object parameter, so it satisfies any const
     // or reference qualification of `interface`.
     if (!same_name_and_parameters(candidate, interface)) return false;
-  } else if (same_function_with_explicit_object(candidate, interface)) {
-    // No additional conditions.
-  } else {
-    if (!same_signature_ignoring_const(candidate, interface)) return false;
-    // `const` qualifiers must match.
-    if (is_const(interface) != is_const(candidate)) return false;
+  } else if (!same_function_with_explicit_object(candidate, interface)) {
+    if (!same_signature(candidate, interface)) return false;
   }
   // If interface is `noexcept`, `candidate` must be noexcept.
   return !is_noexcept(interface) || is_noexcept(candidate);
@@ -265,11 +262,11 @@ consteval std::meta::info find_vtable_entry() {
 }
 
 enum class member_options {
-  none = 1 << 0,
-  is_noexcept = 1 << 1,
-  is_const = 1 << 2,
-  is_lvalue = 1 << 3,
-  is_rvalue = 1 << 4
+  none = 0,
+  is_noexcept = 1 << 0,
+  is_const = 1 << 1,
+  is_lvalue = 1 << 2,
+  is_rvalue = 1 << 3
 };
 
 constexpr member_options operator|(member_options lhs, member_options rhs) {
@@ -500,13 +497,15 @@ consteval bool generates_wrapper_for(std::meta::info member,
     case member_policy::propagate:
       return true;
     case member_policy::const_only:
-      return is_const(member);
+      return !is_rvalue_reference_qualified(member) && is_const(member);
     case member_policy::all_const:
-      return !is_const(member) ||
-             std::ranges::none_of(members, [&](std::meta::info other) {
-               return !is_const(other) &&
-                      same_signature_ignoring_const(other, member);
-             });
+      return !is_rvalue_reference_qualified(member) &&
+             (!is_const(member) ||
+              std::ranges::none_of(members, [&](std::meta::info other) {
+                return !is_rvalue_reference_qualified(other) &&
+                       !is_const(other) &&
+                       same_name_and_parameters(other, member);
+              }));
   }
   std::unreachable();
 }
@@ -594,11 +593,6 @@ consteval std::meta::info generate_wrapper_bases() {
     for (std::meta::info member : members) {
       if (!same_name(member, first) ||
           !generates_wrapper_for<MemberPolicy>(member, members))
-        continue;
-
-      // protocol_view cannot hold rvalue qualified methods.
-      if (MemberPolicy != member_policy::propagate &&
-          is_rvalue_reference_qualified(member))
         continue;
 
       // clang-format off
--- a/protocol.hh
+++ b/protocol.hh
@@ -145,9 +145,6 @@ consteval bool same_name_and_parameters(std::meta::info candidate,
 // parameter types; noexcept is not compared.
 consteval bool same_signature(std::meta::info candidate,
                               std::meta::info interface) {
-  if (is_lvalue_reference_qualified(interface) !=
-      is_lvalue_reference_qualified(candidate))
-    return false;
   if (is_rvalue_reference_qualified(interface) !=
       is_rvalue_reference_qualified(candidate))
     return false;
--- a/protocol.hh
+++ b/protocol.hh
@@ -472,27 +472,31 @@ struct call_operator_base
   call_operator_base& operator=(call_operator_base&&) = default;
 };
 
-// How generated wrappers treat the const-qualification of interface members.
+// How generated wrappers treat the const- and ref-qualification of interface
+// members.
 enum class member_policy {
   // `protocol<I>`: as declared in `I`, so `const protocol<I>` exposes only the
-  // const member functions of `I` (const propagates).
+  // const member functions of `I` (const propagates). Reference qualification
+  // propagates similarly.
   propagate,
   // `protocol_view<I>`: every wrapper is const-qualified regardless of `I`
-  // (shallow const, as for `std::span`).
+  // (shallow const, as for `std::span`). Any rvalue qualified interface
+  // member function is ignored.
   all_const,
-  // `protocol_view<const I>`: only the const member functions of `I`.
+  // `protocol_view<const I>`: only the const member functions of `I`. Rvalue
+  // qualified member functions are ignored.
   const_only,
 };
 
-// Returns `true` if `ConstPolicy` generates a wrapper for `member`.
+// Returns `true` if `MemberPolicy` generates a wrapper for `member`.
 //
 // Under `all_const` every wrapper is const, so a const/non-const overload
 // pair `R f() const; R f();` would collide; the const overload is dropped as
 // a non-const reference to `I` would also resolve `f()` to `R f()`.
-template <member_policy ConstPolicy>
+template <member_policy MemberPolicy>
 consteval bool generates_wrapper_for(std::meta::info member,
                                      std::span<const std::meta::info> members) {
-  switch (ConstPolicy) {
+  switch (MemberPolicy) {
     case member_policy::propagate:
       return true;
     case member_policy::const_only:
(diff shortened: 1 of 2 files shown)


RyanJK5:
I made the following design decisions while implementing this PR. Happy to discuss/change any of them as appropriate. The tests cover these edge cases should we want to modify behavior.

- An unqualified interface function can match an lvalue-qualified member function, because the dispatch assumes the underlying type is an lvalue anyway unless otherwise specified. Therefore, an interface with `int foo()` will match a candidate function `int foo() &`. A candidate function `int foo() &&` will not conform, however.
- Lvalue- and rvalue-qualified overloads of the same function can exist in an interface. They will provide two different dispatch paths depending on the value category of `protocol`. Reference qualification can also be applied to `const` interface member functions (though there's probably little reason to do this).
- `protocol_view<I>` and `protocol_view<const I>` do NOT propagate value category. They will both only pull in unqualified interface members and lvalue-qualified interface members. Rvalue interface members are not included. Calling a member function from an rvalue `protocol_view` will always dispatch to the lvalue overload. This follows the precedent of `const` propagation.
- An unqualified or lvalue-qualified interface function will match functions with an lvalue-qualified explicit object parameter. It will NOT match functions with an rvalue-qualified explicit object parameter; an rvalue-qualified interface function will, however.
- `static` candidate functions will match regardless of the reference qualification of the interface function.

I tried to keep the architecture mostly intact, but refactored some things to avoid passing a million distinct booleans around everywhere. Any feedback is much appreciated!

protocol.hh
```diff
@@ -168,15 +166,27 @@ consteval bool same_function_with_explicit_object(std::meta::info candidate,
...
                             std::meta::type_of, std::meta::type_of);
 }
 
+// Returns `true` if the `candidate` member function is rvalue reference
+// qualified OR if it has an explicit object parameter that is rvalue reference
+// qualified.
+consteval bool is_called_with_rvalue(std::meta::info candidate) {
+  const auto params = parameters_of(candidate);
+  return is_rvalue_reference_qualified(candidate) ||
+         (!params.empty() && is_rvalue_reference_type(type_of(params.front())));
```
philipcraig: This treats any rvalue-reference-typed first parameter as an rvalue explicit-object parameter. For `struct Candidate { int f(int&& x) & { return x; } };` against interface `int f(int&&)`, conformance passes but `protocol<Interface> p(Candidate{});` fails inside `make_view_vtable` with "passing Candidate as this argument discards qualifiers" because the trampoline calls on `std::move(*ptr)`. Unqualified and static candidates with a `T&&` first parameter silently get the same move.

```suggestion
         (!params.empty() && is_explicit_object_parameter(params.front()) &&
          is_rvalue_reference_type(type_of(params.front())));
```

Alternatively drive the move off the interface member's qualifier, which `make_view_vtable` already has, and drop this helper.

protocol.hh
```diff
@@ -141,16 +141,14 @@ consteval bool same_name_and_parameters(std::meta::info candidate,
...
-consteval bool same_signature_ignoring_const(std::meta::info candidate,
-                                             std::meta::info interface) {
-  if (is_lvalue_reference_qualified(interface) !=
-      is_lvalue_reference_qualified(candidate))
-    return false;
+// the same name, cvref qualifiers, de-aliased return type and de-aliased
+// parameter types; noexcept is not compared.
+consteval bool same_signature(std::meta::info candidate,
+                              std::meta::info interface) {
   if (is_rvalue_reference_qualified(interface) !=
```
philipcraig, line 148: Only the `&&` qualifier is compared, so `&` is never checked in either direction and #268 is answered implicitly. With `struct A { int foo() & { return 2; } int foo() && { return 1; } };` against interface `int foo()`, `std::move(p).foo()` returns 2 while `std::move(a).foo()` returns 1. A `&`-only candidate also conforms to `int foo()` even though moving the concrete object is ill-formed, so the protocol is more permissive than the type. The comment above says "cvref qualifiers" and the test is named `RefQualifiersMatchExactly`, but `&` is not part of the rule.

Either pick one of the #268 options here (reject `&`-only candidates for unqualified members, or two vtable entries) and say so in this comment, the `member_policy::propagate` comment and DRAFT.md, or reword the comment and test name to state the rule as implemented.

protocol.hh
```diff
@@ -168,15 +166,27 @@ consteval bool same_function_with_explicit_object(std::meta::info candidate,
   if (params.empty() || !is_explicit_object_parameter(params.front()))
     return false;
 
-  if (is_const(interface) !=
-      is_const(remove_reference(type_of(params.front()))))
+  const auto obj_type = type_of(params.front());
+  if (is_const(interface) != is_const(remove_reference(obj_type))) return false;
+
+  if (is_rvalue_reference_qualified(interface) !=
+      is_rvalue_reference_type(obj_type))
```
philipcraig, line 173: The other direction is stricter than it needs to be. For interface `int take() &&`, an unqualified candidate `int take()` is rejected here and at line 148, and a by-value explicit-object `int take(this C self)` is rejected too, while the static branch accepts `static int take()`. Any ordinary type, including std types, therefore cannot satisfy an `&&` interface even though the trampoline's `std::move(*ptr).take()` would dispatch correctly. A callability-based rule (the candidate conforms if it is callable in every value category the interface member admits) fixes this and the `&` case above together.

protocol_test.cc
```diff
@@ -2773,4 +2791,148 @@ TEST(ReflectionProtocolViewTest, MovedFromTargetType) {
...
+  Conforming c{};
+  protocol_view<LvalueInterface> p(c);
+  EXPECT_EQ(p.foo(), 5);
+}
+
+TEST(ReflectionProtocolTest, OverloadedQualifiers) {
+  struct Interface {
+    int foo() &;
+    int foo() &&;
+    int foo() const;
```
philipcraig: `int foo() &; int foo() &&; int foo() const;` is ill-formed: [over.load]/2.3 requires either all or none of the overloads with the same parameter-type-list to have a ref-qualifier. GCC 15 rejects it ("cannot be overloaded"); GCC trunk and clang-p2996 accept it only because they miss the check. `ConstRefQualifiers` below uses `const&`, which is the legal shape.

```suggestion
    int foo() const&;
```

protocol.hh
```diff
@@ -668,26 +756,31 @@ consteval typename vtable_generator<T>::vtable make_view_vtable() {
...
-      constexpr std::meta::info candidate =
-          find_conforming_member<member, ^^U>();
-      result.[:vtable_member:] = &const_view_trampoline<FnPtrType, U,
-                                                        candidate>::call;
-    } else if constexpr (ConstPolicy != const_policy::const_only) {
-      constexpr std::meta::info candidate =
-          find_conforming_member<member, ^^U>();
-      result.[:vtable_member:] = &mutable_view_trampoline<FnPtrType, U,
+
+    // View vtables cannot use rvalue member functions.
```
philipcraig, line 761: The drop is silent. `is_valid_view_interface` still accepts an interface whose only members are `&&`-qualified, so `struct I { int f() &&; }; protocol_view<I> v(c);` compiles and `v` has no member `f` at all; the first diagnostic is "no member named f" at the call site. Either reject names that would vanish under `all_const`/`const_only` in `is_valid_view_interface` or `generate_wrapper_bases`, or state the rule on `protocol_view` and in DRAFT.md.

protocol.hh
```diff
@@ -499,12 +552,38 @@ using member_base_generator_t =
 template <typename... MemberBases>
 struct wrapper_bases : MemberBases... {};
 
+// Returns the appropriate const, noexcept, and reference qualification for the
+// member based on the const policy.
```
philipcraig, line 557: "const policy" is the old name; this now takes a `member_policy`.

protocol.hh
```diff
@@ -312,8 +332,8 @@ struct method_thunk<R (*)(Args...), EnclosingType, ProtocolType, Vtable, Member,
   // to the enclosing protocol/protocol_view object, then calls through its
   // stored vtable pointer's matching function pointer, passing the
   // viewed/owned object.
-  R operator()(Args... args) noexcept(IsNoexcept)
-    requires(!IsConst)
+  R operator()(Args... args) & noexcept(is_noexcept)
```
philipcraig, line 336: The four `operator()` bodies differ only in the `const` on the `enclosing()` cast, so the valueless assert, the vtable load and the splice call now have to change in four places. A private static

```cpp
template <typename P>
static R call(P* protocol_object, Args&&... args) noexcept(is_noexcept);
```

holding the body, with each qualified overload a one-line `return call(static_cast<ProtocolType*>(enclosing(this)), std::forward<Args>(args)...);`, passes the PR's test file and removes about thirty lines. A single deducing-this `operator()` is not a drop-in: under `all_const`, `p.foo()` becomes ambiguous between the `&` and `const&` thunks brought in by `member_thunk`'s using-declarations.

protocol.hh
```diff
@@ -272,28 +270,50 @@ consteval std::meta::info find_vtable_entry() {
                            std::string(name) + "'");
 }
 
+enum class member_options {
```
philipcraig, line 274: Two of these flags are derivable from `Member` (`is_noexcept`, and `is_const` under `propagate`), and encoding "unqualified" as `is_lvalue | is_rvalue` means every unqualified member and every view member now gets two constrained ref-qualified overloads where one unqualified or `const` `operator()` did before. On GCC trunk the unchanged pre-PR test corpus compiles about 10% slower with the new header, and a 40-member TU calling members in both value categories emits 180 thunk bodies instead of 100 at `-O0` (identical at `-O2`). The static members `is_noexcept`/`is_const` also shadow `std::meta::is_noexcept`/`is_const` inside `method_thunk`.

Passing `overload_spec<Member, member_policy>` and deriving the booleans inside `method_thunk`, with a three-state ref qualifier (none, `&`, `&&`), keeps the single `operator()` for unqualified and view members. GCC trunk accepts mutually exclusive constrained unqualified and ref-qualified overloads in one class, so the thunk does not need splitting.

philipcraig (commented):
Ran the tests and a few probes on GCC trunk. Inline comments cover the conformance rules, the trampolines and the thunk layout. Three things sit outside the diff:

- Stale comments: `protocol_interface_function_infos` (line 236) still says ref-qualified members are rejected and points at a TODO on `method_thunk` that this PR deletes; `is_protocol_conformant` (line 802) still explains why it calls the function instead of reading the variable in terms of a throw that no longer happens; `make_view_vtable` (line 739) still says `all_const` populates every entry.
- `conformance_check_rejects` in protocol_test.cc (line 407) has no callers left and its comment describes the old rejection. Deleting the `#ifdef __cpp_constexpr_exceptions` block with it. A positive test for the deleted static case would be worth keeping: `struct I { int take() &&; }; struct C { static int take(); };` conforms and dispatches through `protocol<I>`.
- Docs: the Limitations bullet in CONTRIBUTING.md (line 190) still says the generated wrapper does not apply the qualifier, and DRAFT.md says nothing about value-category propagation or about views omitting `&&` members, which #97 asked the draft to acknowledge.

philipcraig:
Thanks for listing the decisions, that made them easy to check. Three notes beyond the inline review.

**Decision 1 and #268.** On #268 @jbcoe preferred `A` to be non-conforming for now, since relaxing later is free and tightening later breaks users. This PR goes the other way: `int foo() &` conforms to `int foo()`, so `std::move(p).foo()` compiles on the protocol while `std::move(a).foo()` does not on the concrete type. Worth settling explicitly before this closes #268.

**One rule for all five.** The decisions fall out of a single statement: a candidate conforms if it is callable in every value category the interface member admits. That gives: unqualified interface members accept unqualified, `const`, static and by-value explicit-object candidates, and reject `&`-only and `&&`-only ones; `&&` interface members additionally accept unqualified candidates (which the current code rejects at line 148/173 while accepting static ones); `&` interface members accept `&`, unqualified, static and `this C&`. It also answers #97's request for a sentence in DRAFT.md.

**Legal overload sets.** [over.load]/2.3 forbids mixing ref-qualified and unqualified overloads of one name, so once an interface has `foo() &` it must spell the const overload `foo() const&`, not `foo() const`. GCC trunk and clang-p2996 do not diagnose this (GCC 15 does), which is why `OverloadedQualifiers` compiles. That is a user-facing constraint on interface shape and belongs next to the value-category text in DRAFT.md.

On views I agree with not propagating, though for a different reason than the const precedent: a view is a handle, and moving the handle says nothing about the viewed object, as with a moved `std::reference_wrapper`. The follow-on is that a name which vanishes entirely from `protocol_view<I>` should fail at instantiation rather than silently.

jbcoe:
The design decision here feels like the right one. I'm going to try to land my huge PR chain before merging this.

jbcoe:
If we did want r-value-qualified member functions called from views, we probably want `protocol_view<T&&>` but I'm not sure why anyone would want that.

jbcoe (changes requested):
This is good but needs rebasing onto #355 which I'm happy to do when it lands.

RyanJK5:
@jbcoe I think this has probably fallen far enough behind that it's worth doing over, but curious what you think.

jbcoe:
If you'd like to pair-program a painful manual rebase over coffee it could be instructive, or character building.

RyanJK5:
> If you'd like to pair-program a painful manual rebase over coffee it could be instructive, or character building.

That would certainly wake me up in the morning. Any preferred times?

jbcoe:
@RyanJK5 I'm sorry that we never got to rebase this face-to-face. Would you like me to rebase this PR on the plane?

RyanJK5:
> @RyanJK5 I'm sorry that we never got to rebase this face-to-face. Would you like me to rebase this PR on the plane?

No worries. If you're up for it, that would be really helpful.

jbcoe:
I've not managed to get to this yet but will take a look at the weekend.

PR #357, opened 2026-09-12
Bump the clang-p2996 pin to 0664c3f6

### LINKED ISSUE #350: Member splice through -> on an lvalue pointer skips the lvalue-to-rvalue conversion of the base (assertion "-> base must be a pointer prvalue")
**Describe the problem**

Accessing a data member through a splice with `->` when the base is an lvalue of pointer type trips an assertion in `Sema::BuildMemberExpr`:

```cpp
#include <meta>

struct Vtable {
  int (*entry)(int);
};

int use(const Vtable* vtable) {
  constexpr std::meta::info member = ^^Vtable::entry;
  return vtable->[:member:](1);
}
```

```
clang/lib/Sema/SemaExprMember.cpp:844: MemberExpr *clang::Sema::BuildMemberExpr(...):
Assertion `(!IsArrow || Base->isPRValue()) && "-> base must be a pointer prvalue"' failed.
...
#9  clang::Sema::BuildMemberExpr(clang::Expr*, bool, clang::SourceLocation, clang::NestedNameSpecifierLoc, ...)
#10 clang::Sema::BuildFieldReferenceExpr(clang::Expr*, bool, clang::SourceLocation, clang::CXXScopeSpec const&, clang::FieldDecl*, ...)
#11 clang::Sema::BuildMemberReferenceExpr(clang::Expr*, clang::QualType, clang::SourceLocation, bool, clang::CXXScopeSpec const&, ...)
#12 clang::Sema::BuildMemberReferenceExpr(clang::Scope*, clang::Expr*, clang::SourceLocation, clang::tok::TokenKind, clang::CXXSpliceExpr*)
#13 clang::Parser::ParsePostfixExpressionSuffix(clang::ActionResult<clang::Expr*, true>)
```

The same assertion fires from template instantiation (`TreeTransform<TemplateInstantiator>::TransformCallExpr` → the same `BuildMemberReferenceExpr` overload) when the splice operand is a template parameter. `vtable->entry(1)` and `(*vtable).[:member:](1)` are fine.

Observed with clang-p2996 060be17654102019e14810c3f948ef85a490755f, `-std=c++26 -freflection-latest -stdlib=libc++`, assertions enabled. A release build accepts the program and generates correct code, so this is likely benign in practice, but it makes every asserting build (and anything built on it, such as a debug clang-tidy or clangd) unusable on code that uses `->[: :]` with an lvalue pointer. GCC trunk accepts the program.

**Cause**

`Sema::BuildMemberReferenceExpr(Scope*, Expr* Base, SourceLocation, tok::TokenKind, CXXSpliceExpr*)` (SemaExprMember.cpp) checks that `Base` is a pointer when `->` is used but never applies `DefaultFunctionArrayLvalueConversion` (or `PerformMemberExprBaseConversion`) to it, unlike `Sema::ActOnMemberAccessExpr`, so the `MemberExpr` is built with an lvalue base.

**Environment**

- clang-p2996 at 060be17654102019e14810c3f948ef85a490755f, `-DLLVM_ENABLE_ASSERTIONS=ON`, Linux x86_64.

philipcraig:
Picks up bloomberg/clang-p2996#352 (merged 2026-09-12), which fixes bloomberg/clang-p2996#350: `ptr->[:member:]` on an lvalue pointer no longer trips the "-> base must be a pointer prvalue" assertion, so assertion-enabled builds of the toolchain can compile `protocol.hh`. Release builds, which CI uses, were unaffected, so there is no workaround to remove.

Part of #277.


jbcoe (approved):
Thanks for this

PR #363, opened 2026-09-12
Add macOS CI jobs for Bazel and CMake builds

jbcoe:
Adds macOS jobs to the bazel.yml and cmake.yml matrices. 

Install the reflection compiler via Homebrew's gcc@16 formula, since the Linux jobs' compiler sources aren't available on macOS. 

No sanitizer jobs are added as sanitizer support on MacOS is limited.

Note: macOS jobs should not be made blocking as provisioning macOS action-runners on GitHub can be slow.

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -81,6 +81,7 @@ jobs:
     if: ${{ !cancelled() }}
     name: ${{ matrix.settings.name }}
     runs-on: ${{ matrix.settings.os }}
+    continue-on-error: ${{ matrix.settings.compiler.type == 'GCC16_MACOS' }}
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -114,6 +114,21 @@ jobs:
                   std: 26,
                 },
             }
+          # Neither the GCC trunk snapshot (a .deb) nor Ubuntu's gcc-16
+          # package is available on macOS, so this job gets its reflection
+          # compiler from the gcc@16 Homebrew formula instead.
+          - {
+              name: "GCC-16 (macOS)",
+              os: macos-15,
+              compiler:
+                {
+                  type: GCC16_MACOS,
+                  version: 16,
+                  cc: "gcc-16",
+                  cxx: "g++-16",
+                  std: 26,
+                },
+            }
     steps:
       - uses: actions/checkout@v5
         if: needs.detect-changes.outputs.relevant != 'false'
@@ -130,6 +145,12 @@ jobs:
           sudo apt-get install -y --no-install-recommends gcc-16 g++-16
           echo "CXX=$(which ${{ matrix.settings.compiler.cxx }})" >> "$GITHUB_ENV"
           echo "CC=$(which ${{ matrix.settings.compiler.cc }})" >> "$GITHUB_ENV"
+      - name: Install GCC 16 (macOS)
+        if: needs.detect-changes.outputs.relevant != 'false' && matrix.settings.compiler.type == 'GCC16_MACOS'
+        run: |
+          brew install gcc@16
+          echo "CXX=$(brew --prefix gcc@16)/bin/${{ matrix.settings.compiler.cxx }}" >> "$GITHUB_ENV"
+          echo "CC=$(brew --prefix gcc@16)/bin/${{ matrix.settings.compiler.cc }}" >> "$GITHUB_ENV"
       - name: Install uv
         if: needs.detect-changes.outputs.relevant != 'false'
         uses: astral-sh/setup-uv@v10.0.1
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -107,6 +107,21 @@ jobs:
                   std: 26,
                 },
             }
+          # Neither the GCC trunk snapshot (a .deb) nor Ubuntu's gcc-16
+          # package is available on macOS, so this job gets its reflection
+          # compiler from the gcc@16 Homebrew formula instead.
+          - {
+              name: "GCC-16 (macOS)",
+              os: macos-15,
+              compiler:
+                {
+                  type: GCC16_MACOS,
+                  version: 16,
+                  cc: "gcc-16",
+                  cxx: "g++-16",
+                  std: 26,
+                },
+            }
     steps:
       - uses: actions/checkout@v5
         if: needs.detect-changes.outputs.relevant != 'false'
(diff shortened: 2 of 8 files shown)


.github/workflows/bazel.yml
```diff
@@ -114,6 +114,21 @@ jobs:
                   std: 26,
                 },
             }
+          # Neither the GCC trunk snapshot (a .deb) nor Ubuntu's gcc-16
+          # package is available on macOS, so this job gets its reflection
+          # compiler from the gcc@16 Homebrew formula instead.
+          - {
+              name: "GCC-16 (macOS)",
```
philipcraig: The PR body says the macOS jobs should not be made blocking, but this entry is part of the `build` matrix job that the `bazel` summary job depends on. `fail-fast: false` keeps the Linux legs running, but `needs.build.result` is still `failure` when only this leg fails, and `summary` cannot start until the slowest leg finishes. Today the ruleset requires only `pre-commit` and `uv-lock`, so nothing breaks yet; once #249 makes `bazel`/`cmake` required, a Homebrew bottle outage or a queued macOS runner blocks every C++ PR.

Two ways to make it advisory:

```yaml
continue-on-error: ${{ matrix.settings.compiler.type == 'GCC16_MACOS' }}
```

on `build` (a failed leg then reports `success` to `needs`), or a separate `build-macos` job that `summary` does not list in `needs`.

.github/workflows/cmake.yml
```diff
@@ -107,6 +107,21 @@ jobs:
                   std: 26,
                 },
             }
+          # Neither the GCC trunk snapshot (a .deb) nor Ubuntu's gcc-16
+          # package is available on macOS, so this job gets its reflection
+          # compiler from the gcc@16 Homebrew formula instead.
+          - {
+              name: "GCC-16 (macOS)",
```
philipcraig: Same as the bazel.yml comment: this leg gates the `cmake` summary job, and the Release/Debug cross-product makes it two macOS legs.

philipcraig (commented):
Three points outside the diff that this PR makes reachable.

**rpath probe hard-codes `libstdc++.so`** (`CMakeLists.txt:79`, `scripts/compiler_discovery.py:54`). On macOS, `g++-16 -print-file-name=libstdc++.so` echoes the name back unresolved. In CMake, `get_filename_component(REALPATH)` then resolves it against the source directory, so the repo root is appended to `CMAKE_BUILD_RPATH` of every test binary. The Python side guards with `os.path.exists`, so `bazel.py`/`cmake.py` set no rpath on macOS at all. The jobs are green because Homebrew's `libstdc++.6.dylib` carries an absolute install name; a relocated keg or self-built GCC fails at load time with no fallback. Using `libstdc++${CMAKE_SHARED_LIBRARY_SUFFIX}` (and the platform suffix in `_find_library_directory`) plus an `if(EXISTS ...)` guard fixes both.

**`CONTRIBUTING.md` prerequisites** (line 14) still name only the trunk snapshot or Ubuntu's `gcc-16` package, and the CI section (line 72) lists job names without the new `GCC-16 (macOS)` ones. A sentence on `brew install gcc@16` and the job names would keep it current.

**Silent sanitizer skip** (`cmake/sanitizers.cmake:13`). Each sanitizer target exists only if `check_cxx_compiler_flag` succeeds, with no message otherwise. On a macOS GCC without a Darwin sanitizer runtime, which is why this PR adds no macOS sanitizer legs, `./scripts/cmake.sh --asan` configures, builds and passes fully uninstrumented. A `message(FATAL_ERROR ...)` or at least `WARNING` when a requested sanitizer is unsupported would stop that, or a note beside the new matrix comment.

.github/workflows/bazel.yml
```diff
@@ -81,6 +81,7 @@ jobs:
     if: ${{ !cancelled() }}
     name: ${{ matrix.settings.name }}
     runs-on: ${{ matrix.settings.os }}
+    continue-on-error: ${{ matrix.settings.compiler.type == 'GCC16_MACOS' }}
```
philipcraig: Job-level `continue-on-error` doesn't make the leg advisory, it makes a failure invisible. A failed macOS run is recorded with conclusion `success` and only an annotation, so `needs.build.result` is `success`, `contains(needs.*.result, 'failure')` in `summary` is false, and the run, the PR checks and the scheduled-run notifications all show green. The first Homebrew gcc bottle that breaks reflection goes unnoticed.

It also doesn't address the slow-provisioning concern in the PR body: `summary` still has `needs: [detect-changes, build]`, so it can't start until the macOS legs finish. All the runs on this branch show `summary` starting after the macOS legs.

A separate `build-macos` job with no `continue-on-error` and not listed in `summary`'s `needs` gives what the PR body asks for: red when broken, never gating `bazel`/`cmake`.

.github/workflows/cmake.yml
```diff
@@ -73,6 +73,7 @@ jobs:
     if: ${{ !cancelled() }}
     name: ${{ matrix.settings.name }} ${{ matrix.configuration }}
     runs-on: ${{ matrix.settings.os }}
+    continue-on-error: ${{ matrix.settings.compiler.type == 'GCC16_MACOS' }}
```
philipcraig: Same as bazel.yml: this hides a macOS failure from `summary` and from the run conclusion, and `summary` still waits for the macOS legs. A separate `build-macos` job outside `summary`'s `needs` is the advisory-but-visible shape.

philipcraig (changes requested):
Thanks for the follow-up push. The `CMAKE_SHARED_LIBRARY_SUFFIX` probe with the `EXISTS` guard, the CONTRIBUTING update and the `FATAL_ERROR` on an unsupported sanitizer all address my earlier points.

The remaining smaller items from a pass over 2a4b241 are collected in #373 so they don't hold this PR up. The one thing I'd like changed here is the job-level `continue-on-error`, see the inline comments.

.github/workflows/bazel.yml
jbcoe: I added another job.

.github/workflows/cmake.yml
jbcoe: Fixed.

.github/workflows/bazel.yml
jbcoe: Fixed as suggested.

.github/workflows/cmake.yml
jbcoe: Fixed as suggested.

philipcraig (approved):
The separate `build-macos` jobs do what the PR body asks: they show up red when broken and are not in the `bazel`/`cmake` summary jobs' `needs`, so they never gate the required checks or delay them on runner provisioning.

The #373 commit looks right too: the Linux-only rpath gate with the callers already handling `None`, the `brew --prefix gcc@16` probe, the `foreach` in `cmake/sanitizers.cmake` probing the same list it applies, and the `GLOBAL` imported targets. Thanks.

PR #367, opened 2026-09-12
Reject binary operator* with a clear diagnostic

jbcoe:
Fixes @philipcraig 's review comment on conformance.hh:146 (PR #355): a binary `operator*` fell through to the unspecialised `operator_thunk` primary template instead of failing with a clear diagnostic. Full binary `operator*` support is deferred; this rejects it explicitly at the interface boundary instead.


conformance.hh
```diff
@@ -165,9 +165,21 @@ template <std::meta::info Type>
 constexpr inline auto conformance_candidates_of =
     std::define_static_array(conformance_candidate_infos<Type>());
 
+// The number of parameters `member` declares beyond its object parameter:
+// `parameters_of` includes an explicit object parameter (from "deducing
+// this"), so that one is excluded to count only the operands a caller
+// supplies.
+consteval size_t declared_argument_count(std::meta::info member) {
+  auto params = parameters_of(member);
+  if (!params.empty() && is_explicit_object_parameter(params.front())) {
```
philipcraig, line 174: This carve-out admits `operator*(this const I&)` past the guard below, but nothing downstream strips the object parameter (`operator_thunk_for::fn_ptr_type`, `generate_vtable_specs`, `same_signature_ignoring_const`), so the admitted form cannot work:

```cpp
struct I { int operator*(this const I&); };
sizeof(protocol<I>);  // operator_thunks.hh:288: error: 'operator*' has not been declared in
                      //   operator_thunk_t<overload_spec<^^I::operator*(this const I&), false>, ...>
```

Explicit-object interface members already fail conformance on main, so the branch is dead and the comment promises handling that does not exist. Counting `parameters_of(member).size()` and rejecting explicit-object operators alongside would be honest about what is supported.

philipcraig (commented):
Second-round pass over the whole stack (#336 → #369). The guard does what the title says for the ordinary binary form; two gaps inline.

conformance.hh
```diff
@@ -181,6 +193,11 @@ consteval std::vector<std::meta::info> protocol_interface_function_infos() {
       throw std::runtime_error("ref-qualified member function '" + name +
                                "' is not supported in a protocol interface");
     }
+    if (is_operator<std::meta::operators::op_star>(member) &&
```
philipcraig, line 196: Deleting these four lines leaves `protocol_test.cc` compiling, so the fix has no regression test. `conformance_check_rejects<>` already exists for the ref-qualified case; the same shape covers this:

```cpp
struct BinaryStarInterface { int operator*(int) const; };
struct BinaryStarByRef { int operator*(const BinaryStarByRef&) const; };
static_assert(conformance_check_rejects<BinaryStarInterface, Candidate>());
static_assert(conformance_check_rejects<BinaryStarByRef, Candidate>());
static_assert(!conformance_check_rejects<UnaryStarInterface, UnaryStarCandidate>());
```

under the same `__cpp_constexpr_exceptions` guard as the ref-qualified tests.

jbcoe: We'll clear this up when we support operator* fully. That's similar work to fixing this PR.

philipcraig (approved):

conformance.hh
jbcoe: We'll clean this up when we handle more operators.

PR #368, opened 2026-09-12
Restore anti-slicing guard on operator_overload_set

jbcoe:
Fixes @philipcraig 's review comment on operator_thunks.hh:243 (PR #355): `operator_overload_set` had public defaulted special members, so it could be sliced through a non-view pointer and read garbage.


operator_thunks.hh
```diff
@@ -238,6 +238,14 @@ struct operator_overload_set<std::meta::operators::op_parentheses, ProtocolType,
                              Vtable, OverloadSpecs...>
     : operator_thunk_t<OverloadSpecs, ProtocolType, Vtable>... {
   using operator_thunk_t<OverloadSpecs, ProtocolType, Vtable>::operator()...;
+
+ protected:
```
philipcraig, line 278: This seals `operator_overload_set`, but `operator_thunk` is the class whose bodies do `static_cast<ProtocolType*>(this)`, and it is still a public base with public implicit special members. `member_function_thunk` in this stack is sealed (`member_function_thunks.hh:82`), as `method_thunk` was on main; the four `operator_thunk` specialisations are not:

```cpp
struct I { int operator()(int) const; };
protocol<I> p(C{});
template <std::meta::operators Op, class F, class P, class V, std::meta::info M, bool K, bool N>
int detached(xyz::detail::operator_thunk<Op, F, P, V, M, K, N> t) { return t(1); }
detached(p);  // main: error, method_thunk(const method_thunk&) is protected
              // this PR: compiles; the 1-byte copy's operator() reads vtable_/object_ past its end
```

The same six-line block belongs in all four specialisations (lines 50, 87, 124, 159).

protocol_test.cc
```diff
@@ -47,6 +47,17 @@ concept is_callable = requires(P& p) { p(); };
 template <typename P>
 concept is_callable_with_int = requires(P& p) { p(0); };
 
+// A generic lambda's parameter type deduces `Op`, `P`, `V` and `S...` from
+// any `xyz::detail::operator_overload_set` base of `Derived`, so slicing can
+// be probed without naming the protocol's internal vtable/spec types.
+// `requires` treats the resulting inaccessible-special-member error as a
+// substitution failure rather than a hard error.
+template <typename Derived>
+concept operator_overload_set_can_be_sliced_from = requires(Derived d) {
```
philipcraig, line 56: Deducing the base through a generic lambda is ambiguous for any interface with two or more operator kinds: for `operator()` plus `operator[]` the call fails with "`operator_overload_set<Op, P, V, S...>` is an ambiguous base class", so the concept is `false` whether or not the guard exists. It holds for the single-operator interfaces used here, but any reuse on a richer interface passes vacuously. It also probes copy construction only: dropping the protected move constructor, destructor or either assignment leaves both new `static_assert`s green. `MemberThunksCannotBeDetached` asserts all six traits against the deduced overload-set type; the same shape works here.

philipcraig (commented):
Second-round pass over the whole stack (#336 → #369). This PR seals the overload sets, but the class that does the `this` cast is still open; details inline.

protocol_test.cc
jbcoe: The test ensures that slicing won't work, that's sufficient.

philipcraig (approved):

PR #369, opened 2026-09-12
Instrument all consteval headers in coverage script

jbcoe:
Fixes a point raised in philipcraig's review of the #336-#355 stack (out of diff on #355): consteval_coverage.py only instruments protocol.hh, which keeps 3 of 21 consteval sites after the detail.hh split, so the coverage job measures almost nothing.


jbcoe:
This check takes ~35 mins to run. I'm not happy with that and am comfortable losing coverage checks on compile time code.

@philipcraig WDYT?

philipcraig:
> This check takes ~35 mins to run. I'm not happy with that and am comfortable losing coverage checks on compile time code.
> 
> @philipcraig WDYT?

I don't like the time either and we can lose them

philipcraig:
> This check takes ~35 mins to run. I'm not happy with that and am comfortable losing coverage checks on compile time code.
> 
> @philipcraig WDYT?

I have a threading of this work idea -- stay tuned

philipcraig:
> > This check takes ~35 mins to run. I'm not happy with that and am comfortable losing coverage checks on compile time code.
> > @philipcraig WDYT?
> 
> I have a threading of this work idea -- stay tuned

oh! looks like from your last commit you have the same idea

philipcraig (commented):
Second-round pass over the whole stack (#336 → #369). Multi-header instrumentation works, and the per-shard LCOV output only lists that shard's probe points, so the Codecov merge across the matrix is sound. Two things inline.

On the run time: the sharding here may well be enough, and #374 collects the further ideas from this review (keying the trap on header and line so there is one work directory and one verify pass, running the verify compiles through the pool, probing the cheapest translation unit first). Either route needs a look once the timings from this PR's matrix are in, so I will re-review that part later rather than now.

scripts/consteval_coverage.py
```diff
@@ -516,45 +585,68 @@ def main() -> None:
...
+    # uninstrumented form: quote-include resolution is relative to the
+    # including file's own directory, so an instrumented header reached only
+    # transitively (e.g. name_mangling.h via protocol.hh) would otherwise
+    # resolve back to SOURCE_ROOT before reaching the target header's
+    # directory. Headers outside INSTRUMENTED_HEADERS still fall through to
+    # -I SOURCE_ROOT.
+    probe_jobs: list[tuple[ProbeConfiguration, ProbePoint]] = []
+    for header in shard_headers:
+        header_work_dir = os.path.join(arguments.work_dir, header)
+        os.makedirs(header_work_dir, exist_ok=True)
```
philipcraig, line 602: `exist_ok=True` only tolerates an existing directory. After a run on main, `build/consteval-coverage/protocol.hh` is a regular file written by the single-header layout, so the first run on this branch dies before any compile:

```
File "scripts/consteval_coverage.py", line 602, in main
FileExistsError: [Errno 17] File exists: '.../build/consteval-coverage/protocol.hh'
```

CI is unaffected (fresh checkout). A `shutil.rmtree(arguments.work_dir, ignore_errors=True)` before the loop fixes it.

scripts/consteval_coverage.py
```diff
@@ -61,7 +76,7 @@
 TRAP_ERROR_MARKER = "consteval_coverage_probe"
 
 # Translation units that instantiate protocols and so drive the consteval
-# machinery in protocol.hh.
+# machinery in the instrumented headers.
```
philipcraig, line 79: `name_mangling.h` is now instrumented (74 of 177 probe points) but `name_mangling_tests.cc` is not a default translation unit, and it is the only TU that reaches `signed_decimal`, `integral_decimal`, `mangle_template_argument_value` and the `is_value` branch. Against the three defaults those probes never fire, so 11 lines that `name_mangling_tests.cc` does exercise (`Tagged<3>`, `Tagged<-3>`, `Tagged64<4294967296LL>`) land as `DA:...,0` on Codecov. Adding it costs about two seconds per compile, and it is the cheapest TU, so probing it first also saves time.

philipcraig:
Measured the shards: the probe step is 9.3–11.8 min per shard against 36.3 min unsharded, at four times the runner minutes. Each shard still repeats 18–21 serial verify compiles, because round-robin sharding puts every header in every shard.

Let's land this as is, with the two inline fixes, and take the serial and per-probe costs in #374 afterwards. The numbers are on that issue.

philipcraig (approved):

jbcoe:
This is too slow and @philipcraig has a better PR Closing..

Issue #374, opened 2026-09-13
Cut the consteval coverage job's serial and per-probe costs


philipcraig:
The `consteval coverage` job is the slowest check in CI: jbcoe measured about 35 minutes on #369 before sharding, and #369 now splits the probe points across a four-way matrix to bring the wall time down. Each shard still pays a serial cost that sharding does not remove, and the probe phase itself can be made cheaper. Ideas, roughly in order of payoff:

**Key the trap on header and line, not line alone.** `trap()` takes a bare `__LINE__`, so the script needs one work directory per instrumented header, with a copy of every header and translation unit in each, and one unarmed verify compile per header per TU. `Instrumenter` already stores `self.header`, so it can emit `trap(<header index>, __LINE__)` and the armed build can pass a second `-D` for the header index. That gives one work directory, one set of copies and one verify pass, whatever the header count.

**Run the verify compiles through the pool.** `verify_unarmed_compile` runs on the main thread before the `ThreadPoolExecutor` starts, once per header in the shard. Probe points are round-robined, so every shard holds all seven headers and does 7 × 3 syntax-only compiles up front, against 3 on main; `protocol_test.cc` alone is roughly 14 s per compile on a fast machine, so that is a few minutes of single-threaded time on the critical path of each shard. Even without the trap change, submitting these to the same pool takes them off the critical path.

**Probe the cheapest translation unit first and stop at the first hit.** A probe point is covered as soon as one TU trips the trap. `name_mangling_tests.cc` compiles in about 2 s, `allocator_tests.cc` in about 2.3 s, `forwarding_test.cc` in 6 s and `protocol_test.cc` in 14 s. Ordering the TUs cheapest first and skipping the rest once a probe fires avoids most of the expensive compiles for the points that any TU covers. `name_mangling_tests.cc` also needs adding to `DEFAULT_TRANSLATION_UNITS` now that `name_mangling.h` is instrumented (see #369).

**Batch probe points per compile.** Arming one line per compile means one full front-end run per probe point. The trap could instead take a bitmap or a list of armed lines and record which ones fired, which means fewer, slightly slower compiles. Whether this helps depends on how many probe points fire per TU, so it needs measuring.

**Cache the unarmed precompiled state.** Every probe compile re-parses the standard library and googletest. A precompiled header for the unarmed prefix, or `-fsyntax-only` with a module cache once GCC supports it for this code, would cut the fixed cost per compile. This is the least certain of the ideas.

[...]

PR #378, opened 2026-09-13
Add DEATH tests for operators

### LINKED ISSUE #377: Add DEATH tests for operators

jbcoe:
Closes #377


protocol_test.cc
```diff
@@ -2358,6 +2359,176 @@ TEST(ReflectionProtocolTest, ConstValuelessCall) {
   // NOLINTEND(bugprone-use-after-move,hicpp-invalid-access-moved)
 }
 
+TEST(ReflectionProtocolTest, MutableCallOperatorValuelessDeath) {
```
philipcraig: These eight tests are copies of the two above with a single-operator `Interface` each, so this adds eight `protocol<Interface>` instantiations to the translation unit that the consteval-coverage job recompiles once per probe point. Locally that takes the per-probe `-fsyntax-only` compile from about 14.5 s to about 15.75 s, roughly two more minutes on the job #374 is trying to shrink, for tests that exercise a plain runtime `assert`.

Two interfaces would cover the same ground: one declaring the four mutable operators and one declaring the four const ones, each test doing one `std::move` and then four `EXPECT_DEATH` statements. `EXPECT_DEATH` forks per statement, so the parent's moved-from `p` is unchanged between them. That also collapses the ten copies of the NOLINT block and the death-message string.

protocol_test.cc
```diff
@@ -2313,7 +2313,8 @@ TEST(ReflectionProtocolTest, CallOperatorForwardingAfterCopy) {
   EXPECT_EQ(copy(21), 42);
 }
 
-// Tests that dispatching fails gracefully for a moved-from protocol.
+// Tests that dispatching fails gracefully for a moved-from protocol, for
+// named member functions and for each operator.
```
philipcraig, line 2317: Every test here goes through `protocol<I>`. The same operators called through a `protocol_view` built from a valueless protocol still dereference a null vtable entry with no diagnostic, because the thunks gate the assert on `is_protocol_v<ProtocolType>` (#284). Worth a pointer to #284 here so this section does not read as covering valueless dispatch in general.

protocol_test.cc
philipcraig: Nit: the two existing tests in this block are `...ValuelessCall`; these are `...ValuelessDeath`. Either suffix works, but one family name keeps `--gtest_filter=*Valueless*` variants honest.

philipcraig (approved):
Three small things, none blocking.

protocol_test.cc
jbcoe: We'll fix the duplication in due course. It's out of scope for now.

protocol_test.cc
jbcoe: We say `protocol` not `protocol_view`. I'm happy to leave this.

protocol_test.cc
jbcoe: Fixed

PR #379, opened 2026-09-13
Restrict use of deprecated-declarations pragma

### LINKED ISSUE #376: Review use of deprecated-declarations pragma

jbcoe:
Closes #376


protocol_wrappers.hh
```diff
@@ -30,12 +30,6 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 #include "operator_thunks.hh"
 #include "overload_spec.hh"
 
-// clang-p2996 deprecates data_member_options::no_unique_address in favour of
-// a fork-specific attributes member that GCC does not have, so the warning is
-// silenced for this file rather than moving off the standard field.
-#pragma GCC diagnostic push
```
philipcraig, line 36: Deleting this push leaves the file-scope `#pragma GCC diagnostic pop` at the bottom of the file (line 198 after this change) without a matching push; protocol.hh's trailing pop was removed correctly, this one was missed. Clang rejects it:

```
protocol_wrappers.hh:198:27: error: pragma diagnostic pop could not pop, no matching push [-Werror,-Wunknown-pragmas]
```

GCC accepts an unmatched pop silently, so the required jobs stay green and only clang-tidy goes red. Any downstream Clang consumer with `-Wall -Werror` fails too, and a consumer that wraps the include in its own push/pop has its push consumed here. Fix: delete the pop before `#endif`.

philipcraig (changes requested):
The narrowing is the right shape, but one stray pragma breaks the Clang build, and I think the deprecation itself is avoidable.

The warning only exists because we pass `-freflection-latest`, which switches on the fork's P3385 attribute-reflection extension; that is what deprecates `data_member_options::no_unique_address` in its `<meta>`. Replacing `-freflection-latest` with `-freflection -fparameter-reflection` at CMakeLists.txt:50 and :155 builds warning-free under `-Wall -Werror` on the pinned clang-p2996 with no pragmas at all, and ctest passes. Plain `-freflection` is not enough because the header uses `parameters_of`/`return_type_of` (P3096). Nothing else uses `-freflection-latest` (Bazel is GCC-only, consumers get the flags from the INTERFACE generator expression). That would let this PR delete both pragma blocks and the tutorial workaround instead of narrowing them.

If you would rather keep the umbrella flag, the tutorial could take a per-target `-Wno-deprecated-declarations` in tutorials/CMakeLists.txt so the teaching text stays free of fork-specific boilerplate.

philipcraig (approved):
awesome

PR #380, opened 2026-09-13
Use the host's git config in the agentic-sandbox

jbcoe:
`scripts/agentic-sandbox.py` now mounts the host's `~/.gitconfig` read-only into the sandbox so git picks up the host's name, email, aliases, credential helper and other settings directly. 

Commits made inside the sandbox are now attributed to the host user instead of coming out unset.


scripts/agentic-sandbox.py
```diff
@@ -164,6 +170,10 @@ def log(msg: str) -> None:
                 "sandbox",
                 "-t",
                 IMAGE_NAME,
+                "--build-arg",
+                f"GIT_USER_NAME={host_git_config('user.name')}",
```
philipcraig: Baking the identity into `cc-protocol-sandbox` ties it to whoever built the tag last, and the `image_exists` short-circuit above means it never refreshes without `--rebuild-docker`: a changed email, a per-repo `includeIf` identity, or a shared daemon all commit under the wrong name silently.

Git and jj both take identity from the environment, so this can be done at run time with no Dockerfile change. Keep the two `git config --get` reads, drop the `--build-arg`s, and add next to the `TERM`/`COLORTERM` pass-through:

```python
name = _host_git_config("user.name")
email = _host_git_config("user.email")
if name:
    for var in ("GIT_AUTHOR_NAME", "GIT_COMMITTER_NAME", "JJ_USER"):
        run_args.extend(["-e", f"{var}={name}"])
if email:
    for var in ("GIT_AUTHOR_EMAIL", "GIT_COMMITTER_EMAIL", "JJ_EMAIL"):
        run_args.extend(["-e", f"{var}={email}"])
```

The non-empty guard matters: an explicit empty `GIT_AUTHOR_NAME=` makes git refuse to commit. Reading the config from the repo's own directory at run time also means a repo-local identity is the one the sandbox uses, which is the right answer for the one repo it mounts.

docker/Dockerfile
```diff
@@ -122,6 +122,16 @@ ENV UV_CACHE_DIR="/home/vscode/.cache/uv"
 # https://git-scm.com/docs/git-config#Documentation/git-config.txt-safedirectory
 RUN git config --global --add safe.directory /workspace
 
+# Set git and jj commit identity from the host user who builds the image (see
+# scripts/agentic-sandbox.py, which passes these from the host's git config),
+# so commits made inside the container are attributed correctly.
+ARG GIT_USER_NAME
+ARG GIT_USER_EMAIL
+RUN if [ -n "$GIT_USER_NAME" ]; then git config --global user.name "$GIT_USER_NAME"; fi \
```
philipcraig: Two further reasons to prefer the run-time route, beyond staleness. `jj config set --user` parses the value as TOML: a name containing `[` fails the build outright, and a purely numeric name is stored as an integer, after which every later `jj` invocation fails with `invalid type: integer, expected a string` (checked with the pinned jj 0.43.0). And the `ARG`s sit before the bazelisk pre-warm and the googletest clone, so any identity change re-fetches both. With `-e` at run time this whole block goes away.

scripts/agentic-sandbox.py
```diff
@@ -156,6 +156,12 @@ def log(msg: str) -> None:
     )
     if args.rebuild_docker or not image_exists:
         log(f"--- Building Docker Sandbox: {IMAGE_NAME} ---")
+
+        def host_git_config(key: str) -> str:
```
philipcraig: Advisory. `subprocess.run` drops git's exit code and stderr here, so a malformed `~/.gitconfig` (`fatal: bad config line ...`, exit 128) silently produces an empty identity. Exit 1 means unset and should map to `""`; anything else should raise, as the `check_output` calls elsewhere in this file do. Since it captures nothing from `main`, it could also be a module-level `_host_git_config` beside `_seed_config_file`, which it needs to be if the run-args section calls it too.

philipcraig (changes requested):
One change requested: pass the identity at `docker run` time rather than baking it into the image (first comment below). The comment on `host_git_config` is advisory.

jbcoe:
Thanks for the review. I'll undo changes to the dockerfile.

scripts/agentic-sandbox.py
```diff
@@ -202,6 +202,9 @@ def log(msg: str) -> None:
 
     run_args.extend(cache_mounts)
     run_args.extend(_agent_mount_args(args.agent))
+    host_gitconfig = os.path.expanduser("~/.gitconfig")
+    if os.path.isfile(host_gitconfig):
+        run_args.extend(["-v", f"{host_gitconfig}:/home/vscode/.gitconfig:ro"])
```
philipcraig, line 207: Mounting over `~/.gitconfig` replaces the image's file, which is where the Dockerfile puts `safe.directory /workspace` (docker/Dockerfile:123), so a Linux host whose UID is not 1000 is back to `fatal: detected dubious ownership`. And because the mount is read-only, anything that writes global config inside the sandbox fails: `gh auth setup-git`, or an agent running `git config --global --add safe.directory`.

Git also reads `/etc/gitconfig`, below the user files in precedence, and the sandbox image has none. Mounting there keeps the image's file and its writability:

```suggestion
        run_args.extend(["-v", f"{host_gitconfig}:/etc/gitconfig:ro"])
```

Checked with git 2.53: name, email and aliases come from the mounted file, `safe.directory /workspace` still applies, and `git config --global --add` succeeds. The user-level alternative, `~/.config/git/config`, needs a `mkdir` in the Dockerfile first, since Docker would otherwise create `~/.config` root-owned.

scripts/agentic-sandbox.py
```diff
@@ -202,6 +202,9 @@ def log(msg: str) -> None:
 
     run_args.extend(cache_mounts)
     run_args.extend(_agent_mount_args(args.agent))
+    host_gitconfig = os.path.expanduser("~/.gitconfig")
```
philipcraig, line 205: Advisory. Git's other user-level location is `~/.config/git/config`, and hosts that keep their config there get no identity, silently. Falling back to that path when `~/.gitconfig` is absent covers it. Two properties of the mount worth a line in CONTRIBUTING: `git config --global` on the host rewrites the file via rename, so a running sandbox keeps the old contents until restarted; and with `commit.gpgsign = true` on the host, sandbox commits fail because no key is inside.

philipcraig (commented):
Thanks, the run-time mount removes the staleness, TOML and layer-cache problems. One issue with the mount target, with a one-line fix; the second comment is advisory.

jbcoe:
I fear that this will cause more problems than it solves.

PR #383, opened 2026-09-13
Name the fixed-name summary jobs in CONTRIBUTING

### LINKED ISSUE #382: CONTRIBUTING names the wrong required checks
`CONTRIBUTING.md` says the following checks are required for merging to `main`:

> `GCC trunk Release`, `GCC trunk Debug`, `GCC-16 Release`, `GCC-16 Debug`, `asan`, `tsan`, `uv-lock`, `pre-commit`

The `main` ruleset currently requires only `pre-commit` and `uv-lock` (`gh api repos/jbcoe/cc-protocol/rules/branches/main`). The six matrix-leg names cannot be required at all: when the change-detection job skips a leg, GitHub reports it under its unexpanded matrix name, which is why #271 added the fixed-name `bazel`, `cmake` and `sanitizers` summary jobs and why #249 tracks adding those to the ruleset.

Two things to fix:

- [ ] Rewrite the paragraph to name the summary jobs (`bazel`, `cmake`, `sanitizers`, `uv-lock`, `pre-commit`) and to say that the per-configuration legs and the macOS jobs are informational.
- [ ] Keep it in step with #249: either update the paragraph when the ruleset changes, or word it as "the checks the ruleset requires are the fixed-name summary jobs" without listing them, so the document cannot drift again.

philipcraig:
CONTRIBUTING listed matrix legs as the checks required for merging to `main`. Those legs cannot be required: a leg skipped by change detection reports under its unexpanded matrix name, which is why #271 added the fixed-name `bazel`, `cmake` and `sanitizers` summary jobs.

The paragraph now names the summary jobs, says the ruleset's required checks are drawn from them without listing the current set, and marks the per-configuration legs and the macOS jobs as informational.

Closes #382


jbcoe (approved):
Thanks for this.

PR #385, opened 2026-09-13
Give vtable entries a noexcept-carrying pointer type

### LINKED ISSUE #365: Vtable entries lost `noexcept` in the detail split
Follow-up from https://github.com/jbcoe/cc-protocol/pull/355#discussion_r3997095493.

On main, `fn_ptr_t` takes a `bool Noexcept` parameter and `generate_vtable_specs` passes `is_noexcept(member)`, so the vtable entry for `int f(int) const noexcept` has type `int(*)(const void*, int) noexcept`. After the split, `overload_spec.hh` defines `fn_ptr_t` without the parameter and `vtable.hh` no longer passes it, so the same entry is `int(*)(const void*, int)`.

Consequences:

- The comment in `generate_vtable_specs` still promises `noexcept(...)`.
- The `bool Noexcept` parameter on the two trampolines in `vtable.hh` is always deduced as `false`.
- Each `noexcept(IsNoexcept)` thunk in `member_function_thunks.hh` and `operator_thunks.hh` calls through a potentially-throwing pointer, so GCC wraps the indirect call in a terminate-on-throw region instead of a tail jump.

Either restore a `noexcept`-carrying `fn_ptr_t` for the vtable specs, or drop the dead `Noexcept` parameter from the trampolines and fix the comment. Carrying `noexcept` through the split hit reflection compile errors that were deferred from #355.

jbcoe:
`fn_ptr_t` now takes `IsNoexcept` as an explicit argument, and vtable entries are built with it, so a `noexcept` interface member gets a `noexcept` vtable entry. 

`member_function_thunk` and `operator_thunk` deduce their `IsNoexcept` from the pointer type instead of taking it as a separate template argument. Unspecialised instantiations of those two templates now fail with a readable `static_assert` instead of an incomplete-type error.

Fixes #365.


member_function_thunks.hh
```diff
@@ -98,10 +101,12 @@ template <std::meta::info Member, bool IsConst, typename EnclosingType,
           typename ProtocolType, typename Vtable>
 struct member_function_thunk_for<overload_spec<Member, IsConst>, EnclosingType,
                                  ProtocolType, Vtable> {
-  // Build the function-pointer type R(*)(Args...) from the method's return
-  // type and parameter types.
+  // Build the function-pointer type R(*)(Args...) noexcept(...) from the
+  // method's return type, parameter types and noexcept-ness.
   static consteval std::meta::info fn_ptr_type() {
```
philipcraig, line 105: This recipe is now copied three times (here, `operator_thunk_for::fn_ptr_type`, and `generate_vtable_specs`), and this PR exists because one copy drifted from the others. The thunk already holds `vtable_entry`, and the view trampolines already take their pointer type from `type_of(vtable_member)`, so the thunk signature could be `type_of(vtable_entry)` rather than rebuilt. If that is too large for this PR, a single shared consteval helper next to `fn_ptr_t` in `overload_spec.hh` would at least keep the three in step.

operator_thunks.hh
```diff
@@ -38,17 +38,19 @@ namespace xyz::detail {
 // letting `ProtocolType` be recovered with a plain static_cast.
 template <std::meta::operators Operator, typename FnPtrType,
           typename ProtocolType, typename Vtable, std::meta::info Member,
-          bool IsConst, bool IsNoexcept>
+          bool IsConst>
 struct operator_thunk {
-  operator_thunk() =
-      delete ("Unspecialized operator thunk cannot be instantiated");
+  static_assert(false,
```
philipcraig: The primary is also selected for any `Operator` without a specialisation, and for `op_arrow`/`op_star` with a non-nullary pointer, so this message misattributes those cases to the pointer type. Either say "no operator_thunk specialisation for this Operator/FnPtrType" or keep the pre-PR `= delete(...)` idiom that `operator_overload_set` still uses further down.

vtable_tests.cc
```diff
@@ -0,0 +1,48 @@
...
+  int f(int x) const;
+};
+
+using xyz::detail::vtable_t;
+
+// The sole non-typeid entry of a single-member vtable, found by position
+// rather than xyz::detail::find_vtable_entry to avoid instantiating the
+// mangled-name lookup a second time for the same member.
+template <typename T>
+consteval std::meta::info single_vtable_entry() {
```
philipcraig: `find_vtable_entry<^^vtable_t<NoexceptInterface>, ^^NoexceptInterface::f>()` compiles from here and gives the same entry. `mangled_name_of<Member>` is a variable template, so it is already instantiated by `generate_vtable_specs` and the lookup is not repeated; the comment's reason for avoiding it does not hold. Using it also stops the test depending on the typeid slot being first and the member entry being last, which would otherwise let the throwing-interface test pass vacuously on any non-function trailing entry. That also drops the helper and `#include <vector>`.

philipcraig (changes requested):
The head fails to compile under clang-p2996 (the clang-tidy job is red for this). The six friend declarations in `protocol.hh` (lines 247, 252, 563, 568, 659, 664) still list the trailing `bool IsNoexcept` template parameter that this PR removes from `member_function_thunk` and `operator_thunk`:

```
protocol.hh:247:3: error: too many template parameters in template redeclaration
```

then `vtable_` / `object_` become inaccessible from the thunks because friendship is not granted. GCC accepts the arity mismatch, so every other job stays green. Same trap as #347: drop `, bool IsNoexcept` from all six.

Title and second paragraph of the commit message: the body correctly names #350 as the cause, but the title blames the detail split, and the second paragraph describes adding a `noexcept_fn_ptr_t` used only for vtable entries. The diff does the opposite: `fn_ptr_t` itself regains a leading `bool`, and both thunk families now deduce `IsNoexcept` from the pointer type. Worth rewriting so the message matches the code.

`vtable.hh` lines 55–59: the doc comment on `generate_vtable_specs` still describes the entry types as `R(*)(void*, Args...)` with no `noexcept`; the inner comment at line 70 is now the accurate one.

vtable_tests.cc
jbcoe: I'll drop the test. It adds little value.

member_function_thunks.hh
jbcoe: I'll clean up duplication in another PR. It is not in scope here.

operator_thunks.hh
jbcoe: I've switched to "Unspecialized X cannot be instantiated."

That will avoid churn.

vtable_tests.cc
jbcoe: Fixed. Thanks for this.

philipcraig (approved):
All three body items are addressed: the six friend declarations match the new parameter lists and clang-tidy is green, the commit message now describes what the diff does, and the `generate_vtable_specs` comment carries `noexcept(...)`.

On the test helper: you are right and I was wrong. Using `find_vtable_entry` from the test TU does fail on GCC trunk with `symbol mangled_name_of<...> is already defined`. It only happens when the member's class has internal linkage; the same lookup in a named namespace compiles, and `protocol<I>` with an anonymous-namespace `I` compiles and runs, because there the second use is inside a template. That looks like a GCC bug with `define_static_string` emitting a local symbol twice, so the comment's explanation is accurate and the mangle-by-name helper is the right workaround.

One small thing before merging: the PR title still says "lost in the detail split", while the commit title is "Give vtable entries a noexcept-carrying pointer type". With three commits on the branch the squash merge will take the PR title, so it is worth renaming the PR to match.

jbcoe:
Too late to amend the PR title. Thanks for the review.

PR #387, opened 2026-09-13
Select an empty xcode_config for macOS Bazel builds

### LINKED ISSUE #373: Address follow-ups from the macOS CI jobs
The macOS CI jobs from #363 build and pass, but a review of the branch at 2a4b241 turned up a set of smaller problems that are better handled separately from that PR.

- [x] **Duplicate rpath on every macOS link.** With `CMAKE_SHARED_LIBRARY_SUFFIX` the runtime-library lookup now succeeds on macOS and appends the Homebrew Cellar directory to `CMAKE_BUILD_RPATH` (`CMakeLists.txt`), `-Wl,-rpath` (`scripts/bazel.py`) and the install_test configure (`scripts/cmake.py`). The Homebrew GCC driver already embeds that rpath, so every link prints `ld: warning: duplicate -rpath ... ignored`. Gate the mechanism on Linux, where it exists for `/opt/gcc-latest` and `ld.so`.
- [x] **`LD_LIBRARY_PATH` plumbing is dead on macOS.** `scripts/bazel.py` exports `--test_env=LD_LIBRARY_PATH` and sets it in the server environment; dyld never reads it. The `find_libstdcxx_directory` docstring and the `CMakeLists.txt` comment still describe `ld.so` semantics for a platform they now claim to cover. Fix follows from the previous item.
- [x] **Compiler discovery does not find `gcc@16`.** `CONTRIBUTING.md` tells macOS contributors to `brew install gcc@16` and run the scripts, but `find_reflection_compilers` only probes `/opt/gcc-latest` and `g++-16` on `PATH`. Today `gcc@16` is an alias of the linked `gcc` formula, so this works. Once Homebrew's `gcc` moves to 17, `gcc@16` becomes keg-only and off `PATH`, and the scripts fall back to AppleClang with a Linux-only error message. Probe `brew --prefix gcc@16` or tell contributors to export `CXX`. The workflow comments calling it "the gcc@16 formula" are also inaccurate today.
- [x] **`ENABLE_SANITIZERS` contradicts `cmake/sanitizers.cmake`.** The option's help text says "if available", but a requested and unsupported sanitizer is now a `FATAL_ERROR`, and the `COMPILER_SUPPORTS_X AND ENABLE_X` generator expressions in `cmake/xyz_add_test.cmake` carry a half that can no longer be false. Drop the option and key each block on `ENABLE_ASAN`/`ENABLE_UBSAN`/`ENABLE_TSAN` alone. The imported sanitizer targets are not `GLOBAL`, so they are only visible in `tutorials/` because of include order.
- [x] **Three mirrored runtime-library lookups.** The GNU and Clang branches in `CMakeLists.txt` are identical apart from the library stem, and `scripts/compiler_discovery.py` mirrors them in Python with a different exists guard and fallback. `scripts/cmake.py` already passes `-DCMAKE_BUILD_RPATH` for install_test, so it can do the same for the main configure and the CMake block can go, or the reverse. Either way the `.dylib`/`.so` suffix should be computed once.
- [x] **macOS install step duplicated.** "Install GCC 16 (macOS)" is byte-identical in `bazel.yml` and `cmake.yml`, evaluates `brew --prefix gcc@16` twice, and exports `CXX`/`CC` that the linked formula already puts on `PATH`. The `GCC16_MACOS` compiler type exists only to select this step and duplicates the OS dimension the matrix already has in `runner.os`. Key the apt and brew steps on `runner.os`, or make the step a composite action like `.github/actions/install-gcc-trunk`.
- [x] **`cmake/sanitizers.cmake` is three copy-pasted blocks.** Each sanitizer's flags are spelled three times (the list, the hand-joined probe string, the probe argument), so a flag added to the list is applied but never probed. A `foreach(sanitizer ASAN UBSAN TSAN)` with `list(JOIN ...)` collapses it to one block and keeps the link-time probe.
- [x] **macOS Debug CMake leg is redundant.** The Debug preset differs from Release only by build type and IPO. Its distinctive value, assert-enabled death tests, already runs on macOS in the Bazel job (fastbuild, no `-DNDEBUG`). Three macOS jobs per run hit GitHub's macOS concurrency cap, which is the slow-provisioning problem #363 cites. Move the macOS entry to `include:` with a fixed `configuration: Release`.
- [ ] **Xcode toolchain autoconfig on every macOS Bazel run.** The macOS Bazel job spends about 28 s longer than Linux in analysis with every compile a cache hit, and re-runs all link and test actions. This matches `osx_cc_configure` running per job because `@local_config_cc` lives in the uncached output base. `build:macos --repo_env=BAZEL_USE_CPP_ONLY_TOOLCHAIN=1` skips the Xcode toolchain and may also let the `-macos_reproducible`, `-macos_default_link_flags` and `-lstdc++` lines in `.bazelrc` go. Untested; the attribution is inferred from where the stall happens.

philipcraig:
Closes #373 (its last unchecked item).

The macOS Bazel job spends its extra time in analysis, not execution: on main the phase takes 46 s against 7 s on Linux, while the execution phase and disk-cache hit counts match. The item attributed that to Bazel's Apple toolchain autoconfiguration, but under Bazel 9.2 and rules_cc 0.2.22 the job already uses rules_cc's plain unix toolchain on macOS. The stall is `@local_config_xcode`: the toolchain config reads an xcode_config to build `-mmacosx-version-min`, and the default one compiles and runs xcode-locator and then `xcodebuild -version -sdk` for every Xcode on the runner, on every fresh output base.

Off macOS Bazel substitutes an empty `xcode_config` stub, so `build:macos` now selects our own copy of that stub through `--xcode_version_config` and disables the `macos_minimum_os` feature, which would otherwise derive `-mmacosx-version-min=10.11` from it. `apple_support` becomes a direct dependency only to load the rule; it is already fetched through `bazel_tools`. The `-macos_reproducible`, `-macos_default_link_flags` and `-lstdc++` lines stay: they belong to the unix toolchain, so `BAZEL_USE_CPP_ONLY_TOOLCHAIN` would not have removed them.

`GCC-16 (macOS)` job, warm disk cache (main: run 34777601259; this PR: run 34778438547 attempt 2):

| | main | this PR |
|---|---|---|
| analysis phase | 46 s | 10 s |
| Run Bazel step | 83 s | 15 s |
| Linux `GCC-16` Run Bazel step, same runs | 38 s | 13 s |

The first run after this change rebuilds everything once, since the compile flags changed.


jbcoe (approved):
Thanks for this!

PR #388, opened 2026-09-14
Extend operator support

jbcoe:
Extend operator forwarding support to more overloadable operators. `operator=`, `operator co_await`, the comparison operators, explicit-object members and unary `operator&` are rejected on an interface.

Fixes a self-assignment bug: `protocol`'s copy/move assignment used `this == &other`, which a forwarded `operator&` would hijack; now uses `std::addressof`.

We opted not to use macros to reduce code verbosity. The limits of C++26 reflection mean that some boilerplate structure for operator thunks is somewhat repetitive.


name_mangling_tests.cc
```diff
@@ -171,4 +174,205 @@ TEST(NameManglingTest, ExtraOperators) {
...
+  int operator-=(int);
+  int operator*=(int);
+  int operator/=(int);
+  int operator%=(int);
+  int operator^=(int);
+  int operator&=(int);
+  int operator|=(int);
+
+  // Comparison.
+  bool operator==(int) const;
```
jbcoe, line 220: Comparisons should probably be excluded from the set of interface functions.

name_mangling_tests.cc
```diff
@@ -171,4 +174,205 @@ TEST(NameManglingTest, ExtraOperators) {
...
+  int operator,(int);
+  void* operator->();
+  int operator->*(int);
+  int operator()(int);
+  int operator[](int);
+
+  int operator co_await();
+
+  // Implicitly static members: no cv/ref-qualification is possible.
+  void* operator new(unsigned long);
```
jbcoe, line 259: Static members are not eligible as interface functions, they should skipped?

name_mangling_tests.cc
jbcoe: I will address this in a follow up PR as it involves adding logic out of the current scope and this PR is already large.

name_mangling_tests.cc
jbcoe: I will address this in a follow up PR as it involves adding logic out of the current scope and this PR is already large.

conformance.hh
```diff
@@ -165,21 +157,20 @@ template <std::meta::info Type>
...
+// infos` never reaches them: `is_static_member` skips them first.
+consteval bool is_unsupported_operator(std::meta::operators op) {
+  using enum std::meta::operators;
+  return op == op_equals || op == op_co_await;
 }
 
 // The named, non-static, non-special member functions and call operators of
 // `Type`, static or not, in declaration order. Ref-qualified functions and
-// binary `operator*` are unsupported on protocol interfaces.
+// the operators `is_unsupported_operator` names are unsupported on protocol
```
philipcraig: A member `operator==` or `operator<=>` on the interface now becomes a required forwarded member. That contradicts DRAFT.md:563-570 and breaks interfaces that conform on `main`:

```cpp
struct I { int f() const; bool operator==(const I&) const = default; };
protocol<I> p(C{});  // is_protocol_conformant_v<I, C> is now false
```

The failure only says `is_protocol_conformant_v evaluated to false`, and because a defaulted `==` is implicitly `noexcept`, even a `C` that declares `bool operator==(const I&) const` fails unless it is also `noexcept`. For the follow-up that excludes comparisons: a `!is_defaulted(member)` filter covers the `= default` idiom, and a throw naming the operator (as `=`/`co_await` get) would make the remaining cases diagnosable.

conformance.hh
```diff
@@ -165,21 +157,20 @@ template <std::meta::info Type>
...
-  if (!params.empty() && is_explicit_object_parameter(params.front())) {
-    return params.size() - 1;
-  }
-  return params.size();
+// `operator=` would collide with `protocol`/`protocol_view`'s own copy/move
+// assignment, and `operator co_await` has no defined forwarding semantics;
+// neither has an `operator_thunk` specialisation. `operator new`/`delete`
+// and their array forms are always static, so `protocol_interface_function_
+// infos` never reaches them: `is_static_member` skips them first.
+consteval bool is_unsupported_operator(std::meta::operators op) {
```
philipcraig, line 171: This is a third hand-kept list alongside the 44 `std::meta::operators` enumerators and the 38 `operator_thunk` specialisations. If they drift, an operator without a specialisation skips this diagnostic and surfaces as `static assertion failed: Unspecialized operator_overload_set cannot be instantiated` from inside `protocol<I>`'s base list. Deriving "unsupported" from `is_complete_type(substitute(^^operator_overload_set, ...))` in `generate_member_bases_wrapper` works on trunk if the primary templates in operator_thunks.hh become declaration-only instead of `static_assert(false)` bodies. Also `symbol_of(op_equals)` is `"="`, so the message reads `operator = is not supported`; `co_await` reads fine.

name_mangling.hh
```diff
@@ -236,23 +236,90 @@ consteval std::string mangle_type(std::meta::info type) {
...
-    }
-    if (operator_of(function) == std::meta::operators::op_square_brackets) {
-      return "ix";
-    }
-    if (operator_of(function) == std::meta::operators::op_arrow) {
-      return "pt";
+    std::meta::operators op = operator_of(function);
+    // A member `+`, `-`, `*` or `&` is unary when it has no parameters: the
+    // only argument is the implicit object parameter.
+    bool unary = parameters_of(function).empty();
```
philipcraig, line 253: `parameters_of` includes an explicit object parameter, so `int operator-(this const I&)` on an interface mangles as binary `mi`, and vtable.hh:79 and `fn_ptr_type` in operator_thunks.hh:80 make the same assumption. Such members fail conformance for every candidate today, so nothing reaches the wrong shape yet, but this PR deletes `declared_argument_count`, the one place that recorded the invariant. Either reject explicit-object interface members next to the ref-qualifier rejection, or keep the invariant in the comment here.

name_mangling.hh
```diff
@@ -236,23 +236,90 @@ consteval std::string mangle_type(std::meta::info type) {
...
 
-// Returns the mangled function-name atom for `function`: the Itanium
-// <operator-name> `cl` for the call operator, or a length-prefixed
-// <source-name> for its identifier otherwise. `identifier_of` throws for
-// `operator()`, which has no identifier.
+// Itanium ABI <operator-name> codes that don't depend on arity. `+`, `-`,
+// `*` and `&` are excluded: each mangles differently as a unary operator
+// (no parameters, besides the implicit object parameter) than as a binary
+// one, and `base_name_of` handles those four separately.
+consteval std::optional<std::string_view> fixed_operator_code(
```
philipcraig: `base_name_of` already switches on the same enum for the four arity-dependent operators. Folding all 44 enumerators into that switch with no `default` drops the array, the linear scan, the `optional` and the magic 40, and a missing enumerator becomes a `-Wswitch` error at the definition under the `-Wall -Werror` both builds use, instead of a consteval throw the first time an interface declares that operator. `builtin_type_code` needs its table because it keys on `std::meta::info`; this one doesn't.

operator_thunks.hh
```diff
@@ -32,6 +32,19 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 namespace xyz::detail {
 
+template <std::meta::info Member, typename ProtocolType, typename Vtable,
```
philipcraig: member_function_thunks.hh:54-82 still writes this body out twice by hand, so the valueless assert now lives in three places. The helper touches nothing private (`vtable_` and `object_` arrive as arguments, `valueless_after_move()` is public), so it can sit in vtable.hh next to `find_vtable_entry` and serve both thunk kinds. `ProtocolType` and `Vtable` are also derivable from the deduced pointer types; every call site passes the same `protocol_object, protocol_object->vtable_, protocol_object->object_` triple.

operator_thunks.hh
```diff
@@ -43,45 +56,76 @@ struct operator_thunk {
...
+                        std::meta::reflect_constant(IsConst)
+                    }):];
+  // clang-format on
+};
+
+template <typename OverloadSpec, typename ProtocolType, typename Vtable>
+using operator_thunk_t =
+    operator_thunk_for<OverloadSpec, ProtocolType, Vtable>::type;
+
 // operator()
```
philipcraig, line 89: The 38 `operator_thunk` and 38 `operator_overload_set` specialisations differ only in the enumerator and the operator token; the six protected defaulted special members alone account for about 450 of the new lines. Reflection can't synthesise these (`define_aggregate` emits data members only) and a single constrained-member class fails on `operator<=>(Args...)` with empty `Args`, so one declaration pair per token is forced, but a function-like macro in the style of `XYZ_CONSTEVAL_CHECK` would take this file from ~1800 lines to ~100 and make fixes to the shared body apply once. (`cppcoreguidelines-macro-usage` only fires on object-like macros.) Separately, `operator->` at 200-231 is keyed on `R (*)(Args...)` but declares `operator->()` with no `Args`, unlike `operator~`/`operator!` which use the fixed `R (*)()` shape.

protocol_operator_tests.cc
```diff
@@ -0,0 +1,1182 @@
...
+  protocol<Interface> p(Conforming{});
+  // NOLINTBEGIN(performance-unnecessary-copy-initialization): the test
+  // exercises copy construction on purpose.
+  protocol<Interface> copy(p);
+  // NOLINTEND(performance-unnecessary-copy-initialization)
+  EXPECT_EQ(copy(21), 42);
+}
+
+// Tests that dispatching fails gracefully for a moved-from protocol, for
+// each operator.
```
philipcraig: The comment promises death tests for each operator but only `()`, `[]`, `*` and `->` have them. Also no test dispatches unary `+`, `-` or `&` through a protocol, and none declares both the unary and binary form of `+`, `-`, `*` or `&` in one interface, so the `ps/ng/de/ad` vs `pl/mi/ml/an` split is only checked at the mangling level, never through `operator_overload_set` and the vtable. A unary `&` test with an assignment step would have caught the `std::addressof` problem. The compound-assignment and `++`/`--` groups also lack the const-view negative check every other group has.

protocol_operator_tests.cc
```diff
@@ -0,0 +1,1182 @@
...
+#include <gtest/gtest.h>
+
+#include "protocol.hh"
+
+using xyz::reflection::protocol;
+using xyz::reflection::protocol_view;
+
+namespace {
+
+// Concepts for call operator tests. Duplicated from protocol_test.cc, which
```
philipcraig: test_helpers.h is already a source of this `cc_test` target, so these three can move there instead of being duplicated; the slicing probe mirrors `operator_overload_set`'s template parameter list and already changed in this PR when `Operator` was added.

name_mangling_tests.cc
```diff
@@ -171,4 +174,215 @@ TEST(NameManglingTest, ExtraOperators) {
   static_assert(mangle(^^A::operator->) == "fn_ptPv");
 }
 
+// A member function's mangled Itanium <operator-name> for every operator
```
philipcraig, line 177: "See the PR description" won't survive the squash merge; either inline the reproduction or point at a file in the repo.

philipcraig (commented):
Forwarding unary `operator&` breaks `protocol`'s own assignment. The self-assignment checks at protocol.hh:413 and :439 spell `&other`, which now resolves to the interface's `operator&` thunk:

```cpp
struct I { int operator&() const; int f() const; };
struct T { int operator&() const { return 99; } int f() const { return 3; } };
protocol<I> p{T{}}, q{T{}};
q = p;             // protocol.hh:413: comparison between pointer and integer
q = std::move(p);  // protocol.hh:439: same
auto r = &p;       // int, 99
```

If the interface's `operator&` returned a pointer the assignment would compile and compare `this` against the thunk's result, so genuine self-assignment falls through to destroy-then-copy-from-destroyed. `std::addressof(other)` at both sites fixes the wrappers (protocol.hh already uses it at 527 and 622). Whether `&p` on a wrapper should ever dispatch to the interface is a separate decision; at minimum it wants a sentence in the docs. Fine as a follow-up if you prefer.

CONTRIBUTING.md:194-196 still says no operators other than `operator()` are supported, and DRAFT.md:565-570 still says equality and comparison operators are not generated. Both need updating, or DRAFT.md needs a note that the reference implementation has diverged.

Inline comments cover the comparison-operator regression, a latent explicit-object mangling bug, and some duplication. Any of these can go to follow-up PRs; none needs to hold this one.

conformance.hh
jbcoe: I've removed `operator==` and `operator<=>` support for now. We can think about this more in a (more complicated follow-up PR).

conformance.hh
jbcoe: I'm happy with hand kept logic and readable code.

operator_thunks.hh
jbcoe: This is fixed.

operator_thunks.hh
jbcoe: I'd like to avoid using macros. I am happy with code duplication; it cleanly illustrates the limits of C++26 reflection.

jbcoe: Treating `operator->()` differently would require special handling elsewhere. Perhaps we could assert that Args... is empty? I'll leave it as is for now.

protocol_operator_tests.cc
jbcoe: Fixed.

name_mangling_tests.cc
jbcoe: Fixed. Apologies that this leaked in.

protocol_operator_tests.cc
jbcoe: A single death test suffices now.

name_mangling.hh
jbcoe: Fixed. Thanks for this.

name_mangling.hh
jbcoe: Fixed — explicit-object interface members are now rejected.

jbcoe:
Thanks for the detailed review. All should be good to go now.

vtable.hh
```diff
@@ -52,6 +55,21 @@ consteval std::meta::info find_vtable_entry() {
                            std::string(name) + "'");
 }
 
+template <std::meta::info Member, typename Vtable, typename ProtocolObject,
+          typename VtablePtr, typename Object, typename... Args>
+decltype(auto) call_through_vtable(ProtocolObject* protocol_object,
```
philipcraig, line 60: #400: `protocol_object` is only used inside the `assert`, so with `NDEBUG` this is `-Werror=unused-parameter` in an installed header. `[[maybe_unused]]` here; the issue also notes the `-O0` cost of the helper not being inline.

conformance.hh
```diff
@@ -178,21 +170,28 @@ template <std::meta::info Type>
...
+// space after "operator", punctuation tokens only without one.
+consteval std::string operator_diagnostic_name(std::meta::operators op) {
+  std::string_view symbol = symbol_of(op);
+  bool is_word = symbol.front() >= 'a' && symbol.front() <= 'z';
+  return is_word ? "operator " + std::string(symbol)
+                 : "operator" + std::string(symbol);
+}
+
+// Some operators are excluded from protocol interfaces.
+consteval bool is_unsupported_operator(std::meta::operators op) {
```
philipcraig, line 184: #401: on `main` a member `operator==` on the interface was silently ignored; this makes it a hard error, so

```cpp
struct I { int f() const; bool operator==(const I&) const = default; };
```

no longer compiles against any candidate. There is also no rejection test for any of the seven comparisons; `UnsupportedOperatorsAreRejected` covers `=` and `co_await` only.

conformance.hh
```diff
@@ -155,11 +150,8 @@ consteval auto conformance_candidate_infos() {
       members_of(Type, std::meta::access_context::unprivileged()) |
       std::views::filter(std::meta::is_function) |
       std::views::filter([](std::meta::info member) consteval {
-        return has_identifier(member) ||
-               is_operator<std::meta::operators::op_star>(member) ||
-               is_operator<std::meta::operators::op_arrow>(member) ||
-               is_operator<std::meta::operators::op_parentheses>(member) ||
-               is_operator<std::meta::operators::op_square_brackets>(member);
+        return !is_special_member_function(member) &&
+               (has_identifier(member) || is_operator_function(member));
```
philipcraig, line 154: #402: `operator T()` has no identifier and `is_operator_function` is false for it, so conversion functions fall out here before conformance or rejection sees them. A candidate without the conversion conforms, and `static_cast<bool>(p)` is ill-formed even when both sides declare it. Pre-existing, but this is now the only operator category that neither forwards nor throws.

conformance.hh
```diff
@@ -206,10 +205,24 @@ consteval std::vector<std::meta::info> protocol_interface_function_infos() {
       throw std::runtime_error("ref-qualified member function '" + name +
                                "' is not supported in a protocol interface");
     }
-    if (is_operator<std::meta::operators::op_star>(member) &&
-        declared_argument_count(member) != 0) {
+    std::vector<std::meta::info> params = parameters_of(member);
+    if (!params.empty() && is_explicit_object_parameter(params.front())) {
+      std::string name = has_identifier(member)
```
philipcraig, line 210: #406: this name expression is a copy of the one eight lines up, and `operator_diagnostic_name` below renders names a third way (no overload, no parameter list). Hoisting `name` once and nesting the two operator checks under a single `if (is_operator_function(member))` removes the duplication and gives every rejection the same wording.

CONTRIBUTING.md
```diff
@@ -195,15 +195,6 @@ Their public, non-virtual, non-template member functions define the protocol;
...
-- Limitations: No operators other than `operator()` are supported. Member
-  function templates are not matched, since only non-template functions are
-  considered as candidates. Conformance checking accounts for an interface
-  member's lvalue/rvalue reference qualifier, but the generated call wrapper
-  does not itself apply the qualifier. There is no conversion between a
-  `protocol`/`protocol_view` of one interface and another. Conformance
-  checking is O(N\*M) in the number of interface and candidate member
-  functions.
-
 - Guidance: Developers should refer to `protocol_test.cc`,
```
philipcraig, line 198: #407: The deleted Limitations bullet carried four statements that are still true and are now documented nowhere: member function templates are not matched (#393), the generated wrapper does not apply an interface member's ref-qualifier (#356), there is no conversion between wrappers of different interfaces, and conformance checking is O(N*M). They should come back, though they never really belonged in a contributing guide; the issue proposes a `doc/limitations.md` that this bullet links to, which can also list what #388 rejects. The Supported Features bullet above still names only `operator()`.

protocol_operator_tests.cc
```diff
@@ -0,0 +1,1064 @@
...
+  // test exercises the moved-from state on purpose.
+  EXPECT_TRUE(valueless_after_move(p));
+
+  EXPECT_DEATH(p(21), "cannot call member function of valueless protocol");
+  // NOLINTEND(bugprone-use-after-move,hicpp-invalid-access-moved)
+}
+
+#endif
+
+TEST(ReflectionProtocolTest, OperatorSquareBrackets) {
```
philipcraig, line 226: #404: from here down each test declares its own `Interface`/`Conforming` pair and instantiates all three wrapper kinds, about 132 `define_aggregate` generations for the TU. It is roughly 3 s of instantiation in every job that builds the tests; one shared interface per operator family halves it in a matched experiment. Test-structure optimisation only, no design question.

operator_thunks.hh
```diff
@@ -43,45 +41,76 @@ struct operator_thunk {
...
+template <typename OverloadSpec, typename ProtocolType, typename Vtable>
+struct operator_thunk_for;
+
+template <std::meta::info Member, bool IsConst, typename ProtocolType,
+          typename Vtable>
+struct operator_thunk_for<overload_spec<Member, IsConst>, ProtocolType,
+                          Vtable> {
+  // Build the function-pointer type R(*)(Args...) noexcept(...) from the
+  // method's return type, parameter types and noexcept-ness.
+  static consteval std::meta::info fn_ptr_type() {
```
philipcraig, line 63: #405: third copy of this builder, after `member_function_thunk_for::fn_ptr_type()` and the inline one in `generate_vtable_specs`. vtable.hh's copy already differs (`dealias` on parameter types, relevant to #397); all three must agree for the thunk specialisation to match the vtable entry. One `function_pointer_type_of(info)` in vtable.hh would serve all three.

philipcraig (approved):
Two new things since last round, both filed as issues so they can be handled here or after:

- #400: `call_through_vtable` in vtable.hh only uses `protocol_object` inside the `assert`, so the installed header fails under `-DNDEBUG -Wall -Wextra -Werror` with `unused parameter 'protocol_object'`. `main` is clean under the same flags. CI has no `-Wextra`, so it stays green. `[[maybe_unused]]` fixes it.
- #401: rejecting `==`/`<=>` turned a member comparison on the interface from silently ignored (what `main` does and DRAFT.md still says) into a hard consteval throw, so interfaces that compile on `main` stop compiling, including the `= default` idiom. Options are in the issue.

Docs still saying the old thing: protocol.hh:32 ("Neither implementation currently supports operators other than operator()"), DRAFT.md:152-154 (`protocol` does not provide `operator*` or `operator->`) and DRAFT.md:565-567 (member comparisons "are not generated"). The protocol.hh line is a one-word fix; the DRAFT.md ones can wait for #396 if that is imminent.

#403: `operator new`/`delete` on an interface are dropped by the `is_static_member` skip with no diagnostic, while the new mangling tests assert codes for them. Probably intended, so a docs decision rather than a code one.

The rest are inline and each points at an issue (#402, #404, #405, #406, #407). Nothing here blocks merging.

PR #389, opened 2026-09-14
Set user name and email for git and jj in agentic sandbox.

jbcoe:


jbcoe:
This is simpler than messing about with mounts and config files and does what we need.

philipcraig (approved):
nice

PR #391, opened 2026-09-15
Fix protocol_view ambiguity for overloads differing only by return type

### LINKED ISSUE #375: `protocol_view` has ambiguous overloads when return types differ
Under all_const the idiomatic const/non-const operator pair with differing return types is kept as two overloads and both are made const, so every use of the three new operators on protocol_view<I> is ambiguous:

```
struct I {
  const X& operator[](int) const;
  X& operator[](int);
};
protocol_view<I> mv(c);
mv[0];   // error: ambiguous overload for 'operator[]' (operand types are 'protocol_view<I>' and 'int')
*mv;     // same, for operator* / operator-> pairs
```

protocol<I> and protocol_view<const I> are fine. The rule predates this stack for named members, but operator[], unary operator* and operator-> are almost always written in exactly this shape, so protocol_view<I> cannot be used with them. The new operator tests only use pairs that both return int. Either drop the const overload when a non-const one matches ignoring const and return type, or document the limitation and add a differing-return-type test either way.

jbcoe:
closes #375


conformance.hh
```diff
@@ -47,10 +47,22 @@ consteval bool same_name(std::meta::info a, std::meta::info b) {
...
 // Returns `true` if the member functions `candidate` and `interface` have
-// the same name, de-aliased return type and de-aliased parameter types.
+// the same name and de-aliased parameter types.
 consteval bool same_name_and_parameters(std::meta::info candidate,
                                         std::meta::info interface) {
   if (!same_name(candidate, interface)) return false;
+  auto dealiased_type_of = [](std::meta::info parameter) {
+    return dealias(type_of(parameter));
+  };
+  return std::ranges::equal(parameters_of(interface), parameters_of(candidate),
```
RyanJK5, line 57: Just using type_of here would be fine. I believe reflections of parameter types cannot be aliases.

RyanJK5 (approved):

conformance.hh
jbcoe: I'll purge this in follow up. We get this wrong in a few places.

https://github.com/jbcoe/cc-protocol/issues/397

Issue #358, opened 2026-09-12
Bake the uv and Bazel caches into the sandbox image instead of Docker volumes


philipcraig:
On #315 jbcoe asked, after #330 baked the googletest checkout into the sandbox image:

> Should we do the same for UV and bazel cache? We can then dispense with Docker volumes altogether.

Today the sandbox mounts two named volumes (`scripts/agentic-sandbox.py`, `CACHE_VOLUMES`): `cc-protocol-uv-cache` at `~/.cache/uv` (#301) and `cc-protocol-bazel-repository-cache` at `~/.cache/bazel-repo`, fed from the image's `~/.bazelrc` (#304). Both exist so a warm sandbox builds without the network.

Baking them in would mean running `uv sync` and `bazel fetch //...` in the sandbox stage of `docker/Dockerfile`, with the same `~/.cache/uv` and `~/.cache/bazel-repo` paths, so the image already holds every wheel and every external repository the build needs. The Bazel fetch must go through the repository cache and not rely on the output base, which is keyed by the workspace path. What that changes:

- No volumes at all: `--cache-volumes`/`--no-cache-volumes`, `UV_LINK_MODE=copy` and the pre-created mount points go away, and there is no shared writable state, which is what made #315 hard.
- The image is the cache. A `uv.lock` or `MODULE.bazel` change needs `--rebuild-docker`, the same rule #330 set for the googletest tag; until then the sandbox falls back to the network for the missing artefacts, exactly as it does now on a cold volume.
- Image size grows by the wheel set and the Bazel external repositories (everything `MODULE.bazel` pulls in). Worth measuring before deciding.
- `docker run --network none` on the built image then proves the offline claim end to end, which is the check #325 asks CI to make.

Things to decide: whether the Bazel fetch belongs in the image or should stay a volume because external repositories change more often than `uv.lock`; and whether the devcontainer stage should share any of this or keep its local caches as it does today.

jbcoe:
I think that baking everything into the image is **much** simpler. Let's go with that.

PR #392, opened 2026-09-15
Save cache in the sandbox image rather than use docker volumes

jbcoe:
Store uv's package cache/venv and bazel's repository cache into the Docker sandbox.

Syncing past this change requires the agentic-sandbox Docker image to be rebuilt.

Docker volumes are no longer used and can be deleted:
* cc-protocol-uv-cache
* cc-protocol-bazel-repository-cache


docker/Dockerfile
```diff
@@ -122,35 +123,37 @@ ENV UV_CACHE_DIR="/home/vscode/.cache/uv"
...
-# repository_cache is defined in the sandbox user-directory config, not workspace config so only
-# the sandbox uses the respository cache.
 RUN echo "common --repository_cache=$XYZ_BAZEL_REPOSITORY_CACHE_DIR" >> /home/vscode/.bazelrc
 
-# Bake the googletest checkout into the image so CMake builds need no network.
+# Pre-warm bazelisk's download and install of the pinned Bazel binary to avoid
+# fetching and installing it on every run.
+# https://github.com/bazelbuild/bazelisk#how-does-bazelisk-know-which-bazel-version-to-run
+RUN --mount=type=bind,source=.bazelversion,target=.bazelversion \
+    bazel version
```
philipcraig, line 138: bazelisk only reads `.bazelversion` from a workspace root, and a directory holding just that file is not one (`bazel version` here reports "not invoked from within a workspace" and runs latest). The dropped `USE_BAZEL_VERSION="$(cat .bazelversion)"` prefix was what made the old step honour the pin. It works today only because latest is 9.2.0; once 9.3 ships this layer bakes 9.3 and the pinned 9.2.0 is downloaded again inside the `bazel fetch` layer.

```suggestion
RUN --mount=type=bind,source=.bazelversion,target=.bazelversion \
    USE_BAZEL_VERSION="$(cat .bazelversion)" bazel version
```

docker/Dockerfile
```diff
@@ -122,35 +123,37 @@ ENV UV_CACHE_DIR="/home/vscode/.cache/uv"
...
-# fetching and installing it on every run.
-# https://github.com/bazelbuild/bazelisk#how-does-bazelisk-know-which-bazel-version-to-run
-# We use .bazelversion to avoid hard-coding the bazel version number in multiple places.
-COPY .bazelversion .bazelversion
-RUN USE_BAZEL_VERSION="$(cat .bazelversion)" bazel version \
-    && rm .bazelversion
-
-# The sandbox image needs additional setup to use the Docker volumes for persistent cache.
+# The sandbox image needs additional setup to pre-warm caches used at runtime.
 FROM base AS sandbox
```
philipcraig, line 127: #302 put the `bazel version` pre-warm in `base` so the devcontainer (`target: base` in `.devcontainer/devcontainer.json`) got it too. Moving it into `sandbox` means every fresh devcontainer and Codespace downloads Bazel and extracts the install base on the first `./scripts/bazel.sh`. #358 left "whether the devcontainer stage should share any of this" open; I would keep the (corrected) `bazel version` RUN in `base` and move only `bazel fetch` and `uv sync` here.

docker/Dockerfile
```diff
@@ -122,35 +123,37 @@ ENV UV_CACHE_DIR="/home/vscode/.cache/uv"
...
+# https://github.com/bazelbuild/bazelisk#how-does-bazelisk-know-which-bazel-version-to-run
+RUN --mount=type=bind,source=.bazelversion,target=.bazelversion \
+    bazel version
+
+# Pre-warm Bazel's repository cache from the real target graph so builds
+# inside the sandbox need no network for external dependencies. The repo is
+# bind-mounted rather than COPY'd so the source never becomes part of an
+# image layer; only the populated repository cache needs to persist.
+RUN --mount=type=bind,source=.,target=. \
+    bazel fetch //...
```
philipcraig, line 145: There is no `.dockerignore`, so this walks every untracked directory in the checkout. `.bazelignore` lists exact paths (`build`, `_deps`, `.claude`), so `build-asan/`, `out/`, `_build/` and similar trees that `.gitignore` permits are still visited, and the FetchContent googletest checkout under them carries `googletest/test/BUILD.bazel` loading `@rules_python` and deps on `@abseil-cpp`. A `build-x/pkg/BUILD.bazel` loading `@rules_python` fails `bazel fetch //...` with "No repository visible as @rules_python from main repository", so `--rebuild-docker` dies in a package the contributor never wrote. Either add a `.dockerignore` (`.git`, `build*`, `_build`, `out`, `install`, `.venv`, `bazel-*`, `.claude`, `.ruff_cache`, `__pycache__`) or mount only the Bazel graph files (next comment).

docker/Dockerfile
```diff
@@ -122,35 +123,37 @@ ENV UV_CACHE_DIR="/home/vscode/.cache/uv"
...
+# fetching and installing it on every run.
+# https://github.com/bazelbuild/bazelisk#how-does-bazelisk-know-which-bazel-version-to-run
+RUN --mount=type=bind,source=.bazelversion,target=.bazelversion \
+    bazel version
+
+# Pre-warm Bazel's repository cache from the real target graph so builds
+# inside the sandbox need no network for external dependencies. The repo is
+# bind-mounted rather than COPY'd so the source never becomes part of an
+# image layer; only the populated repository cache needs to persist.
+RUN --mount=type=bind,source=.,target=. \
```
philipcraig, line 144: `RUN --mount=type=bind` keys the layer cache on a checksum of the mounted files, so `source=.` makes any change anywhere in the checkout (a one-line `.cc` edit, the `.git` index, `build/` objects) invalidate this layer and re-run the fetch from the network. Before this PR BuildKit sent only `.bazelversion`; now the whole context (`.git`, `build`, `.venv`, `.claude` and sources, roughly 255 MB here) goes to the daemon. The volumes survived rebuilds; discarded layers do not.

Neither `BUILD.bazel` uses `glob()`, so the graph inputs are `MODULE.bazel`, `MODULE.bazel.lock`, `.bazelrc`, `.bazelversion`, `.bazelignore`, `BUILD.bazel` and `tutorials/BUILD.bazel`. Mounting just those keeps the layer stable across source edits and also removes the stray-tree failure above.

docker/Dockerfile
philipcraig, line 145: This runs in the default `--lockfile_mode=update` on a read-only mount. With a stale `MODULE.bazel.lock` Bazel resolves online, the lockfile write fails with "Permission denied" in java.log only, and the command exits 0, so the image looks warm while registry files still need the network at runtime (they are served from `--repository_cache` only via hashes recorded in the lockfile). The first runtime `bazel` call then rewrites the lockfile in the contributor's `/workspace`. `bazel fetch --lockfile_mode=error //...` turns that into an image-build failure instead. Nothing fails today because the lockfile is in sync.

docker/Dockerfile
```diff
@@ -122,35 +123,37 @@ ENV UV_CACHE_DIR="/home/vscode/.cache/uv"
...
+    bazel fetch //...
+
+# Pre-warm uv's package cache and build the venv from the pinned lockfile so
+# `uv run`/`uv sync` inside the sandbox need no network.
+RUN --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
+    --mount=type=bind,source=uv.lock,target=uv.lock \
+    uv sync --locked
```
philipcraig, line 151: Once a layer misses, everything after it re-runs. The `bazel fetch` layer is the volatile one and it precedes the stable `uv sync --locked` (inputs `pyproject.toml`, `uv.lock`) and googletest clone (no inputs), so every fetch miss also re-downloads all wheels and re-clones googletest. The three RUNs write to disjoint trees, so put `uv sync` and the clone first and make the fetch the last RUN of the stage.

docker/Dockerfile
```diff
@@ -122,35 +123,37 @@ ENV UV_CACHE_DIR="/home/vscode/.cache/uv"
...
-# The sandbox image needs additional setup to use the Docker volumes for persistent cache.
+# The sandbox image needs additional setup to pre-warm caches used at runtime.
 FROM base AS sandbox
 
-# uv must copy files as the cache volume is on a different filesystem.
-ENV UV_LINK_MODE="copy"
-
+# repository_cache is defined in the sandbox user-directory config, not
+# workspace config so only the sandbox uses the respository cache.
 ENV XYZ_BAZEL_REPOSITORY_CACHE_DIR="/home/vscode/.cache/bazel-repo"
```
philipcraig, line 131: Bazel already has an always-on repository cache at `<output_user_root>/cache/repos/v1` (`bazel info repository_cache`). It is shared between the build-time `/workspace` fetch and the runtime `/workspace` checkout, is captured by the same image layer, and survives `bazel clean --expunge`. The separate path only mattered when the cache had to be a mountable volume, and this PR deletes the other two consumers of `XYZ_BAZEL_REPOSITORY_CACHE_DIR`. The comment's premise is also off: the devcontainer does use a repository cache, just the default one. I would drop these four lines and rely on the default; failing that, a literal rather than a runtime ENV nothing can act on.

docker/Dockerfile
```diff
@@ -122,35 +123,37 @@ ENV UV_CACHE_DIR="/home/vscode/.cache/uv"
...
+# Pre-warm bazelisk's download and install of the pinned Bazel binary to avoid
+# fetching and installing it on every run.
+# https://github.com/bazelbuild/bazelisk#how-does-bazelisk-know-which-bazel-version-to-run
+RUN --mount=type=bind,source=.bazelversion,target=.bazelversion \
+    bazel version
+
+# Pre-warm Bazel's repository cache from the real target graph so builds
+# inside the sandbox need no network for external dependencies. The repo is
+# bind-mounted rather than COPY'd so the source never becomes part of an
+# image layer; only the populated repository cache needs to persist.
```
philipcraig, line 143: Not only the repository cache persists: this layer also captures Bazel's output base for md5(`/workspace`) under `~/.cache/bazel/_bazel_vscode/`, and the runtime container reuses it because the checkout is mounted at the same path. Worth saying, since it is the part that grows on every rebuild.

docker/Dockerfile
```diff
@@ -122,35 +123,37 @@ ENV UV_CACHE_DIR="/home/vscode/.cache/uv"
...
+    --mount=type=bind,source=uv.lock,target=uv.lock \
+    uv sync --locked
+
+# Clone googletest into the image so CMake builds need no network.
 # With FETCHCONTENT_SOURCE_DIR_GOOGLETEST set, FetchContent performs no download
 # or update step and each build tree still compiles its own copy. The tag must
-# match GIT_TAG in CMakeLists.txt; bumping it needs --rebuild-docker.
+# match GIT_TAG in CMakeLists.txt.
 ENV FETCHCONTENT_SOURCE_DIR_GOOGLETEST="/home/vscode/.cache/googletest-src"
 RUN git clone --depth 1 --branch v1.18.0 https://github.com/google/googletest.git \
```
philipcraig, line 158: The tag is now pinned in three places (`CMakeLists.txt` `GIT_TAG`, `MODULE.bazel`, here), and the diff removes the one reminder that bumping it needs `--rebuild-docker`. With `FETCHCONTENT_SOURCE_DIR_GOOGLETEST` set CMake never checks `GIT_TAG`, so a bump in `CMakeLists.txt` leaves sandbox CMake builds silently on the old googletest until someone rebuilds. The bind-mount pattern this PR introduces can read the tag instead:

```dockerfile
RUN --mount=type=bind,source=CMakeLists.txt,target=CMakeLists.txt \
    git clone --depth 1 --branch "$(sed -n 's/^ *GIT_TAG *//p' CMakeLists.txt)" \
        https://github.com/google/googletest.git "$FETCHCONTENT_SOURCE_DIR_GOOGLETEST"
```

Or restore the reminder.

docker/Dockerfile
```diff
@@ -1,3 +1,4 @@
+# syntax=docker/dockerfile:1
```
philipcraig, line 1: `RUN --mount` is supported by the built-in BuildKit frontend, so this directive is not needed, and it adds a Docker Hub metadata fetch for `docker/dockerfile:1` to both the sandbox and the devcontainer builds.

scripts/agentic-sandbox.py
```diff
@@ -13,14 +13,6 @@
 
 IMAGE_NAME = "cc-protocol-sandbox"
```
philipcraig, line 14: An image built from the pre-PR Dockerfile has the same name, passes `image_exists`, and is reused with the volume mounts gone and only empty `mkdir`'d cache directories inside, so every run starts cold until the contributor guesses `--rebuild-docker`. The PR body carries the migration note but nothing in the tree does, and the deleted CONTRIBUTING sentence was the only in-repo record of the volume names. A one-line CONTRIBUTING note (rebuild once; `docker volume rm cc-protocol-uv-cache cc-protocol-bazel-repository-cache`) covers it, or an image label the script checks.

CONTRIBUTING.md
```diff
@@ -253,17 +253,14 @@ plain shell in the container.
...
-in Docker named volumes, `cc-protocol-uv-cache` and
-`cc-protocol-bazel-repository-cache`. To reset either volume, run `docker volume
-rm <name>`.
-
-The sandbox image also carries a googletest checkout at the tag pinned in
-`CMakeLists.txt`, so CMake builds need no network. Bumping that tag needs
-`--rebuild-docker`, as for `.bazelversion`.
+The sandbox image stores `uv`'s package cache and venv, Bazel's repository
+cache, and a googletest checkout at the tag pinned in `CMakeLists.txt`, so
+builds and installs need no network. Any repository change needs
```
philipcraig, line 251: Wrong in both directions. Source edits never need a rebuild (the checkout is bind-mounted rw at runtime), and a dependency bump without a rebuild degrades to an online fetch rather than breaking. "installs need no network" also overstates it: pre-commit's five hook repos go to `~/.cache/pre-commit`, which was never a volume and is not baked, so `uv run pre-commit run --all-files` still needs the network in a fresh container. Lines 20-22 keep a separate `.bazelversion` rule, so the guidance now lives in two places that disagree.

Suggest listing the trigger files (`.bazelversion`, `MODULE.bazel`/`.lock`, `pyproject.toml`/`uv.lock`, the googletest tag), saying the fallback is an online fetch, noting the googletest tag is the one silent case, and folding in the lines 20-22 rule.

philipcraig (changes requested):
Three regressions (the bazelisk pin, the unfiltered bind mount, and the devcontainer losing the Bazel pre-warm), then layer-cache and documentation points. Inline.

docker/Dockerfile
jbcoe: Fixed. Thanks.

docker/Dockerfile
jbcoe: I worry that this will lead to maintenance headache so would prefer the `.dockerignore` approach above.

docker/Dockerfile
jbcoe: We should add `MODULE.bazel.lock` to `.dockerignore`.

jbcoe:
The scope of this PR is to change _how_ things are cached, not to change _when_ the Docker image gets rebuilt. 

Iff that is to be done (it probably should) it must be done in a follow-up PR.

docker/Dockerfile
jbcoe: Fixed. Thanks

philipcraig:
Agreed, that's a fair line. I'll take the rebuild-timing points out of this review and raise them separately: the stale-image check in `agentic-sandbox.py`, the CONTRIBUTING trigger list, the googletest tag reminder, and the layer ordering. `.dockerignore` is fine by me for the bind mount.

The one remaining point I'd still count as in scope is the `bazel version` pre-warm moving from `base` to `sandbox`, since the devcontainer loses it. The others (default repository cache, the `# syntax` line, the output-base comment) are optional.

The pin and `--lockfile_mode` fixes aren't on the branch yet (head is still a186166). I'll re-review once they're pushed.

jbcoe:
Some issues uploading changes but all should be uploaded now. Thanks for the review.

philipcraig (approved):
The three regressions from the last round are fixed, and `docker-sandbox` now checks offline what the PR is for. The remaining notes (the `GIT_TAG` extraction, `.dockerignore` gaps and two Dockerfile comments) are non-blocking and tracked in #445.

PR #408, opened 2026-09-16
Fix unused-parameter error

### LINKED ISSUE #400: `call_through_vtable` fails under `-DNDEBUG -Wextra -Werror`
`call_through_vtable` (vtable.hh:60, added in #388) uses its `protocol_object` parameter only inside an `assert`. With `NDEBUG` the parameter is unused, and every instantiation reports

```
vtable.hh:60:52: error: unused parameter 'protocol_object' [-Werror=unused-parameter]
```

under `-Wall -Wextra -Werror`. `main` compiles clean under the same flags because the check lived inside `member_function_thunk::operator()`, where `protocol_object` was a local also used for `->vtable_` and `->object_`. vtable.hh is an installed header, so this reaches downstream release builds. CI does not pass `-Wextra` anywhere, so it stays green.

Fix: `[[maybe_unused]]` on the parameter, or leave it unnamed and take the address inside the `if constexpr`.

Two related points worth deciding at the same time:

- The helper is a plain function template, so at `-O0` every protocol call pays an extra out-of-line body, a call and three argument spills. The three `--compilation_mode=dbg` sanitizer configs and the gcov coverage build all run at `-O0`. `[[gnu::always_inline]] inline` removes that.
- Adding `-Wextra` to `.bazelrc` and `cmake/xyz_add_test.cmake` would have caught this class of problem.

jbcoe:
Fix unused-parameter error under -DNDEBUG -Wextra

protocol_object was only referenced inside an assert, so with NDEBUG (which
compiles the assert away) it goes unused; mark it [[maybe_unused]]. Also
mark the function [[gnu::always_inline]] inline, since it's a plain
function template that otherwise pays an out-of-line call at -O0 on every
protocol call.

Add -Wextra to .bazelrc and cmake/xyz_add_test.cmake so this class of bug
is caught by CI, and suppress -Wmissing-field-initializers, which -Wextra
otherwise raises on every idiomatic partial-designated-initializer use of
std::meta::data_member_options. Fix the two -Wextra unused-parameter
findings this surfaced in tutorials/02_type_erasure.cc's null_vtable.

Fixes #400.


jbcoe:
Code coverage drop is sad but hard to fix without extending the scope of an intentionally tight change.

philipcraig (approved):

PR #410, opened 2026-09-16
Do not use dealias for parameter types

### LINKED ISSUE #397: Do not use dealias when reflecting function parameter types

jbcoe:
Do not use dealias for parameter types as dealiasing is performed implicitly.

closes #397


philipcraig (approved):
`type_of` on a parameter returns the dealiased type on GCC trunk and clang-p2996, and [meta.reflection.queries] specifies it as the type in the function's parameter-type-list, so these were no-ops. Keeping `dealias(return_type_of(...))` is right: GCC preserves the alias there.

PR #411, opened 2026-09-16
Consolidate the three function-pointer-type builders

jbcoe:


vtable.hh
```diff
@@ -66,10 +75,43 @@ template <std::meta::info Member, typename Vtable, typename ProtocolObject,
...
+//
+// `Member` is a template parameter rather than a `std::meta::info` function
+// parameter: cc1plus crashes on the latter (GCC ICE).
+template <std::meta::info Member>
+consteval std::meta::info function_pointer_type_of() {
+  std::vector<std::meta::info> fn_args{
+      std::meta::reflect_constant(is_noexcept(Member)),
+      dealias(return_type_of(Member))};
+  for (std::meta::info parameter : parameters_of(Member)) {
+    fn_args.push_back(dealias(type_of(parameter)));
```
philipcraig: Resolved by merge commit 54b116d after #410 landed: bare `type_of(parameter)` here and at line 110, and the comment above now says only the return type is dealiased. `type_of(parameter)` is already dealiased on both compilers, so the generated types are unchanged.

vtable.hh
```diff
@@ -66,10 +75,43 @@ template <std::meta::info Member, typename Vtable, typename ProtocolObject,
...
+  }
+  return substitute(^^fn_ptr_t, fn_args);
+}
+
+// As above, but with `leading_parameter_type` as the function pointer's
+// first parameter, ahead of `Member`'s own parameters; a vtable entry uses
+// this for its type-erased object parameter.
+template <std::meta::info Member>
```
philipcraig, line 103: These two overloads differ by one initialiser element, and drift between them is silent: the thunk's partial specialisation matches whatever its own overload returns, and the two types only meet at the ordinary call through the vtable entry, which accepts implicit conversions. Editing only this overload so `void scale(double)` gets an `int` thunk parameter compiles with no diagnostic and `p.scale(2.75)` stores 2. A noexcept-only drift compiles and terminates at runtime.

One function with `std::initializer_list<std::meta::info> leading_parameter_types = {}` appended after the return type covers both callers; `generate_vtable_specs` passes `{is_const(member) ? ^^const void* : ^^void*}`. Compiles on GCC trunk and clang-p2996.

vtable.hh
```diff
@@ -66,10 +75,43 @@ template <std::meta::info Member, typename Vtable, typename ProtocolObject,
...
+      find_vtable_entry<^^std::remove_cv_t<VtablePtr>, Member>();
   return vtable->[:vtable_entry:](object, std::forward<Args>(args)...);
 }
 
+// Builds the function-pointer type `R(*)(Args...) noexcept(...)` from
+// `Member`'s return type, parameter types and noexcept-ness, with `dealias`
+// applied throughout so an aliased type and its underlying type produce the
+// same function-pointer type.
+//
+// `Member` is a template parameter rather than a `std::meta::info` function
```
philipcraig, line 87: Which ICE? A non-template `function_pointer_type_of(std::meta::info member, std::vector<std::meta::info> leading = {})` with the three callers updated compiles all five test TUs on GCC trunk 17.0.0 20260816 (-O0/-O2/-O3, ASAN, UBSAN, -Wall -Wextra -Werror) and on clang-p2996 bde8bcf3. The pre-PR `generate_vtable_specs` already built the same vector from a plain `std::meta::info` loop variable, and the second overload below takes a `std::meta::info` parameter itself. The template shape has a cost: callers must supply `Member` as a constant, a plain range-for cannot call it, and each `Member` is a separate specialisation. A GCC PR number and the failing form would let a maintainer know when to retire this; otherwise the non-template form also resolves the two-overload comment above.

vtable.hh
```diff
@@ -43,6 +43,8 @@ constexpr inline auto mangled_name_of =
     std::define_static_string(xyz::name_mangling::mangle(Member));
 
 // The `VtableType` entry named after the mangled member function `Member`.
+// Searches base classes too: `protocol<I, Alloc>::vtable` derives from
```
philipcraig, line 46: `bases_of` returns direct bases only, and derived members are returned ahead of base members, so this is a one-level derived-first scan. Only `protocol<I, Alloc>`'s thunks pass the derived type; `make_view_vtable` and both `protocol_view` classes pass `vtable_t<I>` itself. Neither a deeper hierarchy nor a name clash is reachable today, so this is wording: "Searches direct bases too: `protocol<I, Alloc>::vtable` derives from `vtable_t<I>` and its thunks pass that derived type; `make_view_vtable` and `protocol_view` pass `vtable_t<I>` directly."

vtable.hh
```diff
@@ -66,10 +75,43 @@ template <std::meta::info Member, typename Vtable, typename ProtocolObject,
            "cannot call member function of valueless protocol");
   }
   constexpr std::meta::info vtable_entry =
-      find_vtable_entry<^^Vtable, Member>();
+      find_vtable_entry<^^std::remove_cv_t<VtablePtr>, Member>();
```
philipcraig, line 78: This keys the lookup on `protocol<I, Alloc>::vtable` per allocator and, for views, on the alias-specialisation reflection `^^std::remove_cv_t<const vtable_t<I>>`, which GCC treats as a different template argument from `^^vtable_t<I>` (mangled `Dmta` vs `Dmty`). The derived scan also always misses before reaching the base, since the derived type holds only destroy/copy/move. `const Vtable*` in the signature (see the review body) removes this; if the deduction stays, `dealias(^^std::remove_cv_t<VtablePtr>)` at least restores sharing between the two views and `make_view_vtable`.

member_function_thunks.hh
```diff
@@ -51,7 +49,7 @@ struct member_function_thunk<R (*)(Args...) noexcept(IsNoexcept), EnclosingType,
   {
     auto* enclosing = reinterpret_cast<EnclosingType*>(this);
     auto* protocol_object = static_cast<ProtocolType*>(enclosing);
-    return call_through_vtable<Member, Vtable>(
+    return call_through_vtable<Member>(
```
philipcraig, line 52: Nothing checks the invariant the consolidation exists for: thunk type equals vtable entry type minus its leading `void*` or `const void*`. A class-scope `static_assert` here and in the `operator_thunk` specialisation, comparing `typename[:type_of(find_vtable_entry<^^Vtable, Member>()):]` with `R(*)(const void* | void*, Args...) noexcept(IsNoexcept)` keyed on `is_const(Member)` (not `IsConst`, since `protocol_view<Shape>` instantiates const-member thunks with `IsConst=false`), is the first error on both drifts I tried. A `static_assert` inside `call_through_vtable` does not work: `function_pointer_type_of` returns the `fn_ptr_t` alias reflection and `type_of(entry)` the dealiased type, and adding `dealias` makes it a tautology against the overload that built the entry.

The new base-class search also has no test. A `vtable_tests.cc` case with a local `struct Derived : vtable_t<I>` asserting `find_vtable_entry<^^Derived, ^^I::f>() == find_vtable_entry<^^vtable_t<I>, ^^I::f>()` passes on both pins with no duplicate-symbol issue, which would also let `vtable_entry_for` and its GCC caveat go.

member_function_thunks.hh
```diff
@@ -85,15 +83,8 @@ template <std::meta::info Member, bool IsConst, typename EnclosingType,
...
-  // Build the function-pointer type R(*)(Args...) noexcept(...) from the
-  // method's return type, parameter types and noexcept-ness.
   static consteval std::meta::info fn_ptr_type() {
-    std::vector<std::meta::info> fn_args{
-        std::meta::reflect_constant(is_noexcept(Member)),
-        dealias(return_type_of(Member))};
-    fn_args.append_range(parameters_of(Member) |
-                         std::views::transform(std::meta::type_of));
-    return substitute(^^fn_ptr_t, fn_args);
+    return function_pointer_type_of<Member>();
```
philipcraig, line 87: One-line forwarder with a single caller three lines below (same in `operator_thunk_for`). `function_pointer_type_of<Member>()` directly in the `substitute` argument list, and these two statics deleted, compiles with -Wall -Wextra -Werror on both compilers.

philipcraig (commented):
One item from the first version of this review is resolved and the rest stands.

**#410 landed while this was open.** The parameter `dealias` in both `function_pointer_type_of` overloads was the same one #410 removed from the block this PR deletes, so merge commit 54b116d drops it here too and rewords the comment. Only `dealias(return_type_of(Member))` remains, since GCC keeps the alias there.

**A shorter route to the same consolidation.** With `call_through_vtable(ProtocolObject*, const Vtable* vtable, Object, Args&&...)` and `Vtable` explicit, as the thunks already pass it, derived-to-base pointer conversion keeps the lookup on `vtable_t<I>`. That removes the base-class walk in `find_vtable_entry`, the `remove_cv_t<VtablePtr>` key, and the per-allocator and per-view lookup specialisations it creates (98 on this head vs 14 on main for a 14-member interface over five allocators and both views, though wall time is within noise today). It also gives the `Vtable` template parameter a reader again. After this PR nothing reads it: its only consumers were the 64 `call_through_vtable<Member, Vtable>` calls, yet it is still declared in every thunk and overload-set specialisation, the six friend declarations in `protocol.hh`, three base specifiers and `test_helpers.h`. I checked that form against all five test TUs on GCC trunk and clang-p2996 with the thunk call sites unchanged.

If the walk stays, the comment at `protocol.hh:172` needs updating: it still says the thunks take a `const view_vtable*` and call through a `const vtable*` unchanged, but the lookup now happens in the derived type and depends on the walk finding the base. `nonstatic_data_members_of(^^protocol<I, Alloc>::vtable)` yields only destroy/copy/move, so anyone simplifying `find_vtable_entry` on the strength of that comment breaks every owning-protocol member call while the view tests still pass.

jbcoe:
This got bloated and confused. I'll split and re-post.

PR #412, opened 2026-09-16
Fix the Node.js 20 deprecation warnings in CI

philipcraig:
GitHub has deprecated the Node.js 20 runtime, so every job using an action that targets it emits a warning ([example](https://github.com/jbcoe/cc-protocol/actions/runs/35153119519/job/104986057221?pr=411)). This moves every affected action to a Node.js 24 release with the same inputs:

- `actions/cache`, `actions/cache/restore` and `actions/cache/save` v4 → v6
- `dorny/paths-filter` v3 → v4
- `codecov/codecov-action` v5 → v7 (v6 bumped its internal `github-script`)
- the pre-commit job drops `actions/setup-python` v3 and `pre-commit/action` (which pins `actions/cache` v4 internally) and runs pre-commit through uv, as the other jobs already do


jbcoe (approved):
Thanks for this.

Issue #404, opened 2026-09-16
Share interfaces across the operator tests to cut compile time


philipcraig:
Each of the ~38 tests in protocol_operator_tests.cc declares its own `Interface`/`Conforming` pair and instantiates `protocol`, `protocol_view` and `protocol_view<const>`, so the TU drives about 132 separate `define_aggregate` vtable and wrapper generations.

| Build | Time |
|---|---|
| gtest + protocol.hh baseline, `-O2 -c` | 1.1 s |
| protocol_operator_tests.cc, `-fsyntax-only` | 4.3 s |
| protocol_operator_tests.cc, `-O2 -c` | 6.0 s |

Roughly 3 s of that is instantiation, paid in every job that builds the tests (cmake, bazel, three sanitizer configs, runtime coverage, clang-tidy) and more under clang-p2996. A matched experiment with 22 binary operators through all three wrapper kinds costs 1.5 s above baseline with one interface per test and 0.7 s with one shared interface.

This is a test-structure optimisation, not a design question: per-test interfaces keep each test self-contained, and a shared interface per operator family (arithmetic, bitwise, compound assignment, shifts) keeps most of that while halving the work. Related to the CI-time work in #374.

philipcraig:
Re-measured on main (ba3ee12) with GCC trunk 20260906 from `/opt/gcc-latest` on a 2-core GitHub Codespaces VM, median of three runs:

| Build | `-fsyntax-only` | `-O2 -c` |
|---|---|---|
| gtest + protocol.hh baseline | 4.6 s | 4.8 s |
| protocol_operator_tests.cc | 15.2 s | 26.4 s |
| protocol_test.cc | 29.7 s | 50.5 s |

Matched experiment, 20 binary operators through all three wrapper kinds:

| Layout | `-fsyntax-only` | `-O2 -c` |
|---|---|---|
| one interface per test | 8.9 s | 13.0 s |
| shared, 5 operators per interface | 6.9 s | 10.1 s |
| shared, 20 operators on one interface | 6.8 s | 10.0 s |

Family size makes no difference, so most of the cost is per member function: about 0.26 s per const/non-const operator pair against 0.17 s per interface. Collapsing the 35 operator tests to about six family interfaces would save about 5 s of 26.4 s. `protocol_test.cc` takes 50.5 s on its own, so this TU is off the critical path of a parallel build and the wall-clock saving per job is close to zero.

[...]

PR #413, opened 2026-09-17
Consolidate function-pointer-type builders

### LINKED ISSUE #405: Three copies of the function-pointer-type builder
The `R(*)(Args...) noexcept(...)` builder exists three times:

- `operator_thunk_for::fn_ptr_type()` in operator_thunks.hh:63
- `member_function_thunk_for::fn_ptr_type()` in member_function_thunks.hh:90
- inline in `generate_vtable_specs`, vtable.hh:93-101

They have already diverged: vtable.hh wraps each parameter type in `dealias`, the two thunk copies use bare `type_of`. Nothing miscompiles today because alias reflections substitute to the same specialisation, but all three must agree for a thunk's partial specialisation to match the vtable entry's type, and a change to one (#397 is one such change) surfaces only as a `find_vtable_entry` or specialisation mismatch deep in an instantiation.

vtable.hh is included by both thunk headers, so one `function_pointer_type_of(std::meta::info)` there can replace all three. While there: `call_through_vtable` takes `Vtable` as an explicit template argument although it is already deducible from the `VtablePtr*` parameter, so a caller can search one vtable type and splice into a pointer of another.

jbcoe:
Consolidate function-pointer-type builders. 

We leave the vtable function pointer building logic inline as it is different.

Fixes #405


vtable.hh
```diff
@@ -86,16 +101,13 @@ consteval std::vector<std::meta::info> generate_vtable_specs() {
 
   template for (constexpr std::meta::info member :
                 protocol_interface_functions_of<interface_type>) {
-    // Build the function-pointer type R(*)(void*, Args...) noexcept(...)
-    // from the method's return type, parameter types and noexcept-ness; a
-    // const method takes `const void*` instead, matching the constness of
-    // the access path it's called through.
     std::vector<std::meta::info> fn_args{
```
philipcraig, line 104: This is still an independent copy of the builder, differing from `function_pointer_type_of` by the one `push_back` of `^^void*` or `^^const void*`, so the two have to be kept in step by hand.

#405 says a mismatch surfaces as a specialisation mismatch deep in an instantiation. It doesn't: it is silent. The thunk's partial specialisation matches whatever its own builder returns, and the two types only meet at the ordinary call through the vtable entry, which accepts implicit conversions. On this head, changing only `function_pointer_type_of` so that one member's `double` parameter becomes `int` compiles with no diagnostic, and `p.scale(2.75)` through a `protocol` stores 2.

One function covers all three callers:

```cpp
consteval std::meta::info function_pointer_type_of(
    std::meta::info member,
    std::initializer_list<std::meta::info> leading_parameter_types = {}) {
  std::vector<std::meta::info> fn_args{
      std::meta::reflect_constant(is_noexcept(member)),
      dealias(return_type_of(member))};
  fn_args.insert(fn_args.end(), leading_parameter_types);
  for (std::meta::info parameter : parameters_of(member)) {
    fn_args.push_back(type_of(parameter));
  }
  return substitute(^^fn_ptr_t, fn_args);
}
```

with this block reduced to

```cpp
std::meta::info fn_ptr_type = function_pointer_type_of(
    member, {is_const(member) ? ^^const void* : ^^void*});
```

That builds and passes the Bazel tests on GCC trunk at this head.

vtable.hh
```diff
@@ -70,6 +70,21 @@ template <std::meta::info Member, typename Vtable, typename ProtocolObject,
   return vtable->[:vtable_entry:](object, std::forward<Args>(args)...);
 }
 
+// The function-pointer type `R(*)(Args...) noexcept(...)` for `Member`.
+//
+// `Member` is a template parameter rather than a `std::meta::info` function
+// parameter: cc1plus crashes on the latter (GCC ICE).
```
philipcraig, line 76: Which ICE? The non-template form in the comment below takes `member` as a `std::meta::info` function parameter and compiles all the test TUs on GCC trunk at this head; the same form compiled on clang-p2996 bde8bcf3 when I tried it on #411. If there is a failing form, a GCC PR number here (and an entry on #277) would tell a maintainer when the workaround can go. Otherwise the template parameter costs a specialisation per interface member (#404) and stops a plain range-for from calling the helper.

vtable.hh
```diff
@@ -70,6 +70,21 @@ template <std::meta::info Member, typename Vtable, typename ProtocolObject,
...
+template <std::meta::info Member>
+consteval std::meta::info function_pointer_type_of() {
```
philipcraig, line 78: This builds a specialisation of `fn_ptr_t`, which is declared in `overload_spec.hh`, and nothing in it is vtable-specific. Both thunk headers already include `overload_spec.hh`, as does this one. Next to `fn_ptr_t` it keeps the `IsNoexcept, R, Args...` argument order and the code that depends on it in one file.

member_function_thunks.hh
```diff
@@ -85,15 +83,8 @@ template <std::meta::info Member, bool IsConst, typename EnclosingType,
...
-  // method's return type, parameter types and noexcept-ness.
   static consteval std::meta::info fn_ptr_type() {
-    std::vector<std::meta::info> fn_args{
-        std::meta::reflect_constant(is_noexcept(Member)),
-        dealias(return_type_of(Member))};
-    fn_args.append_range(parameters_of(Member) |
-                         std::views::transform(std::meta::type_of));
-    return substitute(^^fn_ptr_t, fn_args);
+    return function_pointer_type_of<Member>();
   }
```
philipcraig, line 88: Carried over from #411: `fn_ptr_type()` is now a one-line forwarder with a single caller just below (same in `operator_thunk_for`). Calling `function_pointer_type_of` directly in the `substitute` argument list removes both statics and the second name for the same thing.

philipcraig (changes requested):
The split is much easier to read than #411, and the helper builds the same `substitute` arguments as the two copies it replaces. The main ask is to fold the third copy in as well: it is a one-element difference, and drift between the two remaining builders is silent rather than a compile error (details inline). With that done `Fixes #405` is fair for the duplication; the explicit-`Vtable` item in #405 is still open, so either keep the issue open or split that item out.

If the inline copy stays, a class-scope `static_assert` in both thunk specialisations comparing `typename[:type_of(find_vtable_entry<^^Vtable, Member>()):]` with `R(*)(const void* | void*, Args...) noexcept(IsNoexcept)`, keyed on `is_const(Member)` rather than `IsConst`, would turn that drift into an error.

philipcraig:
#437 covers this and the rest of #405, so I think this one can close in its favour.

It uses one non-template `function_pointer_type_of(std::meta::info member, std::initializer_list<std::meta::info> leading = {})` next to `fn_ptr_t` in `overload_spec.hh`. The two thunks call it with no `leading` types and `generate_vtable_specs` passes the `void*` or `const void*`, so the third copy goes as well and the `fn_ptr_type()` forwarders are inlined. Taking `std::meta::info` as a function parameter compiles on GCC 16, GCC trunk and clang-p2996, so the ICE note is not needed.

#437 also removes the fourth copy that #421 added in `conversion_thunks.hh`, and changes `call_through_vtable` to take `const Vtable*`.

PR #416, opened 2026-09-17
Install Claude Code with the native installer

philipcraig:
Rebuilt images always came up with Claude Code 2.1.197. The image has Node 20, and since 2.1.198 the npm package declares `engines: node >=22`, so a bare `npm install -g @anthropic-ai/claude-code` falls back to the last release that accepts Node 20.

- `docker/Dockerfile`: install with `curl -fsSL https://claude.ai/install.sh | bash` (the [recommended method](https://code.claude.com/docs/en/setup)). It installs `claude` into `~/.local/bin`, needs no Node.js and updates itself in the background. Node.js and the npm prefix setup are removed; nothing else in the image used them.
- `scripts/agentic-sandbox.py`: `--update` runs `claude update` instead of `npm install -g …@latest`. Each agent now carries its own update command.


jbcoe (approved):
Thanks for this.

PR #417, opened 2026-09-17
Make --dangerously-skip-permissions opt-in in the sandbox

philipcraig:
`scripts/agentic-sandbox.py claude` always ran `claude --dangerously-skip-permissions`. The container keeps the agent away from the rest of the host file system, but the network is unrestricted and the project and `~/.claude`, credentials included, are mounted read-write. The [devcontainer docs](https://code.claude.com/docs/en/devcontainer#run-without-permission-prompts) pair that flag with egress restrictions, which the sandbox does not have.

- `claude` now starts without the flag, in the permission mode from the mounted host `~/.claude/settings.json` (auto mode, for example).
- `--skip-permissions` restores the old behaviour; it is rejected for other agents.
- `CONTRIBUTING.md` documents the option and what the sandbox does and does not isolate.

Touches the same `AGENT_CLIS` entry as #416; whichever lands second needs a small rebase.


jbcoe (approved):
Thanks for this.

PR #418, opened 2026-09-17
Build and lint the sandbox Docker stage in CI

### LINKED ISSUE #325: Nothing builds or lints the sandbox Docker stage
No CI job builds `docker/Dockerfile`. The devcontainer builds the `base`
target only, so the `sandbox` stage (currently the cache-volume setup) is
exercised solely by contributors running `scripts/agentic-sandbox.sh`
with `--rebuild-docker`. Dockerfile linting is a commented-out TODO in
`.pre-commit-config.yaml`.

An edit that breaks the sandbox stage (a quoting change in the
`~/.bazelrc` line, a renamed variable in one of its two copies) keeps CI
green and shows up as a silent re-download on the next contributor's
rebuild.

Suggested cover:

- a `docker build --target sandbox` step in CI, gated on changes to
  `docker/` and `.bazelversion`;
- the hadolint pre-commit hook;
- an `--offline` option on `agentic-sandbox.py` that adds
  `--network none`, so the offline-build claim in #304 is reproducible.

philipcraig:
Closes #325.

- New `Docker` workflow builds `docker/Dockerfile` to the `sandbox` target on
  changes to `docker/`, `.bazelversion` or the workflow. With `--network none`
  it then checks that `bazel info repository_cache` reports the sandbox
  directory, that the cache directories are writable by `vscode`, and that the
  baked googletest tag matches `GIT_TAG` in `CMakeLists.txt`. Not a required
  check, so the trigger has a `paths:` filter.
- hadolint pre-commit hook replaces the commented-out TODO. It uses
  `AleksaC/hadolint-py`, which installs the binary from PyPI, so it needs
  neither Docker nor a system hadolint. Rules ignored by design are in
  `.hadolint.yaml`; the rest are fixed in the Dockerfile (`pipefail` shell,
  quoted URLs, `wget -nv`, `mkdir -p -m`).
- `agentic-sandbox.py --offline` runs the plain-shell container with
  `--network none`.


jbcoe (approved):
Thanks for this.

PR #419, opened 2026-09-17
Add jq and ripgrep to the image and tool notes to AGENTS.md

### LINKED ISSUE #409: Add tools and tool descriptions to the AI sandbox
Providing tools (like clang-query) along with suitable descriptions could help the agent make and test changes faster.

https://platform.claude.com/docs/en/agents-and-tools/tool-use/overview

philipcraig:
Adds `jq` and `ripgrep` to the image, and three lines to `AGENTS.md`: how to
rerun one test from `bazel-bin`, and that released Clang tools cannot parse
the reflection sources.

`clang-query` itself is not added. `clang-query-22` stops at
`unknown argument: '-freflection'` and, without the flag, at
`no member named 'meta' in namespace 'std'`. A working one would have to come
from clang-p2996, which the image does not carry and which takes hours to
build.

The agents in the sandbox are CLIs that use the shell, so a tool's
description is its `AGENTS.md` entry; the API tool definitions linked from
the issue apply only to a custom agent built on the API.

Closes #409


jbcoe (approved):
Thanks for this.

Issue #402, opened 2026-09-16
Conversion functions on an interface are neither forwarded nor rejected


philipcraig:
`conformance_candidate_infos` keeps a member when `has_identifier(member) || is_operator_function(member)`. A conversion function `operator T()` has no identifier and `is_operator_function` is false for it, so it is dropped before conformance or rejection ever sees it:

```cpp
struct I { int f() const; explicit operator bool() const; };
struct C { int f() const { return 0; } };            // no operator bool
static_assert(is_protocol_conformant_v<I, C>);       // passes
protocol<I> p{C{}};                                  // builds
static_cast<bool>(p);                                // ill-formed even if C did declare operator bool
```

So an interface author can put a conversion in the contract, have a type without it accepted, and find out at the call site. This predates #388, but #388 established the rule that every non-forwardable operator throws from `protocol_interface_function_infos` with a message naming it, and conversions are now the only category outside that rule.

Options: reject them there with the same style of diagnostic, forward them (they are nullary and const-qualifiable, so the machinery is close), or document the gap.

philipcraig:
#421 forwards conversion functions, but `operator auto() const;` on an interface is still silently ignored. `members_of` does not return a function with an undeduced return type, so it never reaches the `is_conversion_function` filter:

```cpp
struct I { operator auto() const; };
struct Empty {};
struct C { operator int() const { return 1; } };
static_assert(is_protocol_conformant<I, Empty>());  // passes
protocol<I> p(C{});                                 // builds
static_cast<int>(p);                                // "invalid static_cast"
```

This is separate from the interface conversion-template case deferred in the #421 review. It is the same shape as this issue: neither forwarded nor rejected.

Issue #259, opened 2026-08-31
Support more parameter types in vtable name mangling


philipcraig:
#256 intentionally leaves some parameter types unsupported: using one in an interface member function fails noisily at compile time (a consteval throw such as "name mangling: unsupported parameter type") rather than mis-conforming or misbehaving at runtime.

Currently unsupported, all previously accepted by the index-based naming #256 replaces:

- pointer-to-member types (Itanium `M<class><member type>`)
- non-type template arguments other than `bool`, the standard integral types, and enum values equal to a named enumerator — floating-point, class-type, pointer and reference arguments, template-template arguments, and enum values matching no enumerator (bitmask values, any `std::byte` argument)
- unnamed types: closure types, typedef'd anonymous enums

Filed from the #256 review discussion ("We will leave these as unsupported for now") so support can be added as demand arises.

PR #421, opened 2026-09-18
Forward conversion functions on protocol interfaces (#402)

jbcoe:
Forward conversion functions on protocol interfaces with a new kind of thunk and overload set templated on the return type of the conversion.

Follow up work is needed to better handle `explicit` and conversion to bool in a valueless state.

Tracked in #426 #427


conformance.hh
```diff
@@ -29,15 +29,19 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
...
+// conversion function's name is its (dealiased) target type.
 consteval bool same_name(std::meta::info a, std::meta::info b) {
   if (has_identifier(a) && has_identifier(b)) {
     return identifier_of(a) == identifier_of(b);
   }
   if (is_operator_function(a) && is_operator_function(b)) {
     return operator_of(a) == operator_of(b);
   }
+  if (is_conversion_function(a) && is_conversion_function(b)) {
+    return dealias(return_type_of(a)) == dealias(return_type_of(b));
```
philipcraig, line 44: This makes `same_name_and_parameters` true for two conversions to the same target, so the `all_const` dedup in `is_forwarded_member_function` (protocol_wrappers.hh:70) drops the const overload whenever a non-const one exists, without comparing explicitness:

```cpp
struct I {
  explicit operator bool();
  operator bool() const;
};
struct C { /* the same two members */ };

C c;
bool a = c;              // OK: the const overload
protocol<I> p(C{});
bool b = p;              // OK
protocol_view<I> pv(c);
bool d = pv;             // error: explicit conversion function was not considered
```

The view gets only the explicit non-const thunk. A single const thunk cannot reproduce what the object does here (implicit contexts pick the const overload, `static_cast` picks the non-const one), so I would reject a same-target pair that differs in explicitness when building `protocol_view<I>`, with a diagnostic naming the conversion. Either way the mixed pair wants a test.

conformance.hh
```diff
@@ -142,16 +152,17 @@ concept is_maybe_lambda =
...
 template <std::meta::info Type>
 consteval auto conformance_candidate_infos() {
   auto named =
       members_of(Type, std::meta::access_context::unprivileged()) |
       std::views::filter(std::meta::is_function) |
       std::views::filter([](std::meta::info member) consteval {
         return !is_special_member_function(member) &&
-               (has_identifier(member) || is_operator_function(member));
+               (has_identifier(member) || is_operator_function(member) ||
+                is_conversion_function(member));
```
philipcraig, line 165: A conversion function *template* on an interface still falls out at the `is_function` filter, so it is neither forwarded nor rejected, which is the shape #402 describes:

```cpp
struct I {
  int f() const;
  template <typename T>
  explicit operator T() const;
};

static_assert(is_protocol_conformant<I, C>());  // passes
protocol<I> p(C{});
static_cast<bool>(p);                           // error: invalid static_cast
```

Named member function templates on an interface are dropped the same way today, so this may belong in a follow-up that rejects every interface member template (`is_function_template` / `is_conversion_function_template`) rather than here. `TemplatedConversionFunctionsAreIgnored` pins the candidate side only.

conversion_thunks.hh
```diff
@@ -0,0 +1,129 @@
...
+#include "vtable.hh"
+
+namespace xyz::detail {
+
+// Thunk for one overload of a synthesised conversion function. A conversion
+// function can't be reached through a named member, so
+// `conversion_overload_set` derives from this thunk directly instead of
+// holding it as a data member, letting `ProtocolType` be recovered with a
+// plain static_cast.
+template <typename FnPtrType, typename ProtocolType, typename Vtable,
```
philipcraig, line 36: `FnPtrType` and `IsExplicit` are both derivable from `Member`; the function-pointer round trip exists only to pattern-match `R` and `IsNoexcept`, since there is no `Args...` pack to deduce. Computing them in the thunk works on GCC trunk:

```cpp
using R = typename[:dealias(return_type_of(Member)):];
explicit(is_explicit(Member)) operator R() noexcept(is_noexcept(Member))
```

That makes the primary template the definition, removes the `static_assert(false)` stub and the partial specialisation, and removes `fn_ptr_type()` from this header, which is otherwise a fourth copy of the builder for #405. It also rules out an `IsExplicit` that disagrees with `is_explicit(Member)`.

conversion_thunks.hh
```diff
@@ -0,0 +1,129 @@
...
+struct conversion_thunk {
+  static_assert(false,
+                "Unspecialized conversion_thunk cannot be instantiated.");
+};
+
+template <typename R, bool IsNoexcept, typename ProtocolType, typename Vtable,
+          std::meta::info Member, bool IsConst, bool IsExplicit>
+struct conversion_thunk<R (*)() noexcept(IsNoexcept), ProtocolType, Vtable,
+                        Member, IsConst, IsExplicit> {
+  explicit(IsExplicit) operator R() noexcept(IsNoexcept)
```
philipcraig, line 47: With a forwarded `explicit operator bool`, `if (p)` now compiles on `protocol<I>` and reads like the emptiness check on `unique_ptr`, `optional` or `function`. On a moved-from protocol it is a forwarded call: it asserts in debug and calls the null vtable entry under `NDEBUG`. The shared death test covers the guard; a sentence in the docs that `valueless_after_move` is the only safe query would cover the expectation.

name_mangling.hh
```diff
@@ -236,16 +236,20 @@ consteval std::string mangle_type(std::meta::info type) {
...
+// its identifier otherwise. `identifier_of` throws for an operator or
+// conversion function, neither of which has an identifier.
 //
 // `operator++`/`operator--` mangle to the same atom (`pp`/`mm`) whether
 // prefix or postfix; overload resolution and linkage tell them apart via
 // the dummy `int` parameter C++ requires on the postfix form, which
 // `mangle`'s parameter loop already encodes.
 consteval std::string base_name_of(std::meta::info function) {
+  if (is_conversion_function(function)) {
+    return "cv" + mangle_type(return_type_of(function));
```
philipcraig, line 251: The target type is mangled twice: here in the `cv<type>` atom and again as the return type `mangle()` always appends (`fn_NKcvbEb` in the new test). For `operator std::string() const` the entry name is 152 characters, with the `basic_string<char, char_traits, allocator>` mangling in it twice. Since `mangle()` already folds in the return type, `cv` alone is collision-free; Itanium instead omits the return type for conversion functions.

Separately, an unsupported target (a pointer to member, #259) now reports "unsupported parameter type" for a function with no parameters.

protocol_operator_tests.cc
```diff
@@ -408,6 +410,102 @@ TEST(ReflectionProtocolTest, OperatorLogicalNot) {
...
+
+    explicit operator int() const noexcept { return 42; }
+  };
+
+  protocol<Interface> p(Conforming{});
+  EXPECT_TRUE(static_cast<bool>(p));
+  EXPECT_EQ(static_cast<int>(p), 42);
+}
+
+TEST(ReflectionProtocolTest, NonExplicitConversionIsImplicit) {
```
philipcraig, line 459: The implicit case is only exercised on `protocol<>`; the views are only reached through `static_cast` on explicit conversions. Also untested: a non-`noexcept` conversion, an alias target against its underlying type (`using S = std::string; operator S() const` on the interface, `operator std::string() const` on the candidate, which is what the `dealias` in `same_name` is for), and a class or reference target, which takes `mangle_type`'s qualified-name path under the new `cv` atom.

protocol_wrappers.hh
```diff
@@ -156,6 +159,8 @@ consteval std::meta::info generate_member_bases_wrapper() {
       member_base_args.push_back(reflect_constant(member));
     } else if (is_operator_function(member)) {
       member_base_args.push_back(reflect_constant(operator_of(member)));
+    } else if (is_conversion_function(member)) {
```
philipcraig, line 154: This chain and the one below now carry the same three-way test, and this one has no `else`: a member kind added to the second chain but missed here loses its key argument and fails inside `substitute()` rather than at `std::unreachable()`. One chain choosing a (template, key argument) pair would give a single failure point.

philipcraig (changes requested):
Conversion forwarding works for the cases the tests cover: reference, class, alias and cv-qualified targets all dispatch correctly. One defect, inline on `same_name`: with an `explicit` non-const and an implicit const conversion to the same target, `protocol_view<I>` loses the implicit conversion that the viewed object and `protocol<I>` both have.

The other inline notes are smaller. One that is not on a changed line: the comment on `protocol_interface_function_infos` (conformance.hh:202) still says "member functions and call operators"; the matching comment on `conformance_candidate_infos` was updated.

conformance.hh
jbcoe: I'm not 100% sure what to do here. Requiring `explicit` to match is probably the easiest most conservative approach for now, let's go with that.

Follow-up design issues raised as #426

conformance.hh
jbcoe: This is, as you say, out of scope for this PR.

conversion_thunks.hh
jbcoe: This is a significant issue that we need some thought on. I'll mark is as out of scope for this PR but with a critical design bug to address. As-is the API is very dangerous.

Raised #427

name_mangling.hh
jbcoe: This out of scope, it a general mangling issue. We can probably drop return type across all mangling (as Itanium does) but not in this PR.

conformance.hh
jbcoe: Marking resolved as we now require `explicit` to match an we have raise a follow-up design isssue.

conversion_thunks.hh
jbcoe: Marking as resolved as the code does as intended. Intent needs some further thought and will be addressed in #427. Completely removing conversion operators is a valid solution.

protocol_wrappers.hh
jbcoe: Added another `unreachable`. This function is already flagged for a rewrite `TODO: Rewrite this hard-to-read function.` and handling conversions makes it minimally different.

conversion_thunks.hh
jbcoe: I'll defer this to when we clean up the function builder in #405.

protocol_operator_tests.cc
jbcoe: I've added new tests. We need a test audit in the near future as I'm pretty sure that some tests don't exercise new code paths and may be unnecessary and redundant.

jbcoe:
Thanks for the review. Some cleanup applied. Some more conservative design put in place and a critical design issue raised for the conversion to bool of a valueless object.

conformance.hh
```diff
@@ -123,7 +127,10 @@ consteval bool member_function_conforms_to(std::meta::info candidate,
     if (is_const(interface) != is_const(candidate)) return false;
   }
   // If interface is `noexcept`, `candidate` must be noexcept.
-  return !is_noexcept(interface) || is_noexcept(candidate);
+  if (is_noexcept(interface) && !is_noexcept(candidate)) return false;
+
+  // `explicit` qualifiers must match.
+  return is_explicit(interface) == is_explicit(candidate);
```
philipcraig, line 133: This now rejects an implicit candidate for an explicit interface conversion: `struct I { explicit operator bool() const; }` with `struct C { operator bool() const; }` was conformant at 570ebd2 and no longer is. That pairing is safe: the wrapper takes `explicit(...)` from the interface and the trampoline calls the candidate member directly, so the candidate cannot widen what the protocol exposes. It is also the opposite of the noexcept rule two lines up, where the candidate may be stricter. The earlier one-directional rule (reject an explicit candidate for an implicit interface) was enough; `ConversionFunctionExplicitnessMustMatch` pins the stricter one.

philipcraig (commented):
Second pass, at 5936f9f.

**Mixed explicitness still fails (first review, now #426).** With

```cpp
struct I { explicit operator bool(); operator bool() const; };
struct C { explicit operator bool() { return true; } operator bool() const { return true; } };
C c;
protocol_view<I> pv(c);
bool d = pv;
```

`bool d = pv;` fails with "explicit conversion function was not considered" (conversion_thunks.hh:55), while `bool a = c;` and `bool b = protocol<I>(C{});` compile. The cause is the `all_const` dedup in protocol_wrappers.hh (lines 66-73): `same_name_and_parameters` treats the two conversions as one and drops the implicit const overload without comparing `is_explicit`. The new rule at conformance.hh:133 does not touch that path, so the resolved thread is not fixed. Either compare `is_explicit` in the dedup or reject such an interface, and add a test for the pair. #426 frames this as a conformance question; the fault is in the wrapper dedup.

**Stale comments.** `protocol_interface_function_infos` (conformance.hh:209, raised last time) and `conformance_candidate_infos` (conformance.hh:171) still say "member functions and call operators"; only `named_member_function_infos` was updated. protocol_wrappers.hh:113 lists only `member_base_generator` and `X_operator_overload_set` as sources of bases, and lines 120-122 describe `operator_overload_set` as "for each call operator".

Items you deferred from the first review (interface conversion templates, double mangling of the target type, the fourth `fn_ptr_type()` copy, the if-chain in `generate_member_bases_wrapper`) are not repeated. Probes for lambdas, `operator auto()` candidates, function-pointer/reference/`void`/`std::vector<int>` targets, explicit-object candidates, noexcept propagation and const/non-const selection all pass on head.

conversion_thunks.hh
```diff
@@ -0,0 +1,129 @@
...
+    requires(!IsConst)
+  {
+    auto* protocol_object = static_cast<ProtocolType*>(this);
+    return call_through_vtable<Member, Vtable>(
+        protocol_object, protocol_object->vtable_, protocol_object->object_);
+  }
+
+  explicit(IsExplicit) operator R() const noexcept(IsNoexcept)
```
philipcraig, line 55: Forwarding `explicit operator bool` makes `if (p)` compile on an owning `protocol<I>`, where it reads like an emptiness check but calls through the vtable of a moved-from object. #427 tracks the design. Until that is decided, could this PR reject a conversion to `bool` on owning protocols in `protocol_interface_function_infos`, or at least say in the class comment (protocol.hh:121) that `valueless_after_move` is the only valid query?

protocol_wrappers.hh
```diff
@@ -152,6 +155,10 @@ consteval std::meta::info generate_member_bases_wrapper() {
...
```
philipcraig, line 154: These consteval branches (158, 160, 172 here; conversion_thunks.hh:82, 86) are never evaluated by the consteval-coverage job: every test that builds a protocol or view with a conversion function is in protocol_operator_tests.cc, which is not in `DEFAULT_TRANSLATION_UNITS` in scripts/consteval_coverage.py, and the protocol_test.cc additions are conformance-only `static_assert`s. Add the operator test file to the list, or one protocol_test.cc test that builds a protocol with a conversion.

protocol_operator_tests.cc
```diff
@@ -408,6 +410,174 @@ TEST(ReflectionProtocolTest, OperatorLogicalNot) {
...
+
+TEST(ReflectionProtocolTest, ConversionTargetMatchesThroughAlias) {
+  using S = std::string;
+
+  struct Interface {
+    explicit operator S() const noexcept;
+  };
+
+  struct Conforming {
+    explicit operator std::string() const noexcept { return "hello"; }
```
philipcraig: clang-tidy (required) fails here: `bugprone-exception-escape`, since constructing the `std::string` can throw inside a `noexcept` function. Dropping `noexcept` from this test's `Interface` and `Conforming` clears it.

protocol_operator_tests.cc
```diff
@@ -408,6 +410,174 @@ TEST(ReflectionProtocolTest, OperatorLogicalNot) {
...
+  };
+
+  protocol<Interface> p(Conforming{});
+  EXPECT_TRUE(static_cast<bool>(p));
+  EXPECT_EQ(static_cast<int>(p), 42);
+}
+
+TEST(ReflectionProtocolTest, NonExplicitConversionIsImplicit) {
+  struct Interface {
+    operator int() const noexcept;
```
philipcraig, line 461: `ReflectionProtocolTest.NonExplicitConversionIsImplicit` and the `ReflectionProtocolViewTest` one declare identical `operator int() const noexcept` pairs as two distinct local types, and `ConstOnlyConversionFunction` repeats the shape of `ConversionAlongsideNamedMember` and `NonNoexceptConversion`. Each pair re-instantiates the vtable and wrappers (#423). Sharing one pair between the two implicit-conversion tests and folding `ConstOnlyConversionFunction` into `ConversionToBool` would trim that.

protocol_operator_tests.cc
jbcoe: Fixed.

conversion_thunks.hh
jbcoe: I'll defer all handling of bool conversion in #427

protocol_wrappers.hh
jbcoe: Added  `protocol_operator_tests.cc` and `vtable_tests.cc` and resorted by compile time:

 | File | Wall time |
  |---|---|
  | vtable_tests.cc | 9.9s |
  | name_mangling_tests.cc | 13.4s |
  | allocator_tests.cc | 26.8s |
  | forwarding_test.cc | 31.8s |
  | protocol_operator_tests.cc | 39.3s |
  | protocol_test.cc | 59.9s |

protocol_operator_tests.cc
jbcoe: Tests follow an existing (anti-)pattern. I'll defer the cleanup and consequent compilation improvements to a follow-up PR.

conformance.hh
jbcoe: The conservative approach (exact match is required) is a placeholder until #426 is addressed.

jbcoe:
Explicitness needs follow up design work #426.

I'm happy for this to be merged as-is so that we can make progress.

protocol_wrappers.hh
```diff
@@ -109,15 +110,17 @@ using member_base_t =
...
+// `conversion_overload_set` into one type via multiple inheritance.
 template <typename... MemberBases>
 struct member_bases_wrapper : MemberBases... {};
 
 // Returns a `member_bases_wrapper` specialisation with one base per public,
 // non-special, member function name of `interface_type`, giving named members
-// with an `operator()` for each overload selected by `ConstPolicy`, plus a
-// `operator_overload_set` if `interface_type` has call operators.
+// with an `operator()` for each overload selected by `ConstPolicy`, plus an
+// `operator_overload_set` for each call operator and a
```
philipcraig: Still reads "an `operator_overload_set` for each call operator": it is produced for every supported operator. `conversion_overload_set` is one per target type (const and non-const overloads of a target share a set), not one per conversion function.

conformance.hh
```diff
@@ -197,10 +206,11 @@ consteval bool is_unsupported_operator(std::meta::operators op) {
          op == op_greater || op == op_greater_equals || op == op_spaceship;
 }
 
-// The named, non-static, non-special member functions and call operators of
-// `Type`, static or not, in declaration order. Ref-qualified functions,
-// explicit-object member functions and the operators
-// `is_unsupported_operator` names are unsupported on protocol interfaces.
+// The named, non-static, non-special member functions, operators and
+// conversion functions of `Type`, static or not, in declaration order.
```
philipcraig: "non-static ... of `Type`, static or not" contradicts itself; this function skips static members (line 218), so "static or not" belongs only on `conformance_candidate_infos`.

protocol_operator_tests.cc
```diff
@@ -408,6 +410,174 @@ TEST(ReflectionProtocolTest, OperatorLogicalNot) {
...
+    explicit operator bool() const noexcept { return false; }
+
+    explicit operator bool() noexcept { return true; }
+  };
+
+  static_assert(std::is_constructible_v<bool, protocol<Interface>>);
+  static_assert(!std::is_convertible_v<protocol<Interface>, bool>);
+
+  protocol<Interface> p(Conforming{});
+  EXPECT_TRUE(static_cast<bool>(p));
```
philipcraig, line 429: The const overload returns `false` but is only reached through `protocol_view<const Interface>`. A `const protocol<Interface>` conversion expecting `false` would pin const/non-const selection on an owning protocol that carries both thunks; no test does today.

philipcraig (approved):
Third pass, at 7ed9e3a. The clang-tidy and consteval-coverage points are fixed, and the explicitness and `operator bool` points are tracked in #426 and #427. Nothing here needs to hold the merge; two behaviours worth recording for the follow-up work:

**`protocol_view<I>` can pick a different conversion than the object.** Not an `explicit` question, so outside #426:

```cpp
struct I { operator int(); operator long() const; };
struct C { operator int() { return 1; } operator long() const { return 2; } };
C c;
long a = c;                     // 1
long b = protocol<I>(C{});      // 1
long d = protocol_view<I>(c);   // 2
```

Under `all_const` both thunks are const-qualified, so the object-argument tie-break that prefers the non-const `operator int` is gone and the exact-match `operator long` wins. The dedup in protocol_wrappers.hh (lines 62-73) only compares conversions to the same target, so its "const and non-const access path must yield the same result" comment does not hold for conversions.

**`operator auto() const;` on an interface is silently ignored.** `members_of` does not return a function with an undeduced return type, so it never reaches the `is_conversion_function` filter: `is_protocol_conformant<I, Empty>()` is true for an empty struct, `protocol<I> p(C{})` compiles, and `static_cast<int>(p)` fails with "invalid static_cast". This is separate from the interface conversion-template case you deferred. Noted on #402.

protocol_wrappers.hh
jbcoe: I've rewritten and simplified this comment. It's obtuse and maintaining it is busy work. The function itself needs rewriting.

protocol_operator_tests.cc
jbcoe: Well spotted. Fixed.

jbcoe:
All cleanup issues should be fixed now. Some significant design work to do next #426 #427

PR #422, opened 2026-09-18
Remove the required-checks explanation from CONTRIBUTING

philipcraig:
CONTRIBUTING says a per-configuration leg cannot be required because a leg skipped by change detection reports under its unexpanded matrix name. That stopped being true in #293, and the `main` ruleset now requires five legs.

Rather than correct the explanation, this deletes it: the ruleset shows what is required, and a written copy has to be kept in step with it.


CONTRIBUTING.md
```diff
@@ -72,12 +72,14 @@ fails on stock GCC.
 Pull requests run the workflows in `.github/workflows`. Each build workflow
 ends in a job with a fixed name (`bazel`, `cmake`, `sanitizers`) that reports
 the outcome of its matrix; `uv-lock` and `pre-commit` are single jobs. The
-checks the `main` ruleset requires are drawn from these fixed-name jobs; the
-current list is in the ruleset, not here. The per-configuration legs
-(`GCC trunk Release`, `asan`, and so on) and the macOS jobs (`GCC-16 (macOS)`
-in Bazel, `GCC-16 (macOS) Release` in CMake) are informational: a leg skipped
-by change detection reports under its unexpanded matrix name, so it cannot be
-required.
+checks the `main` ruleset requires are listed in the ruleset, not here.
```
jbcoe: I'm happy to merge or delete this.

I'm not sure how helpful writing this somewhat convoluted logic down is. Folk can see required checks and needing to sync commits and config changes feels like a likely source of inconsistency and busy-work.

jbcoe (approved):

CONTRIBUTING.md
philipcraig: > I'm happy to merge or delete this.
> 
> I'm not sure how helpful writing this somewhat convoluted logic down is. Folk can see required checks and needing to sync commits and config changes feels like a likely source of inconsistency and busy-work.

Agree, I'll revise to delete it

Issue #423, opened 2026-09-18
Reduce test compile time on the critical path and per interface member


philipcraig:
`protocol_test.cc` takes 50.5 s to compile at `-O2` with GCC trunk 20260906 (`/opt/gcc-latest`, as in CI), against 4.8 s for gtest plus `protocol.hh` alone. Timings are from a 2-core GitHub Codespaces VM. It is the longest TU in every job that builds the tests, so it sets the floor for a parallel build.

| TU | `-O2 -c` |
|---|---|
| protocol_test.cc | 50.5 s |
| protocol_operator_tests.cc | 26.4 s |
| forwarding_test.cc | 23.4 s |
| allocator_tests.cc | 12.6 s |
| name_mangling_tests.cc | 6.3 s |
| consteval_check_test.cc | 5.7 s |
| vtable_tests.cc | 4.9 s |

Two levers, from the measurements in #404:

1. Split `protocol_test.cc` (138 tests, 2769 lines) into a few TUs by topic so they compile in parallel. Each TU pays the 4.8 s baseline again, so three or four pieces rather than many.
2. Reduce the per-member cost in the header. A const/non-const operator pair through `protocol`, `protocol_view` and `protocol_view<const>` costs about 0.26 s; the interface itself costs about 0.17 s. Users pay this as well as the tests.

`-ftime-report` for `protocol_test.cc` (rows overlap):

| Timer | Time |
|---|---|
| overload resolution | 15.9 s (31%) |
| template instantiation | 12.4 s (25%) |
| callgraph functions expansion | 10.8 s (21%) |
| constant expression evaluation | 7.7 s (15%) |
| garbage collection | 6.4 s (13%) |
| constraint satisfaction | 6.2 s (12%) |

Overload resolution and constraint satisfaction on the generated wrapper members are the first place to look. The snapshot is built with extra diagnostic checks, which inflates every figure; garbage collection and the tree verifier are 18% of the total here.

philipcraig:
One instance of the per-member cost, deferred from the #421 review (https://github.com/jbcoe/cc-protocol/pull/421#discussion_r4048813238). The conversion tests in protocol_operator_tests.cc declare the same shapes as distinct local types, and each pair instantiates its own vtable and wrappers:

[...]

PR #425, opened 2026-09-18
Move closures over interface members out of templates

philipcraig:
Part of #423.

A lambda in a function template is a distinct closure type in each specialisation, so every `std::ranges` algorithm or view adaptor it was passed to was instantiated, and had its constraints checked, once per interface or candidate type. This moves three such loops into non-template `consteval` functions that take reflections as ordinary arguments:

- `named_member_function_infos`, the `views::filter` pipeline from `conformance_candidate_infos`
- `all_members_conform`, the `all_of`/`any_of` from `is_protocol_conformant`
- `contains_same_name`, the `any_of` from `generate_member_bases_wrapper`

Per-TU compile time, `-O2 -c`, GCC trunk 20260906 on a 2-core Codespaces VM:

| TU | main | this PR |
|---|---|---|
| protocol_test.cc | 50.0 s | 34.2 s |
| protocol_operator_tests.cc | 24.5 s | 19.8 s |
| forwarding_test.cc | 23.5 s | 20.4 s |
| allocator_tests.cc | 12.5 s | 12.4 s |

For 40 one-member interfaces used through `protocol`, the cost above an empty TU falls from 6.1 s to 3.6 s at `-fsyntax-only`.

CI, one run each, main (f01e297, scheduled) against this PR. The CMake figures are the ninja build span and, in brackets, the time until the last test object finished, both from log timestamps; these jobs have no build cache. The Bazel and sanitizer jobs restore a disk cache, so they are not compared.

| Job | main | this PR |
|---|---|---|
| CMake, GCC trunk Release | 73 s (61 s) | 103 s (66 s) |
| CMake, GCC trunk Debug | 51 s (51 s) | 36 s (35 s) |
| CMake, GCC-16 Release | 57 s (48 s) | 32 s (24 s) |
| CMake, GCC-16 Debug | 28 s (27 s) | 24 s (23 s) |
| CMake, GCC-16 macOS Release | 35 s (30 s) | 32 s (23 s) |
| clang-tidy, build step | 322 s | 113 s |
| Runtime coverage, build and test step | 86 s | 55 s |
| Consteval coverage, probe step | 205 s | 243 s |

The GCC trunk Release job landed on a slow runner: `allocator_tests.cc`, which this change does not speed up or slow down locally, finished 9 s into the build against 3 s on main. Single runs vary by that much, so the table shows direction rather than size.


jbcoe (approved):
Thanks for this.

PR #433, opened 2026-09-18
Mangle a conversion function's target type once

### LINKED ISSUE #429: Conversion functions mangle their target type twice
`base_name_of` (name_mangling.hh:251) names a conversion function `cv<type>`, and `mangle()` then appends the return type as it does for every function. For a conversion function the two are the same type, so `operator bool() const` mangles to `fn_NKcvbEb`. For `operator std::string() const` the vtable entry name is 152 characters and carries the `basic_string<char, char_traits, allocator>` mangling twice.

Either half is enough to keep conversions distinct: `cv` alone once the return type follows, or `cv<type>` without the return type, which is what Itanium does. In the #421 review @jbcoe suggested dropping the return type from every mangled name, as Itanium does (https://github.com/jbcoe/cc-protocol/pull/421#discussion_r4047984691). An interface cannot declare two members that differ only in return type, so nothing is lost. `name_mangling_tests.cc` pins the current strings.

A smaller point in the same code: a conversion to an unsupported target, such as a pointer to member (#259), fails with "name mangling: unsupported parameter type" for a function that has no parameters. "unsupported type" covers both uses.

philipcraig:
Closes #429.

`mangle()` no longer appends the return type for a conversion function, whose `cv<type>` atom already carries it, as in Itanium. `operator bool() const` becomes `fn_NKcvbE` (was `fn_NKcvbEb`), and a `std::string` target is mangled once.

Other functions keep their return type. The header documents it as what keeps `get() const` on two interfaces apart when they differ only in return type, and `DistinguishesMembersDifferingOnlyByReturnType` pins that, so dropping it everywhere is a separate decision.

The "unsupported parameter type" error becomes "unsupported type", since it also fires for a conversion target.


jbcoe (approved):
Thanks for this.

Issue #393, opened 2026-09-15
Consider function templates as conformance candidates


jbcoe:
From [Dooez](https://github.com/Dooez)
[Timofey Dovgal](https://github.com/Dooez)

If you decide on explicit signature requirement as part of conformance rule, then I would suggest allowing templates in case an exact signature match can be instantiated.
This can be checked and used via member function pointer conversion from template. gcc 16.1 ICEs, but gcc trunk works
Ret (T::* ptr)(Args...) = &template [:member_function_template_reflection:];
https://godbolt.org/z/qWMr75cbb

PR #434, opened 2026-09-18
Reject member function templates on protocol interfaces

### LINKED ISSUE #428: Member templates on an interface are silently dropped
`named_member_function_infos` (conformance.hh:157) filters `members_of` with `is_function`, and a member function template or conversion function template is not a function. On an interface it is therefore neither forwarded nor rejected:

```cpp
struct I {
  int f() const;
  template <typename T>
  explicit operator T() const;
  template <typename T>
  void g(T) const;
};

static_assert(is_protocol_conformant<I, C>());  // passes
protocol<I> p(C{});                             // builds
p.g(1);                // error: 'class protocol<I>' has no member named 'g'
static_cast<bool>(p);  // error: invalid static_cast
```

An interface author can put a template in the contract, have a type without it accepted, and find out at the call site. This is the shape #402 described for conversion functions, and the comment there on `operator auto() const` is a third case of it. After #388 every non-forwardable operator throws from `protocol_interface_function_infos` with a message naming it; templates are outside that rule.

A template cannot be forwarded through a vtable, so the fix is to reject: walk `members_of` for the interface and throw when `is_function_template` or `is_conversion_function_template` holds, naming the member. `operator auto()` needs a different test, since `members_of` does not return it.

Candidate-side templates are a separate question (#393); `TemplatedConversionFunctionsAreIgnored` pins that side only. #407 lists what an interface silently ignores and would gain or lose this entry. Raised in the #421 review (https://github.com/jbcoe/cc-protocol/pull/421#discussion_r4044611399).

philipcraig:
Closes #428.

`named_member_function_infos` filters `members_of` with `is_function`, which a template is not, so a member function template, operator template or conversion function template on an interface was neither forwarded nor rejected. A type without the member conformed, and the omission surfaced at the call site.

`protocol_interface_function_infos` now walks the interface's members first and throws for any non-static, non-constructor function template:

```
member function template 'g' is not supported in a protocol interface
```

Constructor templates and static member function templates stay ignored, as constructors and static member functions already are.

`operator auto()` is not covered: `members_of` does not return a function whose return type is still undeduced, so reflection cannot see it. That case stays with #402. Candidate-side templates (#393) are unchanged.


jbcoe (approved):
Thanks for this.

PR #435, opened 2026-09-19
Share duplicated interfaces across the conversion tests

philipcraig:
Part of #423.

The two `NonExplicitConversionIsImplicit` tests declared identical interface and conforming types; they now share one pair at namespace scope. `ConstOnlyConversionFunction` is removed: a const-only conversion through `protocol` and both views is covered by `ConversionToBool`, `ConversionAlongsideNamedMember` and the `NonExplicitConversionIsImplicit` tests.

`protocol_operator_tests.cc`, `-O2 -c`, GCC trunk, user+sys CPU over alternating runs: 10.6 s before, 10.3 s after.


jbcoe (approved):
Thanks for this. Tests could probably do with a broader audit/refactor but this is a local improvement.

PR #436, opened 2026-09-19
Name a rejected interface member the same way in every diagnostic

### LINKED ISSUE #406: Duplicated name computation and two naming schemes in `protocol_interface_function_infos`
In conformance.hh:196-228 (after #388):

- the `has_identifier(member) ? identifier_of(member) : display_string_of(member)` expression is written out twice (lines 202 and 210);
- `is_operator_function(member)` is tested in two consecutive `if`s and `operator_of(member)` is called three times;
- `operator_diagnostic_name` (line 176) hand-rolls a lowercase-first-character test over `symbol_of` to decide on a space, while the neighbouring ref-qualifier and explicit-object diagnostics use the identifier/display-string idiom, which already renders an operator correctly and names the exact overload.

So an interface with two rejected `operator==` overloads gets a message naming neither, while the adjacent diagnostic pinpoints its member, and a member that is both ref-qualified and a blocklisted operator prints its name two ways depending on which check fires first. The next rejection rule would copy the name computation a third time.

Hoisting `std::string name = ...` once and nesting the operator checks under a single `if (is_operator_function(member)) { auto op = operator_of(member); ... }` removes all of it. `parameters_of(member)` is also materialised for every member just to test `empty()`, on a path that runs for every (interface, candidate) pair.

philipcraig:
Closes #406.

`reject_interface_member(kind, member)` builds the diagnostic name and throws.
It replaces the three copies of the identifier-or-display-string expression
(`reject_member_function_templates` had added a third) and
`operator_diagnostic_name`, whose `symbol_of` text named no overload. The name
is computed only when a member is rejected.

Every rejection now names its member the same way. For an operator or a
conversion function that is the display string, which identifies the overload:

```
operator 'bool Interface::operator==(const FirstOperand&) const' is not supported in a protocol interface
ref-qualified member function 'bool Interface::operator==(const FirstOperand&) const &' is not supported in a protocol interface
address-of operator 'int* Interface::operator&()' is not supported in a protocol interface
operator 'int Interface::operator co_await()' is not supported in a protocol interface
```

The operator checks nest under one `is_operator_function` test with a single
`operator_of` call.

`parameters_of(member)` stays: since the issue was filed the explicit-object
check has come to need the parameter list for every member, and reflection
has no query for an explicit-object member function that avoids it.

The new tests check the message text, which no test did before. A display
string is implementation-defined, so they look for substrings.


jbcoe (approved):
Thanks for this. A big readability win.

PR #437, opened 2026-09-19
Build every function-pointer type with one function

### LINKED ISSUE #405: Three copies of the function-pointer-type builder

philipcraig:
Closes #405.

`function_pointer_type_of(member, leading = {})` in `overload_spec.hh` replaces the builders in `operator_thunk_for`, `member_function_thunk_for` and `generate_vtable_specs`. A vtable entry differs from its thunk only in the leading `void*` or `const void*`, which `generate_vtable_specs` passes as `leading`, so the two can no longer drift apart.

`conversion_thunk` no longer goes through a function-pointer type. It reads its target type, `noexcept` and `explicit` from `Member`, so its builder, the `static_assert(false)` stub, the partial specialisation and the `IsExplicit` parameter go. The target type is a namespace-scope alias, `conversion_target_t<Member>`, not a member alias, because the thunk is a base class of the protocol and a member alias would be visible on it.

`call_through_vtable` takes `const Vtable*` in place of a deduced `VtablePtr*`. `Vtable` stays an explicit template argument: `protocol`'s owning vtable derives from `vtable_t<I>` and converts at the call, so the type searched for the entry is the type the entry is spliced into.

Covers the same ground as #413, plus the conversion thunk added by #421 and the `call_through_vtable` item.


jbcoe:
This is awesome, a lot of complicated, repetitive code cleaned up.

jbcoe (approved):
Thanks for this!

PR #438, opened 2026-09-19
Reject view interfaces whose conversion functions differ in constness

### LINKED ISSUE #430: `protocol_view<I>` can select a different conversion function than the viewed object
```cpp
struct I { operator int(); operator long() const; };
struct C {
  operator int() { return 1; }
  operator long() const { return 2; }
};

C c;
long a = c;                    // 1: operator int, then int -> long
long b = protocol<I>(C{});     // 1
long d = protocol_view<I>(c);  // 2: operator long
```

On a non-const `C`, overload resolution prefers `operator int` because its implicit object parameter binds without adding `const`. Under `const_policy::all_const` every thunk on the view is const-qualified, so that tie-break is gone and the exact-match `operator long` wins.

The dedup in `is_forwarded_member_function` (protocol_wrappers.hh:62-73) exists so that "overload resolution through a const and non-const access path must yield the same result". It compares name and parameters, and the name of a conversion function is its target type, so conversions to different targets are never compared. Named members do not have this problem, because members with different names do not compete in overload resolution; conversion functions to different targets do.

This is not an `explicit` question, so it is outside #426. Options:

1. On a view, reject an interface whose conversion functions differ in constness across targets.
2. Give `protocol_view<I>` non-const thunks for non-const conversions, so the tie-break survives; a `const protocol_view<I>` then loses those conversions.
3. Document it.

Raised in the third round of the #421 review.

philipcraig:
Closes #430.

Option 1 from the issue. Generating the wrappers for `protocol_view<I>` throws if the conversion functions it forwards differ in constness. A target with both a const and a non-const conversion function counts as non-const, since the view forwards only the non-const one. `explicit` does not exempt a conversion function: explicit conversions compete in direct-initialisation and contextual conversion.

`protocol<I>` and `protocol_view<const I>` select the same conversion function as the object and still accept these interfaces. For that to hold, the `protocol_view<const I>` constructor from `protocol_view<I>` becomes a template constrained with `std::same_as`. Taken by value as a non-template, it completed `protocol_view<I>` during overload resolution for any argument, so a const view of such an interface would have been rejected too.


jbcoe:
I'm not sure if this is the approach that we eventually want to take but it's safe, clear, principled and a definite improvement over what we currently have.

Happy to merge.

jbcoe (approved):
Thanks for this.

PR #439, opened 2026-09-19
Forward comparison operators on protocol interfaces

### LINKED ISSUE #401: Member comparison operators on an interface are now a hard error
#388 lists `==`, `!=`, `<`, `<=`, `>`, `>=` and `<=>` in `is_unsupported_operator`, so `protocol_interface_function_infos` throws for any interface that declares one as a member. On `main` the same members were silently ignored, which is what DRAFT.md ("Equality and comparison operators") still describes:

```cpp
struct I { int f() const; bool operator==(const I&) const; };
protocol<I> p(C{});
// main: compiles, operator== ignored
// #388: conformance.hh:233: uncaught exception ... 'operator== is not supported in a protocol interface'
```

The same happens for `bool operator==(const I&) const = default;` and a defaulted member `operator<=>`; `is_special_member_function` does not cover defaulted comparisons. Hidden-friend comparisons are unaffected.

Options:

1. Skip comparisons the way `main` did, keeping DRAFT.md true.
2. Keep the rejection, update DRAFT.md, and add rejection tests. `ConformsToTest.UnsupportedOperatorsAreRejected` covers only `operator=` and `operator co_await` today.
3. Forward them, which is the follow-up discussed on #388.

Whichever way, the `= default` idiom needs a decision of its own since it is the common way to write a comparison on a struct.

philipcraig:
Closes #401.

Takes option 3 from the issue: a member `==`, `!=`, `<`, `<=`, `>`, `>=` or
`<=>` on an interface is forwarded like any other binary operator, so a
conforming type must declare a member with the same signature.

```cpp
struct I {
  bool operator==(std::string_view) const;
  bool operator==(std::default_sentinel_t) const;
};
protocol<I> p(C{"id"});
p == "id"; "id" == p; p != "other";  // rewritten candidates work
p != std::default_sentinel;
```

A comparison with an operand of type `I`, defaulted or not, is rejected with a
diagnostic naming the operator. Forwarding it would ask every candidate for
`operator==(const I&) const`, which is not what the author of the interface
meant, and ignoring it would silently drop a declared member. Rejection can
be relaxed later without breaking anyone; the reverse is not true. #440 tracks
giving it a meaning.

Hidden-friend comparisons on an interface are still ignored, and nothing
compares two `protocol` objects. DRAFT.md "Equality and comparison
operators" is updated to match.


jbcoe (commented):
Thanks for this.

I'm not sure how useful support for `operator <=>(int)` is.

Presumably we'd want to support `operator <=>(const Interface)` etc. somehow (I don't know how) to forward idiomatic operators.

Nothing here is wrong or dangerous so I'm happy to merge, but the 'obvious' use case won't work.

jbcoe (approved):

conformance.hh
```diff
@@ -201,9 +201,7 @@ consteval std::string operator_diagnostic_name(std::meta::operators op) {
 // Some operators are excluded from protocol interfaces.
 consteval bool is_unsupported_operator(std::meta::operators op) {
   using enum std::meta::operators;
-  return op == op_equals || op == op_co_await || op == op_equals_equals ||
-         op == op_exclamation_equals || op == op_less || op == op_less_equals ||
-         op == op_greater || op == op_greater_equals || op == op_spaceship;
+  return op == op_equals || op == op_co_await;
```
jbcoe, line 208: raising a comment just to prevent auto-merge so that you can consider the review comment.

philipcraig:
> Thanks for this.
> 
> I'm not sure how useful support for `operator <=>(int)` is.
> 
> Presumably we'd want to support `operator <=>(const Interface)` etc. somehow (I don't know how) to forward idiomatic operators.
> 
> Nothing here is wrong or dangerous so I'm happy to merge, but the 'obvious' use case won't work.

I won't merge this yet. Needs more work to show if it can be made useful

philipcraig:
Agreed, `p == q` is the case people will reach for, and this PR doesn't give it to them. It needs a way to say "my own type" in the interface, a type identity in the vtable, and a policy for mismatched types. I've opened #440 for that design.

Two changes here in response:

- A comparison with an operand of type `I` is now rejected whether or not it is defaulted. Before, `bool operator==(const I&) const;` was forwarded literally, so a conforming type had to compare itself with `I`. Rejecting it keeps that spelling free to mean comparison between two protocols later.
- The examples and DRAFT.md now use `operator==(std::string_view)` and `operator==(std::default_sentinel_t)` instead of `int`. Those heterogeneous cases are what this PR is for.

PR #444, opened 2026-09-20
Tell contributors when the sandbox image is stale

### LINKED ISSUE #441: Tell contributors when the sandbox image is stale
Follow-up to #392, which keeps rebuild timing out of scope. Once the caches live in the image, a stale image still works but quietly falls back to the network, and nothing says so.

**Stale image is reused silently.** `scripts/agentic-sandbox.py` rebuilds only when the image is missing or `--rebuild-docker` is passed. An image built before #392 has the same name, so it is reused with the volumes gone and empty cache directories. The same happens after any later bump to `uv.lock` or `MODULE.bazel`. Suggestion: label the image at build time with a hash of its inputs, and have the script print a one-line "image is stale, pass `--rebuild-docker`" hint when the hash differs. The inputs are `docker/Dockerfile`, `.bazelversion`, `MODULE.bazel`, `MODULE.bazel.lock`, `pyproject.toml`, `uv.lock` and the googletest tag. The script would print a hint only and would not rebuild automatically.

**CONTRIBUTING wording.** "Any repository change needs `--rebuild-docker`" is broader than the truth. Source edits never need a rebuild, and a dependency bump without one degrades to an online fetch rather than breaking. Suggestion: list the trigger files above, state that fallback, and fold in the separate `.bazelversion` rule near the top of the file so the guidance lives in one place.

**googletest tag.** The tag is pinned in `CMakeLists.txt`, `MODULE.bazel` and `docker/Dockerfile`. With `FETCHCONTENT_SOURCE_DIR_GOOGLETEST` set, CMake never checks `GIT_TAG`, so a bump in `CMakeLists.txt` leaves sandbox CMake builds on the old googletest until someone rebuilds. This is the one case with no network fallback. The Dockerfile can read the tag rather than repeat it:

```dockerfile
RUN --mount=type=bind,source=CMakeLists.txt,target=CMakeLists.txt \
    git clone --depth 1 --branch "$(sed -n 's/^ *GIT_TAG *//p' CMakeLists.txt)" \
        https://github.com/google/googletest.git "$FETCHCONTENT_SOURCE_DIR_GOOGLETEST"
```

**Layer order.** The `bazel fetch` layer is the one most likely to miss the layer cache, and it precedes `uv sync` and the googletest clone. Every fetch miss therefore re-downloads the wheels and re-clones googletest. The three steps write to disjoint trees, so the fetch can be the last `RUN` of the stage.

Blocked on #392 landing. Happy to write the PR.

### LINKED ISSUE #445: Harden the sandbox image's googletest tag extraction and `.dockerignore`
Non-blocking findings from the review of #392, left for a follow-up so that PR can land.

- [ ] **`docker/Dockerfile`: take only the tag from the `GIT_TAG` line.** `sed -n 's/^ *GIT_TAG *//p'` passes the whole rest of the line to `git clone --branch`. A CRLF checkout (there is no `.gitattributes`), a trailing `# comment` or a second `GIT_TAG` line makes the clone fail with "Remote branch not found". This takes only the tag:
  ```
  sed -n 's/^ *GIT_TAG *\([^ #\r]*\).*/\1/p' CMakeLists.txt | head -n 1
  ```
- [ ] **`.dockerignore`: add `.claude`.** `source=.` keys the `bazel fetch` layer on everything not ignored. `.bazelignore` already lists `.claude`, and agent worktrees under it change constantly, so the layer rarely hits its cache.
- [ ] **`.dockerignore`: use `**/__pycache__`.** Patterns match from the context root, unlike `.gitignore`, so `__pycache__` misses `scripts/__pycache__`.
- [ ] **`docker/Dockerfile`: correct the `bazel fetch` comment.** It calls the fetch layer "the only input here a source edit invalidates", but the googletest clone is keyed on the whole of `CMakeLists.txt`, so any CMake edit re-clones googletest and re-runs the fetch.
- [ ] **`docker/Dockerfile`: mention the output base.** The fetch layer also captures Bazel's output base for `/workspace` under `~/.cache/bazel/_bazel_vscode/`, which the runtime container reuses because the checkout mounts at the same path. It is the part that grows on every rebuild.
- [ ] #441, which tracks the rebuild-timing points taken out of the #392 review: the stale-image hint in `scripts/agentic-sandbox.py`, the CONTRIBUTING trigger list with the `.bazelversion` rule folded in, and the googletest tag reminder. #392 already reads the tag from `CMakeLists.txt` and puts the fetch layer last.

philipcraig:
Closes #441. Closes #445.

- `scripts/agentic-sandbox.py` labels the image with a hash of its inputs and prints a one-line hint to pass `--rebuild-docker` when the hash no longer matches. It never rebuilds on its own. An image with no label, such as one built before #392, counts as stale.
- The inputs are `docker/Dockerfile`, `.bazelversion`, `MODULE.bazel`, `MODULE.bazel.lock`, `pyproject.toml`, `uv.lock` and the googletest `GIT_TAG` in `CMakeLists.txt`. Only the tag is hashed from `CMakeLists.txt`, so adding a test does not raise the hint.
- The `bazel fetch` layer mounts only the files that define the graph, which need no sources. That removes `.dockerignore` and the whole-checkout mount, and an ordinary source edit no longer invalidates the layer. `BUILD.bazel` and `.bazelrc` are mounted but not hashed: a new dependency has to change `MODULE.bazel` as well.
- `CONTRIBUTING.md` lists the inputs once, states the network fallback and the googletest exception, and the `.bazelversion` rule near the top now points there.
- The Docker workflow drops `.dockerignore` from its `paths` filters.

From #445:

- The googletest clone takes only the tag from the first `GIT_TAG` line, so a CRLF checkout, a trailing comment or a second `GIT_TAG` line no longer breaks `git clone --branch`. This is the commit from #446.
- The `bazel fetch` comment now says that any `CMakeLists.txt` edit re-clones googletest and re-runs the fetch, and notes the repository cache and output base the layer captures.
- The two `.dockerignore` items are moot, because this PR deletes the file.


CONTRIBUTING.md
```diff
@@ -249,14 +249,27 @@ the rest of the host file system, but the network is unrestricted and the
...
-every container, so it (currently) needs the network every time.
+network. The image is built from:
+
+- `docker/Dockerfile`
+- `.bazelversion`
+- `MODULE.bazel` and `MODULE.bazel.lock`
+- `pyproject.toml` and `uv.lock`
+- the googletest `GIT_TAG` in `CMakeLists.txt`
+
+Source edits never need a rebuild. When one of the inputs changes, the script
```
jbcoe: nit: maybe 'container rebuild' 

I cannot imagine anyone being more than momentarily confused by this though.

Feel free to ignore.

jbcoe (approved):
Thanks for this.

PR #446, opened 2026-09-20
Take only the tag from the googletest GIT_TAG line

philipcraig:
Part of #445.

- `docker/Dockerfile` passed the whole rest of the `GIT_TAG` line to `git clone --branch`. A CRLF checkout, a trailing `# comment` or a second `GIT_TAG` line made the clone fail with "Remote branch not found".
- The `sed` expression now captures the tag alone, accepts tab indentation, ignores a longer keyword such as `GIT_TAGGED`, and stops at the first match, so it needs no pipe to `head`.
- The rest of #445 overlaps #444, which deletes `.dockerignore` and rewrites the `bazel fetch` comment. This change touches neither, so the two merge cleanly in either order.


philipcraig:
Folded into #444 as 7cecd46, a cherry-pick of this PR's commit, so the sandbox image changes land together. #444 now closes #445. Closing this one.

PR #447, opened 2026-09-20
Document that allocation operators on an interface are ignored

### LINKED ISSUE #403: Allocation operators on an interface are silently dropped
`operator new`, `operator delete` and their array forms are implicitly static, so `protocol_interface_function_infos` drops them at `if (is_static_member(member)) continue;` with no diagnostic:

```cpp
struct I { int f() const; void* operator new(std::size_t); void operator delete(void*) noexcept; };
// conforms against any C with int f() const; the allocation functions vanish
```

Meanwhile #388 adds `NamesAllocationOperatorsWithNoQualifiersAsTheyAreStatic` to name_mangling_tests.cc and `nw`/`dl`/`na`/`da` cases to `base_name_of`, none of which is reachable through the library. `operator co_await`, by contrast, gets a clear "is not supported in a protocol interface" throw.

CONTRIBUTING.md's "Static member functions declared on the interface itself are ignored" arguably covers this. If it is intended, the docs should name the allocation operators explicitly and the mangling cases can go; if not, they should be rejected like `co_await`.

philipcraig:
Fixes #403.

`operator new`, `operator delete` and their array forms are implicitly static, so `protocol_interface_function_infos` skips them with every other static member of an interface. This takes the issue's first option and keeps that behaviour, now documented and tested, rather than rejecting them as `operator co_await` is.

The reason is parity with `std::polymorphic`. It allocates with `allocator_traits::allocate` and constructs with `allocator_traits::construct`, so a class-specific allocation function on the stored type is never called, and the allocator is the only way to control allocation. `protocol` already does the same for the object it owns. An interface is never instantiated, so its allocation functions have nothing to allocate, and forwarding them would change `new protocol<I>` rather than the owned object. Rejecting them would add a restriction `polymorphic` does not have and treat them differently from any other ignored static member.

- CONTRIBUTING.md names the allocation operators under the ignored static members and adds a note that allocation goes only through the allocator.
- `conformance.hh` says the same on `protocol_interface_function_infos`.
- `InterfaceAllocationOperatorsAreIgnored` pins the interface half.
- `ClassSpecificAllocationFunctionsAreNotUsed` pins the conforming-type half: `protocol` owns and copies a type whose `operator new` and `operator delete` are deleted.

The issue suggests the `nw`/`dl`/`na`/`da` mangling arms could go. They stay: the build uses `-Wall -Werror`, so the `base_name_of` switch has to name every `std::meta::operators` enumerator, and removing the arms would mean replacing them with a throwing arm. `mangle` already names other forms the library rejects (`operator co_await`, ref-qualifiers). A comment on the arms now says why they are there.


conformance.hh
```diff
@@ -240,6 +240,10 @@ consteval void reject_member_function_templates(std::meta::info type) {
 // templates, defaulted comparison operators, comparisons with `Type` and the
 // operators `is_unsupported_operator` names are unsupported on protocol
 // interfaces.
+// Static member functions are ignored. That includes `operator new`,
```
jbcoe: I think that we do consider static member functions of a candidate when conforming to an interface. 

The `operator new` / `operator delete` restrictions stand.

jbcoe (approved):
Thanks for this. 

One comment needs revision but otherwise LGTM.

conformance.hh
philipcraig: Agreed, the comment was too broad: it is only the interface's static member functions that are ignored, and a candidate's static member function can conform (`member_function_conforms_to`). Reworded in 1e90838 to say which side it means. No change to the `operator new` / `operator delete` handling.

PR #448, opened 2026-09-20
Reject operator auto on a protocol interface

### LINKED ISSUE #402: Conversion functions on an interface are neither forwarded nor rejected

philipcraig:
Closes #402.

#421 forwards conversion functions and #434 rejects conversion function templates, which left `operator auto() const;`: `members_of` does not return a function whose return type is an undeduced placeholder ([meta.reflection.member.queries]/3), so the member was dropped before any check saw it.

`protocol_interface_function_infos` now names `operator auto` and `operator decltype(auto)` directly and rejects either with the usual diagnostic:

```
conversion function with a placeholder return type 'I::operator auto() const' is not supported in a protocol interface
```

This matches `std::polymorphic`, where the equivalent contract is a virtual function and `virtual operator auto()` is ill-formed.

A function like this can only be found by naming it, so other placeholder functions declared without a body (`auto f() const;`, `operator const auto&() const;`) are still lost. `OtherPlaceholderFunctionsWithoutABodyAreLost` records that. One with a body has a deduced return type and is an ordinary member.


conformance_tests.cc
```diff
@@ -499,6 +499,99 @@ TEST(ConformsToTest, InterfaceConstructorAndStaticTemplatesAreIgnored) {
...
+      conformance_rejection_message_contains<AutoInterface, Candidate>(
+          "operator auto"));
+  static_assert(
+      conformance_rejection_message_contains<DecltypeAutoInterface, Candidate>(
+          "operator decltype(auto)"));
+#endif  // __cpp_constexpr_exceptions
+}
+
+TEST(ConformsToTest, PrivatePlaceholderConversionFunctionIsIgnored) {
+  class Interface {
```
jbcoe, line 538: Not in scope, but should we reject an interface with private or protected functions?

jbcoe (approved):
Thanks for this.

conformance_tests.cc
philipcraig: #451

PR #449, opened 2026-09-20
Reject interface member functions that use a name protocol reserves

### LINKED ISSUE #318: Interface member names that collide with protocol's own members are silently hidden
`protocol` declares `swap`, `get_allocator` and `valueless_after_move` as members. The interface's functions are synthesised as members of a base class, so an interface that uses one of these names compiles but the library's function wins on `protocol`, while `protocol_view` still exposes the interface's function. There is no diagnostic.

```c++
struct I { int get_allocator() const; };
struct T { int get_allocator() const { return 42; } };

protocol<I> p{T{}};
protocol_view<I> v{p};

static_assert(!std::is_same_v<decltype(p.get_allocator()), int>);  // library's member
static_assert( std::is_same_v<decltype(v.get_allocator()), int>);  // interface's member
```

Confirmed on GCC trunk at f409dd2.

Options:

1. Reject such interfaces in `is_valid_interface` with a clear message.
2. Move the library's own operations out of the member namespace where the standard allows it (`swap` already has a hidden-friend form).
3. Document the reserved names in DRAFT.md.

1 and 3 together seem right; 2 is not available for `get_allocator` and `valueless_after_move`, which the standard spells as members on `polymorphic` and the containers.

This is also the reason #314 should not add a member named `type`.

philipcraig:
Closes #318.

`protocol` declares `allocator_type`, `get_allocator` and `swap`. Interface member functions are forwarded by members of a base class, so an interface that used one of these names compiled, the library's member won on `protocol`, and `protocol_view` still forwarded the interface's function.

- `protocol_interface_function_infos` rejects a non-static interface member function with a reserved name: `reserved member name 'swap' is not supported in a protocol interface`. Static member functions are ignored, as before.
- A test compares `reserved_member_names` with the public members of `protocol` and of both views, so a member added later has to be listed or made a hidden friend.
- DRAFT.md gains a "Reserved member names" subsection and lists `allocator_type`, `get_allocator` and `swap` in the `protocol` synopsis, which omitted them.

`valueless_after_move` from the issue is no longer reserved: it became a hidden friend in #395. `target_type` and `protocol_cast` are hidden friends too. No open PR adds a member to `protocol` or `protocol_view`.

Not covered here: the private members of `protocol` (`create`, `vtable`, `traits`, `object_`, …) and the injected-class-names of the generated bases (`type`, `member_bases_wrapper`) hide interface members as well, with an access or ambiguity error at the call rather than silently. Those are implementation names and are better renamed than reserved.


conformance_tests.cc
```diff
@@ -524,6 +527,93 @@ struct InterfaceWithNamedMemberFunctionTemplate {
...
+  static_assert(conformance_check_rejects<AllocatorTypeInterface, Candidate>());
+  static_assert(
+      conformance_rejection_message_contains<GetAllocatorInterface, Candidate>(
+          "reserved member name 'get_allocator'"));
+#endif  // __cpp_constexpr_exceptions
+}
+
+TEST(ConformsToTest, StaticInterfaceMemberWithReservedNameIsIgnored) {
+  struct Interface {
+    static void swap(Interface&, Interface&) noexcept;
```
jbcoe, line 656: Perhaps we should reject this too for safety's sake?

It's possibly better to be strict than surprising.

jbcoe (approved):
This is definitely necessary and definitely an improvement. I wonder do we need to go further (can be a follow up)?

RyanJK5:
> This is definitely necessary and definitely an improvement. I wonder do we need to go further (can be a follow up)?

Do we need swap to be a member function? Could it just be a free function? And in that case, does the name need to be reserved?

jbcoe:
> > This is definitely necessary and definitely an improvement. I wonder do we need to go further (can be a follow up)?
> 
> Do we need swap to be a member function? Could it just be a free function? And in that case, does the name need to be reserved?

I think that we run the risk of confusing folk if `swap(p,pp)` and `p.swap(pp)` do different things.

There is already a library issues for `std::swap(i,ii)` and `swap(i,ii)` doing possibly different things for `indirect` and `polymorphic`. `indirect` and `polymorphic` provide `swap` as a hidden friend and as a member function.

https://github.com/cplusplus/papers/issues/2478

philipcraig:
> I wonder do we need to go further (can be a follow up)?

#450 is the follow-up for the names #449 does not cover: the private members of `protocol` and the views (`create`, `vtable`, `object_`, …) and the generated bases' injected-class-names (`type`, `member_bases_wrapper`) hide interface members too, with a compile error rather than silently. It proposes renaming them rather than reserving them.

Rejecting a static `swap` is #452, which also asks whether a data member or alias with a reserved name should be rejected.

conformance_tests.cc
philipcraig: Agreed that strict is better. Filed as #452 so the non-function members get decided with it.

RyanJK5:
> > > This is definitely necessary and definitely an improvement. I wonder do we need to go further (can be a follow up)?
> 
> > 
> 
> > Do we need swap to be a member function? Could it just be a free function? And in that case, does the name need to be reserved?
> 
> 
> 
> I think that we run the risk of confusing folk if `swap(p,pp)` and `p.swap(pp)` do different things.
> 
> 
> 
> There is already a library issues for `std::swap(i,ii)` and `swap(i,ii)` doing possibly different things for `indirect` and `polymorphic`. `indirect` and `polymorphic` provide `swap` as a hidden friend and as a member function.
> 
> 
> 
> https://github.com/cplusplus/papers/issues/2478

I don't disagree, but I guess in that case I don't follow why the same doesn't apply to valueless_after_move.

philipcraig:
Fair point: the parity argument that keeps `swap` a member would have kept `valueless_after_move` one too, and #395 moved it only to reserve one name fewer. #453 asks for one rule. #449 reserves whatever `protocol` declares today, so it does not depend on the answer.

jbcoe:
I think that `valueless_after_move` is suitably weird that people won't have such strong expectations about it.

It's not a watertight argument and I could be readily motivated to change my mind. I can't see why a protocol interface might need a valueless check on it.

jbcoe:
This argument actually cuts both ways "I can't see why a protocol interface might need a valueless check on it.". If folk won't need it, banning it is likely fine.

In particular

```
valueless_after_move(p); // works ok
p.valueless_after_move(); // Invokes UB
```

is a footgun that I don't want to be responsible for.

PR #454, opened 2026-09-23
Pin the GCC trunk snapshot to 2026-09-06

philipcraig:
The 2026-09-20 GCC trunk snapshot broke three nightly jobs on main (Coverage `consteval coverage`, CMake test `GCC trunk Release`, Sanitizers `asan`), for example https://github.com/jbcoe/cc-protocol/actions/runs/35836343715/job/107100609664. The same head passed on 2026-09-21 with the 2026-09-06 snapshot.

Both failures come from r17-4197 (P4101 consteval-only values, 2026-09-13), which represents `std::meta::info` as a `LANG_TYPE` in the middle end:

- Any `-flto` build of `protocol_test.cc`, `allocator_tests.cc` or `forwarding_test.cc` fails with `tree code 'lang_type' is not supported in LTO streams`. This is GCC PR126017, which Jakub Jelinek confirmed still reproduces after the P4101 commit. A two-line trigger is an interface type with internal linkage:

  ```c++
  #include "protocol.hh"
  namespace { struct Interface {}; struct Conforming {}; }
  int main() { xyz::reflection::protocol<Interface> p(Conforming{}); }
  ```

  compiled with `-std=c++26 -freflection -flto`.

- `-fsanitize=null` or `-fsanitize=vptr` (both in `-fsanitize=undefined`) reject the `template for` over `enumerators_of` in `tutorials/03_reflection.cc` with `consteval-only value outside an immediate function context`. Standalone reproducer, which the 2026-09-06 snapshot accepts:

  ```c++
  #include <meta>
  #include <string>

  template <typename E>
    requires std::is_enum_v<E>
  std::string enum_to_string(E value) {
    template for (constexpr std::meta::info e :
                  std::define_static_array(std::meta::enumerators_of(^^E))) {
      if (value == [:e:]) return std::string(std::meta::identifier_of(e));
    }
    return "<unknown>";
  }

  enum class Suit { Clubs, Diamonds, Hearts, Spades };

  int main() { return enum_to_string(Suit::Clubs) == "Clubs" ? 0 : 1; }
  ```

  compiled with `-std=c++26 -freflection -fsanitize=null -c`.

This pins the `install-gcc-trunk` action and `docker/Dockerfile` to the last good snapshot, `gcc-latest_17.0.0-20260906gitb6dabcdc335a.deb`, which is still served and which the action's cache key handles (a versioned URL has no redirect, so the file name and `Last-Modified` come from the file itself). The Dockerfile change makes existing sandbox images stale; a rebuild installs the same compiler they already have.

Restore `gcc-latest.deb` once a snapshot fixes both regressions; #455 tracks that.


jbcoe (approved):
Thanks for this!

